PR #1, opened 2022-03-28
Place junction boxes to optimize utilization

tcahill:


power.rb
```diff
@@ -0,0 +1,326 @@
...
+    @adjacency = make_adjacency_list
+  end
+
+  attr_accessor :adjacency, :edges
+
+  def make_adjacency_list
+    adjacency = {}
+    edges.each_value do |edge|
+      edge.vertices.permutation.each do |vertices|
+        adjacency[vertices[0].id] ||= []
```
raldi, line 82: Let’s either call them vertexes here or do a mass-replace and call them vertices everywhere.

raldi (commented):
Should there be a new edges.txt in this PR?

Also, let’s make a subdir for scripts like this instead of putting them in the root.

alextoombs (commented):
very stale review

alextoombs:
@jvyduna: one thing that's kind of weird (but I don't think is mission critical) is that by assigning controllers appropriately per panel and edge (only for edges whose signal originates from a controller, not a proximate edge) I only end up with 41 controllers (instead of the expected 49 controllers) and 141 channels. (instead of the expected 193 channels from the vertices tab of the sheet)

It's not obvious to me where the discrepancy is. But I think the risk is low. There's still some uncertainty here (e.g. do we really care about 49 controllers when we have a lot of solid panels still in design) and also low risk. (I guess worst case we really want 49, so we crimp an extra 8 ethernet runs from the junction boxes to the controllers)

alextoombs:
@jvyduna and I resolved the discrepancy and we're more or less in agreement with the sheet now. The root of the issue was: I was missing the constraint where we can drive a max number of pixels per controller channel, so some large panels require more than 1 channel.

However, we also found some opportunities to human-optimize this to try to avoid placing extra controllers. We can do this by looking at the model and finding controllers that only assign 1-2 channels that are located proximate to another controller with some channel headroom, then changing the values (in the sheets, and in the .tsv here) to use that controller instead. We've assigned 49 controllers today; it'd be nice to scale that down some to save on wiring and keep the remainders as standbys.

tcahill:
It was power - these panels were getting assigned but there was a bug where the assignment wasn't getting propagated to the `circuit` attribute on the `PanelStrip`. It's fixed now.

alextoombs:
rebased and force pushed with fixed LX version so I can do some controller optimization

alextoombs (approved):

PR #6, opened 2022-04-25
initial pulse and image patterns

theyoffa:
Adding the pulse patterns as demo'd as well as the sample image pattern.

Notable re-usable inclusions that were helpful for me for both types of patterns:
- Introduced the concept of sections so you can work with one area of the panels at a time
- Easy calculations for the dimensions of a set of panels


src/main/java/titanicsend/pattern/TEAudioPattern.java
```diff
@@ -51,6 +51,14 @@ protected TEAudioPattern(LX lx) {
         bassBandCount = 2;
     }
 
+    @Override
+    protected void run(double deltaMs) {
+        computeAudio(deltaMs);
```
theyoffa, line 56: i also think it could be helpful to do something similar at the TEPattern level for things like `updateGradients()` if people are on board with it

src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -18,6 +19,7 @@ public class TEWholeModel extends LXModel {
   public HashMap<Integer, TEVertex> vertexesById;
   public HashMap<String, TEEdgeModel> edgesById;
   public HashMap<String, TEPanelModel> panelsById;
```
theyoffa, line 21: I'm thinking we may want to create accessors that return new objects for these to preserve encapsulation similar to panelsBySection. My fear is some pattern changes these objects by accident which creates a bug that's really annoying to reproduce/track down. Lmk your thoughts.

src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -402,4 +412,31 @@ private static Geometry loadGeometry(String subdir) {
...
+
+  public Set<TEPanelModel> getPanelsBySections(Collection<TEPanelSection> sections) {
+    return panelsBySection.entrySet().stream()
+            .filter(entry -> sections.contains(entry.getKey()))
+            .map(Map.Entry::getValue)
+            .flatMap(Set::stream)
+            .collect(Collectors.toSet());
+  }
+
+  public Set<TEPanelModel> getLeftPanels() {
```
raldi, line 428: I'm not sure what you mean here by left, right, etc. Do you mean fore/aft/etc?

src/main/java/titanicsend/pattern/yoffa/EddiePattern.java
```diff
@@ -0,0 +1,36 @@
...
+import titanicsend.pattern.TEPattern;
+
+import java.io.IOException;
+
+// This is just an example to show how we could map an image onto a set of our panels
+// Not intended from production use
+@LXCategory("Audio Examples")
+public class EddiePattern extends TEPattern {
+
+    private static final String IMG_PATH = "src/main/java/titanicsend/pattern/yoffa/assets/eddie.jpg";
```
raldi, line 15: Let's put this in `resources`

src/main/java/titanicsend/util/TEMath.java
```diff
@@ -55,6 +55,16 @@ public static double easeOutPow(double basis, double normalizedPower) {
         return Math.pow(basis, 1. / scaledPow);
     }
 
+    /**
+     *  Distance between two arbitrary three-dimensional points
+     */
+    public static double distance(float x1, float y1, float z1, float x2, float y2, float z2) {
```
raldi, line 61: Seems weird to take floats and return a double

src/main/java/titanicsend/model/TEWholeModel.java
jvyduna: A lovely thing one of us could do at some point is add a toggle-able origin trident that labels visually "+Z is this way - AKA Fore. +Y is Up. +Z is Starboard" (and those might even be reversed because I don't have it memorized yet).

src/main/java/titanicsend/util/TEMath.java
jvyduna: I believe this exists in the framework as LXVector.dist(other point)

src/main/java/titanicsend/model/TEWholeModel.java
theyoffa: i think shouldn't be bad with intellij's refactor, but we'll see how ambitious i feel

src/main/java/titanicsend/model/TEPanelModel.java
```diff
@@ -149,4 +149,22 @@ public Set<TEPanelModel> neighbors() {
     }
     return rv;
   }
+
+  // TODO Replace this with something smarter
+  // This is a really lazy/sloppy way of doing this, though if we're not changing the model it should be fine
```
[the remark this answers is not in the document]
theyoffa: will do

src/main/java/titanicsend/model/TEWholeModel.java
theyoffa: hmm...given that one side is solid and the opposite contains the majority of the LEDs, i was considering the latter as the "front" for the purposes of lighting and i was thinking of it as if i was looking from the audience, since I usually am. i'm fine with whatever we use, let's just be consistent. can use fore/aft if someone tells me which way is the front lol.

src/main/java/titanicsend/pattern/yoffa/EddiePattern.java
theyoffa: hmm...perhaps assets directory would be better? or maybe we should combine them if there isn't a clear difference.

src/main/java/titanicsend/util/TEMath.java
theyoffa: agree, will update to all double. tbf, i did copy that method signature from TEVertex.distance ;)

i did initially try it by creating LXVectors, but this requires also creating LXPoints, which need to be normalized with the model to be initialized properly, and gets assigned an index and has lots of other implications that go with it. for adhoc points, this seemed too heavyweight imo.

src/main/java/titanicsend/model/TEWholeModel.java
jvyduna: It's a good point. As a quick fix, I added this as a reference to [the top of the README](https://github.com/titanicsend/LXStudio-TE/commit/51984fb62c4eff81b3dfacf97f02451126d4d772).

src/main/java/titanicsend/model/TEPanelModel.java
```diff
@@ -150,21 +148,22 @@ public Set<TEPanelModel> neighbors() {
...
-      return centroid.z < 0 ? TEPanelSection.SIDE_RIGHT : TEPanelSection.SIDE_LEFT;
+      return centroid.z < 0 ? TEPanelSection.FORE : TEPanelSection.AFT;
     }
 
     if (centroid.y < 5000000) {
-      return centroid.z < 0 ? TEPanelSection.FRONT_RIGHT : TEPanelSection.FRONT_LEFT;
+      return centroid.z < 0 ? TEPanelSection.STARBOARD_FORE : TEPanelSection.STARBOARD_AFT;
     } else {
-      return centroid.z < 0 ? TEPanelSection.FRONT_RIGHT_SINGLE : TEPanelSection.FRONT_LEFT_SINGLE;
+      return centroid.z < 0 ? TEPanelSection.STARBOARD_FORE_SINGLE : TEPanelSection.STARBOARD_AFT_SINGLE;
```
raldi, line 166: I'm not sure what SINGLE means in this context.

raldi (approved):

PR #7, opened 2022-05-01
more refinement to the PB stuff, checkboxes to fitler point type

simap:
also added model type to the tag, so you can use View selector filters
in the UI on a channel.

Matches what's in the video.

### DIFF
--- a/src/main/java/titanicsend/pattern/pixelblaze/Glue.java
+++ b/src/main/java/titanicsend/pattern/pixelblaze/Glue.java
@@ -30,6 +30,6 @@ public class Glue {
   }
 
   public static int setAlpha(int color, float a) {
-    return Math.min(255, (int)(a * 255.0F)) << 24 | (color & LXColor.RGB_MASK);
+    return Math.max(0,Math.min(255, (int)(a * 255.0F))) << 24 | (color & LXColor.RGB_MASK);
   }
 }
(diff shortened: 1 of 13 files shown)


src/main/java/titanicsend/pattern/pixelblaze/Glue.java
```diff
@@ -30,6 +30,6 @@ public static int rgba(float r, float g, float b, float a) {
   }
 
   public static int setAlpha(int color, float a) {
-    return Math.min(255, (int)(a * 255.0F)) << 24 | (color & LXColor.RGB_MASK);
+    return Math.max(0,Math.min(255, (int)(a * 255.0F))) << 24 | (color & LXColor.RGB_MASK);
```
raldi, line 33: This line would be a lot more readable if broken into several.

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePatternParallel.java
```diff
@@ -14,7 +12,15 @@ public class PixelblazePatternParallel extends TEAudioPattern {
   public static final int N_THREADS = 4;
```
raldi, line 12: If this class were renamed PixelblazeParallelPattern, the UI would automatically label it "PixelBlazeParallel"

PR #10, opened 2022-05-05
shaders

theyoffa:
ported over the following fragment shaders and make them configurable.

https://www.shadertoy.com/view/7l3GDS

neon_heart

https://www.shadertoy.com/view/wstGz4

grid

also abstracted out a lot of the fragment shader stuff so it should be really easy to make new ones.


src/main/java/titanicsend/util/TEMath.java
```diff
@@ -65,6 +67,122 @@ public static double distance(double x1, double y1, double z1, double x2, double
...
+        return sum;
+    }
+
+    /**
+     *  Built-in function from GLSL fragment shader language ported to Java
+     *  Performs smooth Hermite interpolation between 0 and 1
+     *
+     *  See: https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/smoothstep.xhtml
+     */
+    public static double smoothstep(double edge0, double edge1, double x)
```
raldi, line 88: Is there a reason this isn't camelCase?

raldi (approved):
This is a little too big for me to thoroughly review, but seems good at a glance and the results are phenomenal, so merge if you're ready.

src/main/java/titanicsend/util/TEMath.java
theyoffa: it's actually one word ha https://en.wikipedia.org/wiki/Smoothstep

PR #14, opened 2022-05-10
shimmering edge pattern

theyoffa:
Pretty basic, lower key but IMO kinda cool looking edge pattern. Sync'd to the beat.

power_grid


raldi (approved):
I wonder if we could use this technique to wipe between patterns, or (if that's too hard), wipe black over a pattern, switch patterns out underneath it, and wipe the black away to reveal the new one.

theyoffa:
hm...interesting. would be easy to do manually, but not sure if it's something we could somehow integrate with LXStudio's auto-transition between patterns. will look into it.

PR #19, opened 2022-05-18
new layered pattern framework

theyoffa:
Ok so um, fair warning, in this pr i've refactored all of my code :). That said, this change is entirely limited to my own code, so I'm confident I'm not impacting/breaking anyone else. I considered breaking this up into smaller PRs, but idk how much anyone really cares about refactors of my own code. Lmk if you would like this. I also did some renaming and moving of files, so it looks like more than it is. There's like 4 new files that actually matter listed below.

**What have I done/what's the point?**

Basically after spending enough time experimenting with/demoing different combinations of different variants of different patterns on different set of models, I felt the standard pattern file process to be very tedious, leading to lots of code duplication, and painful to experiment with.  What I wanted was a framework that would do the following:

- Allow me to point a pattern (or pattern effect as I have now dubbed them) at any combo of points and/or models in one line, each configurable as tiled or as one canvas
- Allow me to create multiple variants of the same effect point at a different set of points (eg edges instead of panels) or have a modified configuration with just one more line and no duplication
- Allow me to add different combinations of different effects and optionally blend them in one line
- Automatically add all the parameters associated with all of the included pattern effects to all the variants by default
- Do this all in one file so I don't have to create/register a whole new file for every single variant

[Future TODO features]
- Automatically infer proper swatch colors
- One line mapping from the beat/bass/treble/energy to parameters of an effect

This may not seem like crazy functionality, but if you've been messing with patterns as much as I have, it's a huge QoL improvement and makes experimentation/innovation much easier. Also the context required to port a shader, hook it up to panels, and start messing with it is basically none now.

The effective result is PatternConfig.java and you can see the features in action there. Due to the LX registry's (imo obnoxious) insistence on using reflection I needed actual separate java classes so I couldn't quite get one line, but it's pretty close. From there you can take a look at ConstructedPattern, PatternEffect, and PatternTarget and that's basically everything. The rest is just conforming my previous patterns to that.

Also I'm aware there is an actual LX Pattern Effect, but I prefer mine as it's more TE-centric (targeting panels and whatnot) , better for creating preset variants, and I'm more comfortable in code than LX studio.


raldi (approved):
I accept your apology for being prolific and awesome.

theyoffa:
:)

PR #22, opened 2022-05-23
[WIP] Calculate needed power cable lengths

tcahill:


script/power_distribution/calculate_line_lengths.rb
```diff
@@ -30,48 +22,36 @@ def straight_line_distance(point1, point2)
...
     end
   end
+
+  cable_lengths
 end
-edge_lengths_count = edge_lengths_count.sort
 
-panel_lengths_count = {}
+def bucket_cable_lengths(cable_lengths)
+  buckets = [5, 10, 15, 20, 25, 30]
```
alextoombs, line 47: this seems like a way better way of doing what I was going to do, thank you

tcahill: btw, this currently outputs:

```
{5=>170, 10=>290, 15=>136, 20=>71, 25=>18, 30=>0}
```

which is kind of good - cable lengths are generally weighted towards the shorter end - but something is amiss because the placement algorithm was supposed to ensure that every edge is assigned to a box less than 17 feet away.

tcahill: Improved by always checking the box distance from the "signal in vertex" of each edge when deciding if it's close enough:

```
{5=>190, 10=>319, 15=>176, 20=>18, 25=>6, 30=>0}
```

tcahill: That's 7290 feet of wire by the way

PR #23, opened 2022-05-23
Add script to render power connections as SVG

tcahill:
image


tcahill:
This is a WIP - currently the diagram just displays the number of enclosures at each vertex and the connections between boxes at a vertex and edges starting at some other vertex, but I'd like to add more info including:
  - lines from enclosures to panel centroids to show where we're powering panels from
  - counts next to each dotted line for the number of conductor pairs being routed between points
  - where we're using the internal 10AWG conductors in edges instead of running external wiring
  
I'd also like to produce narrowed versions of this diagram that are scoped to a single power supply enclosure and just show what that enclosure powers

raldi (commented):
This is cool. You should show it off.

PR #26, opened 2022-06-10
Added OSC listener code

worldveil:
This is very simple - just adds a OSC listener to the TEApp code. 

I tested this with ShowKontrol, and it seems to work:

Screen Shot 2022-06-10 at 3 54 52 PM

### DIFF
--- a/src/main/java/titanicsend/app/TEApp.java
+++ b/src/main/java/titanicsend/app/TEApp.java
@@ -20,6 +20,7 @@ package titanicsend.app;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.SocketException;
 
 import heronarts.lx.LX;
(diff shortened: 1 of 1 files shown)


src/main/java/titanicsend/app/TEApp.java
```diff
@@ -52,6 +53,8 @@ public class TEApp extends PApplet implements LXPlugin  {
   private GigglePixelListener gpListener;
   private GigglePixelBroadcaster gpBroadcaster;
 
+  private static int OSC_PORT = 9876;
```
raldi, line 56: Add `final`

raldi (approved):

worldveil:
Closing this since the built-in listener seems to work fine for OSC.

PR #27, opened 2022-06-13
README: retool for easy adoption

alextoombs:
This moves the full IDE installation (what any pattern-writer
would need) ahead of the command-line installation (more just
to play around, but presumes a pretty sophisticated install
of maven et al) under the assumption that most people doing
this will want to tweak patterns.

Also, add a whole lot more images.


raldi:
I'm not going to have a chance to properly review this anytime soon, but it seems like a worthy improvement, so if a couple others sign off, that's good enough for me.

README.md
```diff
@@ -1,83 +1,92 @@
...
-* [Using Tempo and Sound](https://docs.google.com/document/d/17iICAfbhCzPL77KbmFDL4-lN0zgBb1k6wdWnoBSPDjk/edit) 
-* [The APC40 and LX Studio](https://docs.google.com/document/d/110qgYR_4wtE0gN8K3QZdqU75Xq0W115qe7J6mcgm1So/edit)
+Steps for setup:
+1. Clone the git repo you're looking at:
+   ```
+   git clone https://github.com/titanicsend/LXStudio-TE.git
+   ```
 
-As you really get rolling, you’ll appreciate the [API docs](https://lx.studio/api/) and public portion of [the source](https://github.com/heronarts/LX/tree/master/src/main/java/heronarts/lx).
+2. Download and install [Processing 4](https://processing.org/download); then,
```
tcahill, line 29: This shouldn't be necessary - Processing is brought in as a JAR during the build

tcahill (approved):

PR #30, opened 2022-06-15
[WIP] power_distribution: place outlet_banks; assign to boxes

alextoombs:
And output the AC lengths that we need.

Probably doing more with outlets/circuits than we really need to tbh

### TODO

- [x] figure out offset length; see TODO In source
- [x] figure out some way to sanity check these lengths for reasonable-ness
- [x] figure out some way to track extension cords more

Output of the script is: (`length in feet => count of cables needed in that size`)

```
AC power cable lengths:
{15=>6, 20=>16, 25=>11, 35=>5}
---------
```

### DIFF
--- a/script/power_distribution/calculate_line_lengths.rb
+++ b/script/power_distribution/calculate_line_lengths.rb
@@ -62,8 +62,21 @@ def ethernet_cable_lengths(graph:, boxes:)
   cable_lengths
 end
 
+def ac_power_cable_lengths(graph:, boxes:)
+  cable_lengths = []
+
+  boxes.each do |box|
+    length = min_distance_between_vertices_in_feet(graph, box.vertex.id, box.outlet_bank.vertex.id)
+    length += AC_POWER_WIGGLE_FEET
+
+    cable_lengths << length
+  end
+
+  cable_lengths
+end
+
 def bucket_cable_lengths(cable_lengths)
-  buckets = [5, 10, 15, 20, 25, 30]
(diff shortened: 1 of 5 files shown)


script/power_distribution/calculate_line_lengths.rb
```diff
@@ -62,6 +62,25 @@ def ethernet_cable_lengths(graph:, boxes:)
...
 
+def ac_power_cable_lengths(graph:, boxes:)
+  cable_lengths = []
+
+  boxes.each do |box|
+    length = min_distance_between_vertices_in_feet(graph, box.vertex.id, box.outlet_bank.vertex.id)
+    length += AC_POWER_WIGGLE_FEET
+
+    # TODO: apply fudge factors for length. I think what we want to do is:
+    #       1. figure out if the outlet bank is physically (not in this model) on the port (crowd) or starboard side
```
alextoombs, line 73: starboard is crowd, not port

alextoombs:
I think the 35 foot ones are cases where we'd probably want to run an extension cord. Mostly I'm unsure if our estimate of the "wiggle" (offset from center, extra length to account for following the edges) is acceptable or a huge overestimate. OVerestimating seems like the right move; we can always cut these down to size more later on.

alextoombs:
https://titanicsend.slack.com/archives/C03650K7AE7/p1655766560717079?thread_ts=1655751180.553809&cid=C03650K7AE7

PR #31, opened 2022-06-16
Add fire pattern

tcahill:
Forgot to port this over when we switched repos. This is now connected to our
palette! It looks a little weird just lerp-ing the color from black -> [palette
color] -> white (when the palette color is purple, there's a lot of green in the
result for example), so I may want to add some additional parameters

fire


src/main/java/titanicsend/effect/Kaleidoscope.java
```diff
@@ -14,17 +17,20 @@
 
 @LXCategory(LXCategory.TEST)
 public class Kaleidoscope extends BasicEffect {
-
-    private int numSegments = 6;
+    private final DiscreteParameter segments = new DiscreteParameter("Segments", 6, 2, 12);
```
tcahill, line 20: Whoops, some changes from another PR that I forgot to commit before merging - but I do want to get these in.

PR #39, opened 2022-06-18
better support for side and single panels

theyoffa:
support for side panels for shaders

also include single panels in splitPanelSections while I'm at it


worldveil:
LGTM - panels show up now.

PR #41, opened 2022-06-22
native glsl implementation with shadertoy integration

theyoffa:
we're down the rabbit hole now boys!

as an aside, i feel like these apis were made intentionally obnoxious. maybe that's just c-style in general.


theyoffa:
given the blast radius is basically 0, i'm going to start auto-merging changes that are confined to my own patterns package. lmk if there is any objection.

PR #42, opened 2022-06-22
change back panels from solid to lit

theyoffa:


resources/vehicle/panels.txt
```diff
@@ -40,29 +40,29 @@ SAC	28-115	26-28	26-115	unflipped	lit
...
+PBA	128-129	78-128	78-129	unflipped	lit
+PAC	73-75	73-128	75-128	unflipped	lit
+PAD	73-92	73-75	75-92	unflipped	lit
+PBB	97-128	78-97	78-128	unflipped	lit
+PBC	75-97	75-128	97-128	unflipped	lit
+PBD	75-96	75-97	96-97	unflipped	lit
+PCA	78-97	78-79	79-97	unflipped	lit
+PCB	80-97	79-80	79-97	unflipped	lit
+PCC	96-97	80-96	80-97	unflipped	lit
+PBE	75-96	52-75	52-96	flipped	lit
```
theyoffa, line 54: @jvyduna @raldi I'm guessing this panel is flipped for a reason? what's the context here? wanted to double check before i looked into putting the lights on the other side of the panel

raldi: The LX UI renders an opaque triangle on the back of each panel, to look more like reality. The problem is, how does it know which side is the back? Well, it's the one closer to the vehicle centroid ... usually. I started with that, and the ones that were wrong, I marked as "flipped", which just means "invert the heuristic guess about which side of this panel is the back"

raldi (approved):

PR #43, opened 2022-06-26
Add FollowThatStar pattern

zranger1:
Hi!  This is my first attempt at a pattern for the vehicle.  It's beat and sound reactive, and gets quite a bit crazier as you increase the "energy" parameter.  Please let me know if it performs reasonably, and if I need to fix any of the project structure I added to make the pattern work.

Thanks!
 Jon


zranger1:
Slightly broken... will resubmit once fixed.

PR #45, opened 2022-06-27
TEAutopilot MVP

worldveil:
Introduces `TEAutopilot` class which is an `LXLoopTask` that now runs on normal FPS schedule along with main thread. 

Screen Shot 2022-06-30 at 10 04 31 PM

`TEAutopilot`'s job (only when `enabled=true`) is to observe OSC messages and audio state to automatically trigger channels, patterns, and clips when appropriate. 

A few notes:

- `TEBeatEvent`, `TEPhraseEvent`, `TEOscMessage` : structs to hold data about past events
- `TEChannelName`, `TEPhrase` : enums corresponding to channels in AutoVJ .lxp and phrase types, respectively
- `TETimeUtils`, `TEMixerUtils` : utiility functions to aid in calculations or manipulating LX mixer
- `TEHistorian` : wrapper around historical data that `TEAutopilot` relies on to understand the past
- `CircularArray`: constant size storage and access time buffer to handle storage of a small window of previous events

Two main settings control how autopilot works:

- `enabled` is whether or not autopilot will actually act, with this off the loop simply exits out after checking state. You may toggle this in Global UI of LX Studio
- `autoBpmSyncEnabled`: ALPHA! if this is `true`, autopilot will detect when LX's BPM drifts too far outsize what OSC messages on beats are indicating. ANY OSC message from ShowKontrol overrides this -- this is just to fine tune when sometimes setBPM messages do not come or are slightly wrong. I have found this in practice to happen occasionally.  It is my hope we can find a way to have ShowKontrol just output the correct value or some other fix and we can delete this setting!


worldveil:
WIP, don't merge

src/main/java/titanicsend/app/TEApp.java
```diff
@@ -76,13 +90,31 @@ public void setup() {
     this.surface.setTitle(this.model.name);
   }
 
+  public static class AutopilotComponent extends LXComponent implements LXOscComponent {
```
raldi, line 93: This too would probably be better in its own file.

src/main/java/titanicsend/app/TEApp.java
```diff
@@ -160,6 +224,14 @@ public void initializeUI(LXStudio lx, LXStudio.UI ui) {
     // for headless mode should go in the raw initialize method above.
   }
 
+  public static class AutopilotUIComponent extends UICollapsibleSection {
```
raldi, line 227: This too

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+    private boolean autoBpmSyncEnabled = true;
+
+    private static double LEVEL_FULL = 1.0,
+                           LEVEL_BARELY_ON = 0.01,
+                           LEVEL_HALF = 0.5,
+                           LEVEL_OFF = 0.0;
+
+    private static float PROB_CLIPS_ON_SAME_PHRASE = 0.5f;
+
+    // our ref to global LX object
```
raldi, line 29: This one probably doesn't need a comment, but the one above does

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+
+        // log beat event
+        history.logBeat(beatAt);
+
+        // Tempo fine adjustment due to irregularities with OSC
+        // If you want to ignore this, simply turn `autoBpmSyncEnabled`=false
+        if (autoBpmSyncEnabled && history.readyForTempoEstimation()) {
+            double estBPMAvg = history.estimateBPM();
+            if (Math.abs(lx.engine.tempo.bpm() - estBPMAvg) > BPM_ERROR_ADJUST) {
+                System.out.printf("BPM est from beats: %f, LX BPM is: %f, overriding BPM --> %f!\n"
```
raldi, line 116: Why not lx.log()?

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+
+            // collect audio/FFT statistics
+            // TODO(will) for when we want to act without OSC messages -- just based on pure audio
+
+            // check for new OSC messages
+            while (unprocessedOscMessages.size() > 0) {
+                // grab a new message off the queue
+                TEOscMessage oscTE = unprocessedOscMessages.poll();
+                if (oscTE == null)
+                    // this should never happen, since we test for size() of queue, but good to check
```
raldi, line 138: Maybe log something here?

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+
+                } else if (TEOscPath.isDownbeat(address)) {
+                    // nothing to do yet
+
+                } else if (TEOscPath.isBeat(address)) {
+                    onBeatEvent(oscTE.timestamp);
+
+                } else {
+                    // unrecognized OSC message!
+                    System.out.printf("Don't recognize OSC message: %s\n", address);
```
raldi, line 156: LX.error()

raldi (changes requested):
Wow! Just some cleanups. Let's get all the printf's changed to proper logging.

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+
+    public void setAutoBpmSyncEnabled(boolean enableSync) {
+        if (enableSync && !this.enabled) {
+            System.out.println("Cannot autosync BPM, requires: autopilot.setEnabled(true)!\n");
+            return;
+        }
+        this.autoBpmSyncEnabled = enableSync;
+    }
+
+    public ArrayList<LXClip> collectClipsToTrigger(TEPhrase newPhrase, boolean isSamePhrase) {
```
theyoffa, line 313: not really familiar with LXClips and what we're doing here but it seems cool! would love to chat about it next meeting.

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+            if (predictedCorrectly) {
+                // we nailed it!
+                System.out.printf("[AUTOVJ] We predicted correctly!\n");
+                setChannelTo(curChannel, LEVEL_FULL);
+                setChannelTo(prevChannel, LEVEL_OFF);
+
+            } else {
+                // we didn't predict the phrase change correctly, turn off
+                // the channel we were trying to transition into
+                oldNextChannelName = TEMixerUtils.getChannelNameFromPhraseType(oldNextPhrase);
```
theyoffa, line 251: so i guess when this happens we would be teasing a transition to a new pattern, then shutting it off when the next phrase comes? maybe fine/looks on purpose :)

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+                oldNextChannel = TEMixerUtils.getChannelByName(lx, oldNextChannelName);
+                echoMode = (oldNextChannel != null);
+                System.out.printf("[AUTOVJ] We didn't predict right, oldNextChannelName=%s, echoMode=%s\n", oldNextChannelName, echoMode);
+
+                setChannelTo(curChannel, LEVEL_FULL);
+                //setChannelTo(oldNextChannel, LEVEL_OFF);
```
theyoffa, line 257: i think we need this?

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        // set strobes channels autocycle time fast
+        double msPerDivision = TETimeUtils.bpmToMsPerBeat(lx.engine.tempo.bpm()) / 16.; // sixteeth notes
+        strobesChannel.autoCycleTimeSecs.setValue(msPerDivision);
+
+        return clips;
+    }
+
+    public TEPhrase guessNextPhrase(TEPhrase newPhrase) {
```
theyoffa, line 350: what will happen here if the DJ starts playing loops? will all our predictions be wrong?

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+
+    @Override
+    public void loop(double deltaMs) {
+        long now = System.currentTimeMillis();
+        try {
+            // if autopilot isn't enabled, just ignore for now
+            if (!isEnabled()) return;
+
+            // collect audio/FFT statistics
+            // TODO(will) for when we want to act without OSC messages -- just based on pure audio
```
theyoffa, line 131: we're already doing some of this in TEAudioPattern. when the time comes, might sense to extract into something that works for both use cases...like a TE wrapper to the lx engine

src/main/java/titanicsend/app/autopilot/TEBeatEvent.java
```diff
@@ -0,0 +1,22 @@
+package titanicsend.app.autopilot;
+
+/**
+ * Struct for tracking OSC beat events. Useful for
+ * computing BPM or performing analysis on previous
+ * audio or other signal events to predict future ones.
+ */
+public class TEBeatEvent {
```
theyoffa, line 8: unless you're planning on adding more here, you could prolly just use the a long directly rather than wrapping it.

src/main/java/titanicsend/app/autopilot/TEPhrase.java
```diff
@@ -0,0 +1,46 @@
...
+    private TEPhrase(String name) {
+        this.name = name;
+    }
+
+    private static ConcurrentHashMap<String, TEPhrase> initializeMap() {
+        /*
+            Static map initialization:
+            https://stackoverflow.com/a/509016
+         */
+        ConcurrentHashMap<String, TEPhrase> m = new ConcurrentHashMap<String, TEPhrase>();
```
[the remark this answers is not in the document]
worldveil: oh this is super cool!

but doesn't seem to work for ConcurrentHashMaps?

```
no instance(s) of type variable(s) K, V exist so that Map<K, V> conforms to ConcurrentHashMap<String, TEPhrase>
```

?

src/main/java/titanicsend/app/autopilot/TEPhrase.java
```diff
@@ -0,0 +1,46 @@
...
+
+        // ADD ALL PHRASES HERE
+        m.put("tro", TRO);
+        m.put("up", UP);
+        m.put("down", DOWN);
+        m.put("chorus", CHORUS);
+
+        //TODO(will) this line causes thread to freeze...no idea why...
+        //m = (ConcurrentHashMap<String, TECanonicalPhrase>) Collections.unmodifiableMap(stringToPhrase);
```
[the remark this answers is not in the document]
worldveil: oh duh

src/main/java/titanicsend/util/CircularArray.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * This class is an immutable, constant size array. As you add elements,
+ * they are appended to the end. Very simple -- there is no way to remove
+ * elements. They just get overwritten with time. Nice for small, constant
+ * time access latency and memory usage.
+ *
+ * @param <T> for the type you wish to hold in the array.
+ */
+public class CircularArray<T> {
```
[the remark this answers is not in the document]
worldveil: Ah, but I wanted index access - don't think these have them right? I couldn't find anything that was both a circular array and had support to query by index

src/main/java/titanicsend/app/autopilot/TEChannelName.java
```diff
@@ -0,0 +1,35 @@
...
+    TRIGGERS(4),
+    TRO(5); // in-tro and out-tro's;
+
+    private final int index;
+
+    private TEChannelName(int idx) {
+        index = idx;
+    }
+
+    public int getIndex() {
```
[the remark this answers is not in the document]
worldveil: that constrains us to have contiguous, ordered indices though right? I wanted flexibility in case we wanted to skip a channel or something

src/main/java/titanicsend/app/autopilot/TEOscPath.java
```diff
@@ -0,0 +1,45 @@
+package titanicsend.app.autopilot;
+
+/**
+ * Class to define and extract actions and events from
+ * OSC messages received in LX.
+ *
+ * Please keep these up to date if you change ShowKontrol
+ * automation fields!
+ */
+public class TEOscPath {
```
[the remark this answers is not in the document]
worldveil: ah great point

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+import java.util.Random;
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+public class TEAutopilot implements LXLoopTask {
+    // TODO(will) make false, users can enable from global UI toggle panel
+    private boolean enabled = false;
+    // should we try to sync BPM when ProDJlink seems off?
+    private boolean autoBpmSyncEnabled = true;
+
+    private static double LEVEL_FULL = 1.0,
```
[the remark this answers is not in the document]
worldveil: added!

src/main/java/titanicsend/app/TEAutopilot.java
worldveil: hmm yeah could yes

src/main/java/titanicsend/app/TEAutopilot.java
worldveil: added one!

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -0,0 +1,378 @@
...
+        } catch (IndexOutOfBoundsException e) {
+            // no phrase events detected yet
+        }
+    }
+
+    private void updatePhraseState(TEPhrase newPhrase) {
+        echoMode = false;
+
+        // phrase state
+        oldNextPhrase = nextPhrase;
```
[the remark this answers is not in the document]
worldveil: curPhrase is always certain

nextPhrase is uncertain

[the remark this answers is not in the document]
worldveil: the point of reference is always at current moment, since that's when we receive the OSC phrase message. that's the reason for the awkward wording

PR #46, opened 2022-06-28
Additional Patterns

zranger1:
EdgeKITT - a tempo reactive KITT edge foreground pattern
FollowThatStar - a combo foreground train of stars that swirls around the display in time to music
Iceflow - transparent, partially frozen fluid.  Very music reactive.


simap:
These are awesome! Very nice effects. 

One suggestion is for EdgeKit, the cycles parameter can make it gltichy between measures unless it has an integer value. Maybe use one of the tempo divisions like in EdgeProgressions?

zranger1:
Thanks, Ben!   Good catch - now fixed to allow time division changes only at the start of cycles, which smooths things out.   Also, I expanded the range of possible speeds, because you never know what the DJ is going to want to do!

PR #52, opened 2022-07-07
Added logging convience function

worldveil:
Makes it faster to log/error in TE!


raldi (approved):
This is great; you should send it upstream to Mark.

worldveil:
awesome, will merge. 

and oh cool, maybe when I do the BPM checking PR I'll include that in there?

PR #56, opened 2022-07-12
Disable non-used channels to save FPS and ignore old OSC messages

worldveil:
This saves CPU cycles by disabling channels when setting the fader value to 0 or very close. 

Also discards old OSC messages, which would often mess up if the change to master deck in ShowKontrol took too long.

### DIFF
--- a/src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
+++ b/src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
@@ -26,6 +26,18 @@ public class TEMixerUtils {
         }
     }
 
+    public static void setFaderTo(LX lx, TEChannelName name, double faderLevel) {
+        LXChannel channel = (LXChannel) lx.engine.mixer.channels.get(name.getIndex());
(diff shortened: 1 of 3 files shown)


src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
```diff
@@ -26,6 +26,18 @@ public static void setChannelExclusivelyVisible(LX lx, TEChannelName channel) {
         }
     }
 
+    public static void setFaderTo(LX lx, TEChannelName name, double faderLevel) {
+        LXChannel channel = (LXChannel) lx.engine.mixer.channels.get(name.getIndex());
+        channel.fader.setValue(faderLevel);
+
+        // save CPU cycles by disabling OFF channels
+        if (faderLevel < 0.01) {
```
raldi, line 34: Make a constant

worldveil: oops, I clicked merge before this loaded, will do in another commit!

PR #57, opened 2022-07-13
power_distribution: account for boxes with/without switches

alextoombs:
This doesn't do much beyond just printing this info out, but
we'll want to reflect it in the wiring diagram later on.

We only have 25 boxes with switches due to supply constraints.


alextoombs:
20 boxes need switches

tcahill (approved):

PR #66, opened 2022-07-21
Add LaserPanels pattern

zranger1:


raldi:
Can we rename this to something else, unless it involves actual lasers?

zranger1:
Of course!  (... yikes,  now I have to think of a name ...)

PR #72, opened 2022-07-25
 Provisional Fix:  Change to allow projects to load w/o autopilot channels

zranger1:
Prior to this, new projects created without the channel setup expected by autopilot would cause a chain of exceptions, resulting in the UI not working.   This allows non-autopilot projects to load successfully.

However, would greatly appreciate review for downstream side-effects.


raldi:
Looks good. Please merge if ready.

PR #73, opened 2022-07-26
Wait for LX mixer to load

worldveil:
In some cases the LX mixer isn't ready yet, so trying to access channels will break things. This fixes that.

### DIFF
--- a/src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
+++ b/src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
@@ -70,6 +70,7 @@ public class TEMixerUtils {
             } catch (IndexOutOfBoundsException e) {
                 numTries++;
                 double waitMs = Math.pow(2.0, (double)numTries) * 1000;
+                TE.log("LX mixer wasn't ready yet, waiting %f seconds...", waitMs / 1000.);
                 Thread.sleep((long)waitMs);
             }
         }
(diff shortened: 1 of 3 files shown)


src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
```diff
@@ -56,10 +56,24 @@ else if (phraseType == TEPhrase.TRO)
...
+
+        // this backoff/retry loop is needed in case there's a delay in intializing
+        // the LX mixer / channels. this often happens in LX projects where there are
+        // a large number of shaders that need to be loaded
+        int numTries = 0;
+        while (numTries < 8) {
+            try {
+                return (LXChannel) lx.engine.mixer.channels.get(name.getIndex());
+            } catch (IndexOutOfBoundsException e) {
+                numTries++;
```
raldi, line 71: Add LX.log()

raldi: Also, is it intention that you say InterruptedException above and IOOBE here?

worldveil: done!

worldveil: yes, the InterruptedException is for the wait(). java requires it

worldveil: @raldi to be fair, we could do a nested try/catch to prevent other methods from needing that throws in their signature. up to you

PR #76, opened 2022-07-26
Update panel signal paths (again)

alextoombs:
This squashes a couple of controllers and regenerates the signal paths to match some of the vertices we chose as signal start points. It also makes `panel_signal_paths.tsv` the source of truth to generate this.

I've already updated the sheet: https://docs.google.com/spreadsheets/d/1C7VPybckgH9bWGxwtgMN_Ij1T__c5qc-k7yIhG-592Y/edit#gid=2075432802

We have some disconnects between "what is the longest edge" and "what vertices does the cut panel have as the smooth edge", but we've had those for a few days now, and we've just been routing around them. I want to ensure we come back and update this in ruby when we're done though. The "panel build tracker" sheet has a bunch of hidden columns to compare the calculations from this, the OG panels tab, and the `strip_assignment` script that Jeff ran.


script/power_distribution/constants.rb
```diff
@@ -20,4 +20,4 @@
 # extra length is not the end of the world.
 AC_POWER_WIGGLE_FEET = 12
 
-MAX_CONTROLLER_DISTANCE_SIGNAL_TO_FIRST_PIXEL_FEET = 7
\ No newline at end of file
+MAX_CONTROLLER_DISTANCE_SIGNAL_TO_FIRST_PIXEL_FEET = 7.1
```
alextoombs, line 23: There are a couple that are just a hair over.

alextoombs:
This saves a couple controllers btw, so I think it's worth changing now.

alextoombs:
Raldi said ship it, I want to update the ones that reflect reality at least

PR #77, opened 2022-07-26
Fix for stuttering on native shader pattern load/activate

zranger1:
Implemented new shader caching system which should greatly speed up the loading of shader-based patterns and clips.   

Details:
- Once a shader is loaded, compiled and linked, the binary is kept in memory until its pattern instance is removed from the channel.
- When shaders are compiled and linked, the resulting binary is stored in the */resources/shaders/cache* directory.  On subsequent loads, if the shader has not been modified (determined by the timestamp of the .fs file), the precompiled binary is loaded.
- Add *resources/shaders/cache* directory to gitignore so shader binaries won't get copied around.  They're about the least portable thing in existence, pretty much tied to a single machine's GPU, OS, driver version, etc.

To reset your cache completely, erase all the .bin files from /resources/shaders/cache.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ bin/
 target/
 classes/
 Projects/project.lxp
+resources/shaders/cache
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,4 @@ bin/
 target/
 classes/
 Projects/project.lxp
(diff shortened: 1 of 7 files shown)


worldveil:
This greatly improves shader performance! Wooooooot

PR #100, opened 2022-08-05
signal paths: year 1 edge pathing fixes

alextoombs:
**The problem:**

The low priority edges ([according to this tracker](https://docs.google.com/spreadsheets/d/1dQS0hEfm1FXL7tJPlmNtxtZmMggipbxqrDQaaxlxFmo/edit#gid=1994456714)) are unlikely to be completed. So, we won't get signal from any edges that expect a chain from one of these.

**The solution:**

1. Sort the [`edge signal paths` tab here alphabetically](https://docs.google.com/spreadsheets/d/1C7VPybckgH9bWGxwtgMN_Ij1T__c5qc-k7yIhG-592Y/edit#gid=1572666054)
2. Copy build priority in from [the build sheet](https://docs.google.com/spreadsheets/d/1dQS0hEfm1FXL7tJPlmNtxtZmMggipbxqrDQaaxlxFmo/edit#gid=1994456714)
3. Ingest it in ruby; label it on edge objects
4. If an edge is not low priority, but receives signal from an edge, raise error (I actually printed, but changed to raise to sanity check)
5. Move all afflicted edges elsewhere
6. Bonus: squash a few more panel signals to save controllers

I updated the sheets already.

FYI: I preserved the original signal routing path in a hidden tab in the modules sheet called "edge signal paths (COMPLETE FOR YEAR 2)"


tcahill (approved):
Approach seems sound to me

PR #105, opened 2022-08-07
Add "Triangles" and "Width" parameters to PulsingTriangles

tcahill:
- The Triangles parameter controls the number of nested triangles to paint at
once

- The Width parameter controls the thickness of the border of each triangle

triangles

### DIFF
--- a/src/main/java/titanicsend/pattern/tom/PulsingTriangles.java
+++ b/src/main/java/titanicsend/pattern/tom/PulsingTriangles.java
@@ -51,6 +51,14 @@ public class PulsingTriangles extends TEPattern {
             registerColor("Color", "color", ColorType.PANEL,
                     "Color of the triangles");
 
+    protected final BoundedParameter numTriangles =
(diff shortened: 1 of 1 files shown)


src/main/java/titanicsend/pattern/tom/PulsingTriangles.java
```diff
@@ -156,6 +171,8 @@ public void onParameterChanged(LXParameter parameter) {
         } else if (parameter.getPath().equals("tempoLock")) {
             BooleanParameter p = (BooleanParameter) parameter;
             this.phase.tempoLock.setValue(p.getValueb());
+        } else if (parameter.getPath().equals("width")) {
+            pointMap = buildPointMap(model.panelsById);
```
tcahill, line 175: Probably need to throttle this since it's kind of expensive to recalculate as you adjust across a wide range of values.

raldi: Maybe precache a few fixed values

PR #106, opened 2022-08-08
Added Spiral Squares and Random Panel MF64 patterns

zranger1:
A couple more new mini patterns for the MF64

### DIFF
--- a/src/main/java/titanicsend/pattern/mf64/MF64RandomPanel.java
+++ b/src/main/java/titanicsend/pattern/mf64/MF64RandomPanel.java
@@ -1,0 +1,122 @@
+package titanicsend.pattern.mf64;
+
+import heronarts.lx.color.LXColor;
+import titanicsend.model.TEPanelModel;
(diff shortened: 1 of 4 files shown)


src/main/java/titanicsend/pattern/mf64/MF64RandomPanel.java
```diff
@@ -0,0 +1,122 @@
...
+        time.setScale((float) (driver.getTempo().bpm() / 60.0));
+
+        // clear the decks if we're getting ready to stop
+        if (stopRequest) {
+            stopRequest = false;
+            active = false;
+            color = TRANSPARENT;
+        }
+
+        // negative time b/c VariableSpeedTimer has a bug which I'm
```
raldi, line 80: I don't understand this but you've earned my trust

raldi (approved):
There could be some refactoring here to pull common code into a parent class, but .. we're leaving for the playa in a couple weeks so <sound of rubber stamp>

PR #118, opened 2022-08-12
signal paths: recover some channels

alextoombs:
Any low-pri edges that we've built since last week, we can
chain signal of the next edge from the previous channel. This
saves us 4 controllers and any ethernet/power/signal cables
necessary for those.

Follow-up from https://github.com/titanicsend/LXStudio-TE/pull/100

Fixes https://www.notion.so/titanicsend/5359f9e022e643f79676a1f75c4e62b1?v=2aa32b199e9e4db0b7cad8502f8713bd&p=b3eb9221aeb446e2ad96e01cae3de6b5&pm=s

### TODO 

- [x] Run `map_controllers.py` and update the controller output
- [ ] Update controllers tab in the Great Sheet
- [ ] Update `edge_signal_paths` in the Great Sheet


raldi:
When you merge, import controllers.tsv to the corresponding tab, using “replace data at selected cell” with A2 highlighted, tell it to use tab separators, and to convert numbers.

alextoombs:
An update after Jeff's inventory: at 44 controllers now. https://titanicsend.slack.com/archives/C03650K7AE7/p1660342944020329

tcahill (approved):

PR #133, opened 2022-08-16
Updates to AutoVJ

titanicsend:
(this is @worldveil , I am just committing from the LX machine since it's connected to all the hardware)

Eliminates a couple of major bugs:

- Upon master deck change, stale/invalid phrases are no longer triggered
- Palette change timer resets after each newly set palette change

Also cleans up stdout to make logging less verbose.

Note: I am leaving a lot of commented out debugging statements though, just in case anyone else needs to debug this code since bus factor is a little too high here.

### DIFF
--- a/Projects/AutoVJ.lxp
+++ b/Projects/AutoVJ.lxp
@@ -1,6 +1,6 @@
 {
   "version": "0.4.1",
-  "timestamp": 1660603939126,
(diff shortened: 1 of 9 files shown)


worldveil:
Merging this with OK from @raldi

worldveil (approved):

PR #135, opened 2022-08-16
Updated ShowKontrol OSC message formats

worldveil:
- Updated ShowKontrol formats for executors
- Skeleton + flag for sending to Pangolin lasers


titanicsend (approved):
Approving with OK from @raldi

PR #139, opened 2022-08-17
Disable poke button when MF64 not attached

raldi:


titanicsend:
I tested this -- it works!

titanicsend (approved):

PR #140, opened 2022-08-17
Made Jeff report

raldi:


resources/vehicle/controllers-jeff.tsv
```diff
@@ -0,0 +1,194 @@
+25-0	25	0	0	Edge	25-114
```
alextoombs, line 1: These controllers should move to be 25-1 etc right?

raldi: Yep, thanks — forgot to check in that update.

alextoombs (approved):
shipit! also probably update the tab that's now in the god spreadsheet

PR #142, opened 2022-08-17
Sending OSC to Pangolin lasers

titanicsend:
(This is @worldveil committing again from studio machine)

This adds a lasers class which loads from disk the IP address of the lasers. 

Limitations

- Only sends the primary color of the swatch
- Needs the IP address in the .txt file set correctly to function

Also includes updates to the Max patch that forward other data to the lasers:

- OSC beats
- BPM
- Phrase data


worldveil:
@raldi are you able to review?

PR #145, opened 2022-08-17
Fix for OpenGL buffer handle leak

zranger1:
Fixed slow leak of int-sized buffer handles.  This should improve stability over time.

Tested with vehicle.lxp and autovj.lxp, running GL shaders on as many channels as possible.  Still uses a decent sized chunk-o-memory per shader, but usage stabilizes quickly and normal Java garbage collection keeps everything under control without further intervention.

Edit:  Have been running AutoVJ continuously for 4 1/2 hours, with two recorded osc feeds.  No crashes or stalls, memory usage is stable so far.  
osc_test2


.lxpreferences
```diff
@@ -1,6 +1,6 @@
 {
   "version": "0.4.1",
-  "projectFileName": "Debugging.lxp",
+  "projectFileName": "Vehicle.lxp",
```
worldveil, line 3: maybe just change back to debugging for now

worldveil:
Maybe change back the default .lxp to debugging for now, then I'm good to merge this

zranger1:
Fixed!  Thought I'd done that already...

.lxpreferences
zranger1: I'm going to pull this, resync, retest and resubmit.  The number of merge conflicts caused by other checkins just got too high for comfort.

zranger1:
Too many things coming in at once -- going to sync with main, redo the single changed file and resubmit

PR #146, opened 2022-08-17
Fix some compile warnings

ssilverman:
This fixes four types of warnings:
1. Static-member-access warnings
2. Unchecked and raw type generics-related warnings
3. "Redundant cast" warnings
4. A couple deprecation warnings

I added a bunch of reviewers because those reviewers have touched the code being changed here in some way. Please double-check my work.


zranger1 (commented):
Thanks for doing this!   I'm good with merging it once a couple of other people have had a look.

raldi (approved):
Thanks for doing this! I usually hate when a project spews a bunch of warnings and endeavor to get them to zero, but this project got away from me in that regard.

PR #149, opened 2022-08-18
Fix for OpenGL buffer handle leak

zranger1:
**Replaced old PR shown below because github merge conflicts got too weird for safety for a little while there.**  

https://github.com/titanicsend/LXStudio-TE/pull/145

Fixed slow leak of int-sized buffer handles.  This should improve stability over time.

Tested with vehicle.lxp and autovj.lxp, running GL shaders on as many channels as possible.  Still uses a decent sized chunk-o-memory per shader, but usage stabilizes quickly and normal Java garbage collection keeps everything under control without further intervention.

Edit:  Have been running AutoVJ continuously for 4 1/2 hours, with two recorded osc feeds.  No crashes or stalls, memory usage is stable so far.  
osc_test2


worldveil (approved):
@zranger1 tested this for hours and no crashes, so seems safe

PR #150, opened 2022-08-18
Feature/front of house max patch

batchku:
New max patch for front of house controllers, sent to car computers via 1) OSC and 2) Apple's virtual midi network ports setup thru "Audio Midi Setup".  The latter requires that computers sending/receiving "Connect" to one another's "sessions" in "Audio Midi Setup".  The latter just works as long as there is a good network, and the recipient is ready to parts the incoming messages.


jvyduna (approved):

raldi:
@worldveil  Do you know which way to go on these merge conflicts?

PR #152, opened 2022-08-18
Fix generics-related compile warnings

ssilverman:
This corrects a problem from the last attempt.


worldveil:
Appears to work, but @theyoffa should probably be the one to take a look

ssilverman:
I just updated to avoid changes to lines with trailing spaces.

raldi (approved):

PR #158, opened 2022-08-21
Restore AutoVJ.lxp from a fresh backup at each start

jkbelcher:
This is untested!  Please test!  (our environments are different, and I don't know if the relative path is correct or if the app needs to be resigned)

Scenario:
Your application starts AutoVJ.lxp.  Someone makes some changes and hits Command+S.  Now your AutoVJ.lxp is overwritten and incorrect.  The person messed up some stuff and they restart... but it resumes the saved version.

Seems like you want to start from a fresh copy of AutoVJ.lxp every time.  This simple change seems to be the right idea to make that happen.

### DIFF
--- a/Projects/AutoVJ_backup.lxp
+++ b/Projects/AutoVJ_backup.lxp
@@ -1,0 +1,4916 @@
+{
+  "version": "0.4.1",
+  "timestamp": 1660605447161,
+  "model": {
+    "id": 2,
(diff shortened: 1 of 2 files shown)


jkbelcher:
@worldveil This is your territory.

worldveil (approved):
LGTM, works great

PR #159, opened 2022-08-21
Fix all stranded listeners in AutoVJ.lxp

jkbelcher:
This fixes all the shutdown errors from patterns used in AutoVJ.lxp.

@jvyduna No intent to step on your toes here with Fireflies.  Please review.


jvyduna:
No worries at all - it’s not even really my pattern. Thanks for the fixes!

worldveil (approved):
LGTM, tested it after rebuild and errors went away on startup

PR #160, opened 2022-08-21
MF64 fixes and connectivity improvements

jkbelcher:
Fixed stranded listener errors and  improved connection retention and troubleshooting with Poke button:

- Fix double-poke and double-push, only do action on button down not button up
- If device loses connection, inform user on Poke button push
- If program was started without device connected, then cable is plugged in, connection can be established by pressing the Poke button
- Resend surface lights on reconnection (previously, unplug/plugging midi surface would lose lights)
- Restrict pitch and velocity to valid midi ranges (previously, poking >127 threw error)

### DIFF
--- a/src/main/java/titanicsend/pattern/TEMidiFighter64DriverPattern.java
+++ b/src/main/java/titanicsend/pattern/TEMidiFighter64DriverPattern.java
@@ -5,6 +5,7 @@ import heronarts.lx.LXCategory;
 import heronarts.lx.midi.*;
 import heronarts.lx.parameter.BooleanParameter;
 import heronarts.lx.parameter.DiscreteParameter;
+import heronarts.lx.parameter.LXParameterListener;
 import titanicsend.pattern.mf64.*;
 
 @LXCategory("Combo FG")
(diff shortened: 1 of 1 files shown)


src/main/java/titanicsend/pattern/TEMidiFighter64DriverPattern.java
```diff
@@ -160,11 +161,11 @@ public class TEMidiFighter64DriverPattern extends TEPattern implements LXMidiLis
                   .setDescription("Channel number");
 
   public final DiscreteParameter pokeVelocity =
-          new DiscreteParameter("Vel", 0, 256)
+          new DiscreteParameter("Vel", 0, 128)
                   .setDescription("Velocity");
 
   public final DiscreteParameter pokePitch =
-          new DiscreteParameter("Pitch", 0, 256)
+          new DiscreteParameter("Pitch", 0, 128)
```
raldi, line 168: This means it can be 0-127, right?

src/main/java/titanicsend/pattern/TEMidiFighter64DriverPattern.java
```diff
@@ -230,29 +231,39 @@ public TEMidiFighter64DriverPattern(LX lx) {
...
 
-    this.pokeButton.addListener((p) -> {
-      if (this.midiOut == null) {
-        LX.log("No MF64 attached");
+  private final LXParameterListener pokeListener = (p) -> {
+    if (this.pokeButton.isOn()) {
+	  if (this.midiOut == null) {
+        LX.log("No MF64 attached. Checking again...");
+        connect();
+	  } else if (!this.midiOut.connected.isOn()) {
```
raldi, line 258: Maybe remove the "else" so this prints even if the reconnect fails?

raldi (approved):
Cool! Do you actually have access to a MF64? I thought it was already packed on the truck. Either way, this seems good to go; I'd like to have tested it before shipping, but I think we have no choice in this case.

src/main/java/titanicsend/pattern/TEMidiFighter64DriverPattern.java
jkbelcher: Correct. Max value of 127.

src/main/java/titanicsend/pattern/TEMidiFighter64DriverPattern.java
jkbelcher: I was thinking here that Poke is used in this case to probe the connection state.  The connect() method prints its own results, success or failure.  

A question would be: Do you want the poke to send if this was successfully used to establish connection? I was thinking you don't actually care about the note that much, you just want the device connected again.  In the case that the connection was fine, your poke notes get sent.

jkbelcher:
> Cool! Do you actually have access to a MF64?

I do not have access to a MF64, but I have a Midi Fighter 3D (the 16 button version), and I renamed it to Midi Fighter 64 in OSX>Midi Studio to test all this plugging/unplugging behavior.  The button channels are different so I'm not able to confirm if your sendColors() method is correct.  I didn't change any of the channels or note values though.

> I thought it was already packed on the truck.

No idea - I'm in Boise, just thought I'd get you over a couple hurdles before showtime!

raldi:
If you're ready to merge, I still approve.

jkbelcher:
Thanks, done.

PR #162, opened 2022-08-22
Updates to AutoVJ 

worldveil:
- Update palettes for AutoVJ from Tergel
- Fixes to transition time
- Palettize NeonBlocks
- Palettize SlitheringSnake
- Palettize Outrun
- Fix WaterShader speed to be slower since it's the thing that show up automatically
- Fix for selecting random trigger pattern that could cause errors, causing missed transition effects


zranger1 (approved):

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -723,17 +734,9 @@ private void enableFX(boolean isSamePhrase) {
 
         // make new active patterns
         if (prevPhrase != curPhrase) {
-            // only play strobes if it's the first chorus phrase in a row
```
worldveil, line 726: we can delete these lines because I setup the channel to randomly change patterns every few seconds. code is unecessary now

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -723,17 +734,9 @@ private void enableFX(boolean isSamePhrase) {
...
-            LXPattern newStrobesPattern = TEMixerUtils.pickRandomPatternFromChannel(strobesChannel);
```
worldveil, line 727: we can delete these lines because I setup the channel to randomly change patterns every few seconds. code is unecessary now

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -723,17 +734,9 @@ private void enableFX(boolean isSamePhrase) {
...
-            startPattern(strobesChannel, newStrobesPattern);
             TEMixerUtils.setFaderTo(lx, TEChannelName.STROBES, LEVEL_FULL);
         }
 
-        LXPattern newTriggersPattern = TEMixerUtils.pickRandomPatternFromChannel(triggerChannel);
```
worldveil, line 732: we can delete these lines because I setup the channel to randomly change patterns every few seconds. code is unecessary now

src/main/java/titanicsend/app/TEAutopilot.java
worldveil: in the .lxp, I mean

src/main/java/titanicsend/app/TEAutopilot.java
worldveil: in the .lxp, I mean

src/main/java/titanicsend/app/TEAutopilot.java
worldveil: in the .lxp, I mean

PR #166, opened 2022-08-22
GPOutput looks on all channels for TEPattern instead of just the first

jkbelcher:
Not tested, please test.


raldi (approved):
Can't test this at the moment but the code looks good

raldi:
@ssilverman are you available to test this?

ssilverman:
Yes, this evening.

jkbelcher:
Pushed simple fix - forgot to increment the integer and got stuck in infinite While loop.  Still would be good if someone tests the actual output.

jkbelcher:
Proposed #167 as a better alternative.  Leaving this open for now because it would work if you choose.

PR #168, opened 2022-08-23
Reset autopilot after openProject() to fix AutoVJ error

jkbelcher:
This is untested! Please test!

### DIFF
--- a/src/main/java/titanicsend/app/TEApp.java
+++ b/src/main/java/titanicsend/app/TEApp.java
@@ -29,6 +29,7 @@ import java.util.function.Function;
 
 import heronarts.lx.LX;
 import heronarts.lx.LXPlugin;
+import heronarts.lx.LX.ProjectListener.Change;
 import heronarts.lx.parameter.LXParameterListener;
 import heronarts.lx.pattern.LXPattern;
 import heronarts.lx.pattern.color.GradientPattern;
@@ -59,7 +60,7 @@ import titanicsend.pattern.yoffa.media.ReactiveHeartPattern;
 import titanicsend.pattern.yoffa.config.ShaderPanelsPatternConfig;
 import titanicsend.util.TE;
 
-public class TEApp extends PApplet implements LXPlugin  {
+public class TEApp extends PApplet implements LXPlugin, LX.ProjectListener  {
   private TEWholeModel model;
   static public TEWholeModel wholeModel;
 
@@ -97,7 +98,7 @@ public class TEApp extends PApplet implements LXPlugin  {
 
     new LXStudio(this, flags, this.model);
     this.surface.setTitle(this.model.name);
-
+    
     String logFileName = LOG_FILENAME_FORMAT.format(Calendar.getInstance().getTime());
     LX.setLogFile(new File(flags.mediaPath, LX.Media.LOGS.getDirName() + File.separator + logFileName));
   }
@@ -270,6 +271,8 @@ public class TEApp extends PApplet implements LXPlugin  {
 
     GPOutput gpOutput = new GPOutput(lx, this.gpBroadcaster);
     lx.addOutput(gpOutput);
+    
+    lx.addProjectListener(this);
   }
 
   private TEPatternLibrary initializePatternLibrary(LX lx) {
@@ -441,5 +444,12 @@ public class TEApp extends PApplet implements LXPlugin  {
       PApplet.main("titanicsend.app.TEApp", args);
     }
   }
+  
+	public void projectChanged(File file, Change change) {
+		if (change == Change.TRY || change == Change.NEW) {
+			// Clear for file open
+			this.autopilotComponent.autopilotEnabledToggle.setValue(false);
+		} 
+	}
 
 }
--- a/src/main/java/titanicsend/app/TEAutopilot.java
+++ b/src/main/java/titanicsend/app/TEAutopilot.java
@@ -3,6 +3,7 @@ package titanicsend.app;
 import heronarts.lx.LX;
 import heronarts.lx.LXLoopTask;
 import heronarts.lx.Tempo;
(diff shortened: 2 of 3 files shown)


raldi:
I'm not going to be able to test this; it'll be up to Will.

raldi (approved):
I'm still approving though ;)

jkbelcher:
Changed!  Please test extensively!!! big changes!

worldveil:
I'm seeing that transitions now seem broken, though stability seems good. Investigating further.

worldveil:
OK, repro for what I’m seeing

1. open LX, it opens AutoVJ by default
2. switch to Vehicle.lxp
3. switch back to AutoVJ
4. run OSC script

you should see that it never changes patterns on UP, DOWN, or CHORUS channels from the initial starting patterns. it just fades them in and out

that said: deciding to merge this in the meantime so Ali can load this tomorrow. justin’s PR:

- does fix the freezing issue
- works fine when autoVJ is first project loaded

so the workaround here to start AutoVJ is to a) quit LX and then b) restart it using the TE.app

so this is worth getting to Ali before internet potentially goes down out there, but will continue to investigate

src/main/java/titanicsend/app/TEApp.java
```diff
@@ -441,5 +444,12 @@ public static void main(String[] args) {
       PApplet.main("titanicsend.app.TEApp", args);
     }
   }
+  
+	public void projectChanged(File file, Change change) {
+		if (change == Change.TRY || change == Change.NEW) {
+			// Clear for file open
+			this.autopilotComponent.autopilotEnabledToggle.setValue(false);
```
mcslee, line 451: I see nothing that would ever sets this back to true? Is that intended?

src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
```diff
@@ -31,6 +31,10 @@ public static void turnDownAllChannels(LX lx, boolean onlyAffectPhraseChannels)
 
     public static void setFaderTo(LX lx, TEChannelName name, double faderLevel) {
         try {
+        	// Don't crash
+        	if (lx.engine.mixer.channels.size() <= name.getIndex())
+        		return;
```
mcslee, line 36: Probably want to log a NOISY error here... as we're definitely in a bad state if this is the case and wouldn't want to fail silently.

src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
```diff
@@ -72,7 +76,14 @@ public static LXChannel getChannelByName(LX lx, TEChannelName name) {
...
         while (numTries < 8) {
             try {
-                return (LXChannel) lx.engine.mixer.channels.get(name.getIndex());
+            	// JKB mod, now that autopilot loads after everything else,
+            	//  if channel index isn't there then the .lxp isn't AutoVJ and it's not going to show up...
+            	//if (lx.engine.mixer.channels.contains(name))
+            	if (lx.engine.mixer.channels.size() > name.getIndex()) {            		
+            		return (LXChannel) lx.engine.mixer.channels.get(name.getIndex());
+            	} else {
+            		return null;
```
mcslee, line 85: Again, this means state is corrupted for sure. Should at least log something there...

src/main/java/titanicsend/app/TEAutopilot.java
```diff
@@ -805,4 +832,56 @@ else if (newPhrase == TEPhrase.CHORUS)
...
+	
+	private void releaseProjectReferences() {
+		this.prevChannel = null;
+		this.curChannel = null;
+		this.nextChannel = null;
+		this.oldNextChannel = null;
+		this.triggerChannel = null;
+		this.strobesChannel = null;
+        this.fxChannel = null;
+	}
```
mcslee, line 886: I don't think this is sufficient. Looks like at a minimum we need to:
TEPatternLibrary.indexPatterns() once more...

https://github.com/titanicsend/LXStudio-TE/blob/main/src/main/java/titanicsend/app/autopilot/TEPatternLibrary.java#L423

src/main/java/titanicsend/app/TEApp.java
jkbelcher: Yes, intended.  This thing is wired up one directional, UI->Listener->autopilot.  The way autopilot gets turned on from a file load is the UI element boolean gets set, which cascades.  So if this is off, then it will correctly go to the right state from Autopilot=off in a saved file or Autopilot=on in a saved file.   *Note this was the thing I found last with the fixes, so that's why it's a bit redudant to my having disabed autopilot already.  Maybe could have triggered it all from here but I had to put this in or else: a file opening with autopilot=on wouldn't actually fire the enabled(true) startup sequence because the UI boolean listener wouldn't fire.

src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
jkbelcher: I'm checking for null on the return.  The scenario is that someone turned on autopilot and they're aren't enough channels.  The autopilot wants particular channel indexes.  Yeah it's a "problem", but only soft error.  Not hard error, because don't crash the art when some curious VJ clicks autopilot=on to "see what is does" and there's only 2 channels...

src/main/java/titanicsend/app/autopilot/utils/TEMixerUtils.java
jkbelcher: I'm checking for null on the return.  The scenario is that someone turned on autopilot and they're aren't enough channels.  The autopilot wants particular channel indexes.  Yeah it's a "problem", but only soft error.  Not hard error, because don't crash the art when some curious VJ clicks autopilot=on to "see what is does" and there's only 2 channels...

PR #171, opened 2022-08-24
Replace LX patterns that touch gap pixels

zranger1:
#### Replaced the following LX patterns with TE-specific versions:
- NoisePattern => TENoisePattern
- GradientPattern => TEGradientPattern
- SparklePattern => TESparklePattern

The new "TE" versions avoid gap pixels but are otherwise identical to the originals.

#### Replaced heronarts.lx.pattern.xxx with our versions in AutoVJ.lxp and AutoVJ_backup.lxp

(also remove extra log spam from TEWholeModel.java -- I forgot to take it out last night.)

### DIFF
--- a/Projects/AutoVJ.lxp
+++ b/Projects/AutoVJ.lxp
@@ -2641,7 +2641,7 @@
             "patterns": [
               {
                 "id": 9708,
-                "class": "heronarts.lx.pattern.texture.NoisePattern",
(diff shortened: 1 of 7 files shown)


jvyduna:
THANK YOU!

zranger1:
Keeping this branch around in case we want it later, but since there are more core LX components that are going to affect the gap pixels, I recommend using @jkbelcher's NoGapFilter effect for now.

PR #173, opened 2022-08-25
Handling gap pixels

worldveil:
- Makes gap pixels transparent
- Removes gap pixel logging
- Adds NoGap effect to master for all functioning .lxp files


zranger1 (approved):
Looks great!   Off to the playa it goes!

PR #176, opened 2022-08-31
Beacon

jvyduna:
This is Yoffa on Jeff's computer


raldi (approved):
If it works, ship it — but be careful! There are a lot of monkeypatches on lighting-1 and we do NOT want to lose them when we merge this in. Maybe apply the patch there manually. For extra credit, back everything up to lighting-2

titanicsend (approved):
[Jeff] Looks like this isn't selecting top port panels by ID (PUA, PUF) but probably works based on the microns, and it doesn't quite understand what Keegan articulated (to make the **specific set of edges that form the silhouette** always the primary pattern color), but it's just an effect and we can amend or not use it.

PR #182, opened 2023-02-28
Fix for UI not working correctly on Ventura

zranger1:
Modifies justin/venturaFix to use latest official builds of gluegen and jogl from jogamp.org.  

This should be safe to merge, even with the changes from the production machine still pending.  Apart from one line in TEApp.java, it changes only build-related files.


jkbelcher:
I think we can ditch the natives/ folder now that gluegen and jogl are coming from internet repositories.  Tested as working here.  @zranger1 can you test that on Windows?

zranger1:
Working on Windows 10 and Ubuntu on clean builds with natives/ directory gone.   Just pushed the update to this PR.

worldveil (approved):

PR #183, opened 2023-03-14
defined views dj, top, left, right; all have modifiers '*-e' and '*-p'

ctoscano:
Created the following views. 

top-e, top-p, top, dj-e, dj-p, dj, left-p, left-e, left, right-p, right-e, right

Configuration for these views can be seen, modified, and added to here: `resources/vehicle/views.properties`

While I would have liked to use JS, I asked ChatGPT for a convention in Java and it suggested using Properties
`Properties views = new Properties();`

Note: I added geometry.views to hold Properties, and added views to the list of arguments for `TEPanelFactory.build`

The following screenshot shows the `dj` view being applied to a channel showing Neon Cells. Edges are controlled by a different channel. 

image


jkbelcher (approved):
Looks good, tested as working.  Could be a team choice on whether to allow multiple methods of loading settings from files.  I think this .properties method is interesting though and it's educational to show two techniques.

PR #185, opened 2023-03-15
Common DJ Controls Prototype - TEPerformancePattern

zranger1:
Implements general purpose, user configurable common controls for live DJ performance.   All the controls are in, some work on all shader-based patterns, and more will be working everywhere shortly!

**Quick Highlights:**
Inheriting from TEPerformancePattern instead of TEPattern or TEAudioPattern gives you all the features of TEAudioPattern
 our 10 new common controls for live performance a mechanism for modifying or replacing individual controls as needed.

All control ranges are either 0.0 to 1.0, or -1.0 to 1.0, with default initial values
appropriate for the control. 

The controls are implemented via a Java Interface object - the nearest thing Java has to
run-time polymorphism.  You can modify a common control's functionality from your pattern code, or you can replace it completely with your own control, as long as you implement the standard  interface.  (The one exception is the color control, which is just a normal LX color control and can't be modified via the Interface object mechanism.)

`void setControl(TEControlTag tag,LXParameter ctl,_CommonControlGetter getFn)` installs a completely new control and associated "getter" function for the specified control type.
`void SetGetterFunction(TEControlTag tag, _CommonControlGetter getFn)` lets you use a custom getter function for a currently installed control, if you need to scale the output or calculate its value using other inputs.
For example, to make a custom getter function for the speed control, first define it:
```
        _CommonControlGetter myNewGetter = new _CommonControlGetter() {
            @Override public float getValue(LXParameter ctl) { return ctl.getValuef(); }
        }
```
		
Then call `SetGetterFunction(TEControlTag.SPEED, myNewGetter) `to install it, and everything that uses TEPerformancePattern.getSpeed() will call your new function automatically.
To completely replace the speed control, define a new control that can be cast to LXParameter, build a getter as before, and call `setControl()`, like this:
```
            LXParameter p = new CompoundParameter("Speed", 0.5f, 0f, 1.0)
                            .setDescription("New Speed Control");
            setControl(TEControlTag.SPEED,p,myNewGetter);
```
			
Your new control will replace the existing speed control. (Although at present, the controls in the UI may change position.  I can fix this, but it's lower priority for the moment.)			
These are the supported common controls,their associated helper functions, and their LX control paths:

**-  Name		Helper 					LX path**
- Color		int getCurrentColor()        "color"
- Speed		float getSpeed()		     "te_speed"
- xPos		float getXPos()		     "te_xpos"
- yPos 		float getYPos()		     "te_ypos"
- Size 		float getSize()			      "te_size"
- Quantity	float getQuantity()	              "te_quantity"
- Spin		float getSpin()		                "te_spin"
- Brightness	float getBrightness()	        "te_brightness"
- Wow1		float getWow1()			"te_wow1"
- Wow2		float getWow2()			"te_wow2"	
- WowTrigger   boolean getWowTrigger()	"te_wowtrigger"

The following additional handy helpers are also available:
`float getRotationalAngle()` - returns the current "spin" angle in radians, derived from a real-time LFO, the setting of the Spin control, and the value of max_rotations_per_second.
`float getRotationalAngleOverBeat()` - returns a beat-linked "spin" angle in radians derived from getTempo().basis(), the setting of the spin control, and a preset maximum rotations per beat.
`void setMaxRotationsPerSecond(float maxRotationsPerSecond)`
`void setMaxRotationsPerBeat(float maxRotationsPerBeat)`
`void setMaxTimeMultiplier(float m`) - sets the maximum value of speed.  (the parameter m is basically the number of fake seconds per real second. Default is 1. Higher is faster, lower is slower.)
The following uniforms are available to all shaders, preset with values returned from
the common controls where applicable:
```
// standard shadertoy
uniform float iTime;       // this is actually linked to the speed control
uniform vec2 iResolution;
uniform vec4 iMouse;

// TE Audio
uniform float beat;
uniform float sinPhaseBeat;
uniform float bassLevel;
uniform float trebleLevel;

// TE color
uniform vec3 iPalette[5]; // the complete currently active palette
uniform vec3 iColorRGB;   // the color currently showing in the color control
uniform vec3 iColorHSB;

// TE common controls
uniform float iSpeed;
uniform float iScale;
uniform float iQuantity;
uniform vec2  iTranslate;
uniform float iRotationAngle;
uniform float iBrightness;
uniform float iWow1;
uniform float iWow2;
uniform bool  iWowTrigger;

// Shadertoy audio channel + optional textures
uniform sampler2D iChannel0;
uniform sampler2D iChannel1;
uniform sampler2D iChannel2;
uniform sampler2D iChannel3;

```


src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,298 @@
...
+            for (TEControlTag tag : controlList.keySet()) {
+                removeParameter(tag.getPath());
+            }
+            controlList.clear();
+        }
+
+        public void buildDefaultControlList() {
+            LXParameter p;
+
+            p = new CompoundParameter("Speed", 0.1f, -1.0, 1.0)
```
jkbelcher, line 99: Can be 0.1, don't need the "f".  Same for other lines in this method.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,298 @@
...
+    public void setMaxTimeMultiplier(float m) {
+        this.maxTimeMultiplier = m;
+    }
+
+    /**
+     * @return Returns the current rotation angle in radians, derived from a real-time LFO, the setting
+     * of the "spin" control, and the constant MAX_ROTATIONS_PER_SECOND
+     */
+    public float getRotationAngle() {
+        return (float) (TEMath.TAU * (spinTimer.getTime() % 1));
```
jkbelcher, line 231: TEMath.TAU now available as LX.TWO_PI fyi.

src/main/java/titanicsend/pattern/jon/Phasers.java
jkbelcher, line 1: Probably should make a habit of calling super.onActive() from within onActive(), for people who are using these patterns as examples.  Then it's easier to add something to TEPerformancePattern.onActive() later.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,346 @@
...
+
+    public boolean getWowTrigger() {
+        return controls.getValue(TEControlTag.WOWTRIGGER) > 0.0;
+    }
+
+    protected void run(double deltaMs) {
+
+        iTime.setScale(getSpeed() * maxTimeMultiplier);
+        iTime.tick();
+        spinTimer.setScale(getSpin() * maxRotationsPerSecond);
```
jkbelcher, line 341: The use of maxRotationsPerSecond and maxRotationsPerBeat assumes the range is default of -1 to 1, right?  It seems like a pattern might use maxRotationsPer OR the setRange(), but not intentionally both.  If these are alternatives and combining will yield unintended consequences, maybe print a warning to console if both were used?  In which case should maxRotationsPerSecond/Beat be defaulted to 1?

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,346 @@
...
+            setControl(TEControlTag.SIZE, p, defaultGetFn);
+
+            p = new CompoundParameter("Quantity", 0.5, 0, 1.0)
+                    .setDescription("Quantity");
+            setControl(TEControlTag.QUANTITY, p, defaultGetFn);
+
+            p = (CompoundParameter)
+                    new CompoundParameter("Spin", 0, -1.0, 1.0)
+                            .setPolarity(LXParameter.Polarity.BIPOLAR)
+                            .setExponent(2)
```
jkbelcher, line 151: Add .setNormalizationCurve(NormalizationCurve.BIAS_CENTER)

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,346 @@
...
+            }
+            controlList.clear();
+        }
+
+        public void buildDefaultControlList() {
+            LXListenableParameter p;
+
+            p = new CompoundParameter("Speed", 0.1, -1.0, 1.0)
+                    .setPolarity(LXParameter.Polarity.BIPOLAR)
+                    .setExponent(2.0)
```
jkbelcher, line 126: Recommend adding .setNormalizationCurve(NormalizationCurve.BIAS_CENTER) on any parameter that is bipolar and uses setExponent().

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,346 @@
...
+        return controls.getValue(TEControlTag.SIZE);
+    }
+
+    public double getQuantity() {
+        return controls.getValue(TEControlTag.QUANTITY);
+    }
+
+    /**
+     * For most uses, getRotationAngle() is recommended, but if you
+     * need direct acces to the spin control value, here it is.
```
jkbelcher, line 315: access

jkbelcher (changes requested):
Overall looking great.  Made a list of small details.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Changed so everything uses doubles now.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,298 @@
...
+        public interface _CommonControlGetter {
+            float getValue(LXParameter ctl);
+        }
+
+        protected class _CommonControl {
+            _CommonControl(LXParameter ctl, _CommonControlGetter getFn) {
+                this.control = ctl;
+                this.get = getFn;
+            }
+            LXParameter control;
```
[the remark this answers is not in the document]
zranger1: Done

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Using LX constants where available now.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: The maxRotationsPer... values were useful before we had a working SetRange().   As they're no longer needed, I'm just going to remove them.

resources/docs/CommonControls.md
```diff
@@ -0,0 +1,146 @@
...
+
+For example, to make a custom getter function for the speed control, first define it:
+```_CommonControlGetter myNewGetter = new _CommonControlGetter() {
+@Override public double getValue(LXParameter ctl) { return ctl.getValuef(); }
+}
+```
+Then call ```SetGetterFunction(TEControlTag.SPEED, myNewGetter)``` to install it, and everything that
+uses ```TEPerformancePattern.getSpeed()``` will be calling your new function.
+
+To completely replace the speed control, define a new control that can be cast to LXParameter, build a getter,
```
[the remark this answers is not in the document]
zranger1: Fixed in today's commit

resources/docs/CommonControls.md
```diff
@@ -0,0 +1,146 @@
...
+#### WORK IN PROGRESS -- THERE MAY BE USEFUL STUFF HERE, BUT THERE IS STILL A LOT OF CHAOS
+#### COMING SOON!!!    LESS CHAOS!!!
+
+
+
+## The Super Quickstart Version:
+Derive your pattern's class from TEPerformancePattern (This is already done for native shader patterns
+made using ConstructedPattern)
+
+Use the ```control.setXXX(TECommonControlTag.xxx,value(s)...)``` methods to configure the controls
```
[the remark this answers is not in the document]
zranger1: Fixed in today's commit

resources/docs/CommonControls.md
```diff
@@ -0,0 +1,146 @@
...
+
+To use TEPerformancePattern, derive your pattern class from it rather than
+one of the other TE pattern classes.  Creation and initialization of the common
+controls is handled automatically by the constructor.
+
+## List of Common Controls
+The common controls, their associated helper functions, and the tags used to get and
+set control values are:
+Name		Helper 					tag                     
+Color		int getCurrentColor()   <none>
```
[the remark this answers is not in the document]
zranger1: Gotta love formatting text!  Fixed in today's commit

src/main/java/titanicsend/pattern/jon/Phasers.java
zranger1: Fixed today, for all the patterns I've written in this hybrid Java/Shader style

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Replaced with much nicer getRotationAngleFromSpeed() and getRotationAngleFromSpin() functions.  These functions base their rotation rates on the engine's current bpm, and are controlled respectively by the Speed and Spin controls, so you can have two independent, beat-linked rotations going without worrying about the calculations.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Done, with an abundance of caution, for all bipolar parameters.   It's pretty likely users might change the exponent without thinking about the normalization curve, so let's just fix it for them in advance!

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Fixed in today's commit

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Yikes!  Fixed.

resources/docs/CommonControls.md
```diff
@@ -0,0 +1,146 @@
...
+
+## The Super Quickstart Version:
+Derive your pattern's class from TEPerformancePattern (This is already done for native shader patterns
+made using ConstructedPattern)
+
+Use the ```control.setXXX(TECommonControlTag.xxx,value(s)...)``` methods to configure the controls
+as needed
+In your pattern java code, use the helper methods in TEPerformancePattern (```getSpeed()```, ```getSpin()```, etc.) to get values
+from the controls.
+If your pattern has a shader, use the uniform variables listed below to get control values.
```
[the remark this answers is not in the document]
zranger1: Fixed - this should look better now.

resources/docs/CommonControls.md
```diff
@@ -0,0 +1,146 @@
...
+### Additional handy helpers:
+```double getRotationalAngle()``` - returns the current "spin" angle in radians, derived from a real-time
+LFO, the setting of the Spin control, and the value of max_rotations_per_second.
+
+```double getRotationalAngleOverBeat()``` - returns a beat-linked "spin" angle in radians derived from getTempo().basis(),
+the setting of the spin control, and a preset maximum rotations per beat.
+
+```void setMaxRotationsPerSecond(double maxRotationsPerSecond)```
+```void setMaxRotationsPerBeat(double maxRotationsPerBeat)```
+111void setMaxTimeMultiplier(double m)``` - sets the maximum value of speed.  (the parameter m is basically the number of
```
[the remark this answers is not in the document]
zranger1: Doh!  Fixed.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,346 @@
...
+    protected TEPerformancePattern(LX lx) {
+        this(lx, true);
+    }
+
+    public void registerColorControl() {
+        this.controls.registerColorControl();
+    }
+
+    public FloatBuffer getCurrentPalette() {
+        int col;
```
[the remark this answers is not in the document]
zranger1: Fixed.  Might as well get it now!

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,346 @@
...
+                registerColorControl();
+            }
+
+            // Add the user replaceable controls
+            buildDefaultControlList();
+            //registerCommonControls();
+        }
+    }
+
+    protected double maxRotationsPerSecond;
```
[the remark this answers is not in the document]
zranger1: Got rid of these guys.  No longer needed.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -54,11 +54,7 @@ protected double getValue(TEControlTag tag) {
 
         public TECommonControls setControl(TEControlTag tag, LXListenableParameter lxp, _CommonControlGetter getFn) {
```
jkbelcher, line 55: Can we overload this method with a version that takes only the first two parameters and uses defaultGetFn?

jkbelcher (changes requested):

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Certainly can!  Done, and added to the documentation.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -58,6 +58,12 @@ public TECommonControls setControl(TEControlTag tag, LXListenableParameter lxp,
             return this;
         }
 
+        public TECommonControls setControl(TEControlTag tag, LXListenableParameter lxp) {
```
jkbelcher, line 61: Yes, except to avoid code reuse the content of this method can just be:
`return setControl(tag, lxp, defaultGetFn);`

jkbelcher (changes requested):

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: sigh... meant to do that.  Got distracted by farm chores after the initial cut/paste, I guess.

It's in now.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher: Nice!

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -0,0 +1,347 @@
...
+
+    /**
+     * @return Returns a loosely beat-linked rotation angle in radians.  Overall speed
+     * is determined by the "Spin" control, but will automatically speed up and slow down
+     * as the LX engine's beat speed changes.
+     * If current speed is zero, returned angle will also be zero, to allow easy reset of patterns.
+     */
+    public double getRotationAngleFromSpin() {
+        // See comments in getRotationAngleFromSpeed() above.
+        return (getSpin() != 0) ? (LX.TWO_PI*(lx.engine.tempo.bpm()/60) * spinTimer.getTime()) % LX.TWO_PI : 0;
```
jkbelcher, line 287: If the VariableSpeedTimer has been running for a long time and the engine bpm slows down, I think this will cause the calculated position to leap ahead.  It might be necessary to track absolute rotational position underneath, rather than elapsed time.

jkbelcher (changes requested):

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher: To save you reinventing the wheel here, what’s worked in my own base class is to add a parameter “Angle”, do setWrappable(true) on it, and increment it each frame with angle.incrementNormalized(amount).  Then angle can be adjust both manually and with the spin parameter.

zranger1: I just checked in a version of this that uses a lightweight "Rotor" class to do the calculation incrementally.   It should be stable across beat speeds and extended run times.  

I'm interested in using LX to do this - it could certainly function as the accumulator, and is likely more flexible in the long run - but couldn't figure out exactly which Parameter type would be most sensible to use.   This is one of those things there's not a lot of documentation on at the moment.

jkbelcher: CompoundParameter would work fine for the accumulator.

zranger1: Ok, the getRotationAngleFromX thing is now greatly simplified and working.   There is a setAutospin(true/false) function to control automatic rotation, and Outrun Grid demonstrates how to use the Spin control to directly control rotation w/no autospin.

Having played with this for a while, I like how it gives us some extra beat reactivity for free -- I'm going to have a go at linking the Speed control in general to the engine bpm and see how it looks.

This could be quite helpful for shadertoy shaders, most of which do their thing as a function of time, but aren't directly beat reactive.  If we can feed them a time value that has some correspondence to beat, suddenly they have a whole different, better feel, without us having to change any code.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher, line 1: Adjusting the spin dial across center causes a visual jump as the rotor/angle snaps back to a fixed position.  Centering the Spin parameter should stop the rotation in its current place but not cause a visible jump.

jkbelcher, line 1: Turning the Angle parameter to the right should rotate the pattern in the same direction as turning the Spin parameter to the right.  Currently they have opposite effects.  Currently Spin looks correct: turning to the right creates clockwise motion on Phasers and SpiralDiamonds.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -192,6 +192,13 @@ public void buildDefaultControlList() {
                     .setMode(BooleanParameter.Mode.MOMENTARY)
                     .setDescription("Trigger WoW effects");
             setControl(TEControlTag.WOWTRIGGER, p);
+
+            // in degrees for display 'cause more people think about it that way
+            // the associated getAngle() function converts to radians.
```
jkbelcher, line 197: Might be clearer for pattern developers if getAngle() and the angle parameter use the same units.  You can modify the user-facing display value like this:
```
  .setFormatter((v) -> {
    return Double.toString(Math.toDegrees(v));
  })
  ;
```

jkbelcher (changes requested):
Including an Angle parameter is good.  Need to tweak Spin to remove jumpiness.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Can fix easily - just an artifact of the way reset is handled.

Question: When you hit reset on spin, do you want it to just stop, or to snap back to the default angle?  I can make a case for either, but since you can reset in performance by pressing the knob, it might be best to just make it a handy way to stop spinning.   You've done this live, so I'll take your guidance on what feels best for the VJ.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -192,6 +192,13 @@ public void buildDefaultControlList() {
...
+            p = new CompoundParameter("Angle", 0, -180, 180)
+                    .setDescription("Static Rotation Angle")
+                    .setWrappable(true);
```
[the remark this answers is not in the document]
zranger1: Will be in next commit.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher: >  When you hit reset on spin, do you want it to just stop, or to snap back to the default angle?
Just stop spinning.  I would expect, at least when not spinning, that hitting reset on Angle snaps to the default angle.

jkbelcher: > Will make sure Angle reset does the right thing shortly

If you can override just the reset without it jumping as it goes through zero, then great.  If that gets funky then the current state is usable for now.

zranger1: Yeah, worst case, I have to subclass the Angle control and override reset.  Looking at the LX source, this seems straightforward.  It'd be nice to have that "go back to a good looking angle" functionality.

zranger1: Ok, the Angle control now returns to the previously set angle if reset while not spinning, and has no effect while spin is active.   Also Outrun Grid has some improvements, and I've added this round of new stuff to the documentation.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -15,6 +15,31 @@
...
+
+        @Override
+        public LXParameter reset() {
+            System.out.println("Reset1");
+
+            // if not spinning, resetting angle control clears
+            // the spin angle and leaves the display rotation angle at the
+            // current static setting.
+            if (getSpin() == 0) {
+                spinRotor.setAngle(0);
```
jkbelcher, line 35: Currently not resetting.  Looks like this also needs `super.reset();` right here

jkbelcher (changes requested):

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: That's by design atm.  If you call super.reset() it sets the value back to zero, which is not the behavior we talked about.

If the pattern has been "spun" to a random rotation, spin is zero, and angle is non-zero, reset will set the rotation back to the current setting of angle.   I thought that's what we wanted.

zranger1: Changed Angle reset behavior.  Here's what happens when you reset the Angle control:

- if not spinning, resets both the static angle and the spin angle to 0.
- If spinning, resets static angle to 0, and adds a corresponding offset to spinRotor to avoid a visual glitch in the spin.

jkbelcher: Nice!

jkbelcher (approved):

PR #188, opened 2023-03-28
Added brightness to SpiralDiamonds

worldveil:
Uses @zranger1 's common control code to implement brightness for SpiralDiamonds pattern as a proof of concept for how to do this.

### DIFF
--- a/Projects/will.lxp
+++ b/Projects/will.lxp
@@ -1,6 +1,6 @@
 {
   "version": "0.4.1",
-  "timestamp": 1657912507447,
+  "timestamp": 1680042256531,
   "model": {
     "id": 2,
     "class": "heronarts.lx.structure.LXStructure",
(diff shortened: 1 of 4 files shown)


zranger1 (commented):
Looks good!  No objection to changing the Brightness control's default range to 0 - 100.

If you do though, please also fix the call to ```ctl.getBrightness()``  at around line 252 in NativeShader.java to divide it back into the 0 - 1 range for OpenGL.  That'll make sure all the shaders keep working.  (And I'll make a note to expose "getNormalizedXXX" APIs so we'll not have to worry about this again, no matter what we do to the ranges.)

worldveil:
@zranger1 done!

zranger1:
Instead of multiplying in the API, can you change the range in the default control setting to ```CompoundParameter("Brightness", 100, 0, 100)```?   (In buildDefaultControlList in TEPerformancePattern) That way, Java will be happy, the interface will show an easy-to-understand number, and we'll only have to divide once, when sending to the shader.

zranger1 (approved):

PR #189, opened 2023-03-31
Brightness rework - Now built into common color control

zranger1:
The Brightness control is now automatically used to adjust the output of the color control, so pattern developers do not have to worry about multiplying color by brightness in their patterns.  If you get color from the common control, it will just work.  This applies to both Java and GLSL.

(Also sneaking in the ArcEdges pattern - controls are definitely not tuned yet, but I'd like to get some feedback on how it runs on a Mac.  If it doesn't blow up because I've missed removing some nvidia only feature, I'd expect M1/M2 machines to be fine.  Might slow down the Intel models a bit.)

### DIFF
--- a/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -445,13 +445,12 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     }
 
     /**
-     * NOTE: This control has functional overlap with color and channel fader settings, and
-     * could potentially can confusing brightness behavior.
+     * <b>NOTE:</b> This control has functional overlap with color and channel fader settings, and
+     * could potentially cause confusing brightness behavior.
      * <p></p>
-     * It may be deprecated in the future and at present, should only be used as a final multiplier
-     * for color values as the last step in a pattern.
+     * <b>It may be deprecated or removed in the future and should not be used in patterns.</b>
      *
-     * @return The current value of the brightness control, by default in the range 0-100.
+     * @return The current value of the brightness control, by default in the range 0.0 to 1.0
      */
     public double getBrightness() {
         return controls.getValue(TEControlTag.BRIGHTNESS);
--- a/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -373,10 +373,45 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
         return controls.getValue(TEControlTag.ANGLE);
     }
 
+    /**
+     * @return Color derived from the current setting of the color and brightness controls
+     * <p></p>
+     * NOTE:  The design philosophy here is that palette colors (and the color control)
+     * have precedence.
+     * <p></p>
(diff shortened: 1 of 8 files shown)


src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -409,10 +444,17 @@ public double getSpin() {
...
     }
 
-    // multiply by 100 so that LXColor.hsb( ... ) and similar inputs can use it, since
-    // LX expects brightness in a range of 0-100
+    /**
+     * NOTE: This control has functional overlap with color and channel fader settings, and
+     * could potentially can confusing brightness behavior.
+     * <p></p>
+     * It may be deprecated in the future and at present, should only be used as a final multiplier
+     * for color values as the last step in a pattern.
```
jkbelcher, line 452: *Should *not* be used by patterns, right?

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -373,10 +373,45 @@ public double getStaticRotationAngle() {
...
+     * the color control.
+     */
     public int getCurrentColor() {
+        int k = controls.color.calcColor();
+        float bri = (float) getBrightness();
+
+        float r = (float) (0xff & LXColor.red(k)) * bri;
+        float g = (float) (0xff & LXColor.green(k)) * bri;
+        float b = (float) (0xff & LXColor.blue(k)) * bri;
+        return LXColor.rgb((int) r,(int) g,(int) b);
```
jkbelcher, line 397: Is scaling RBG values safe or does it need to be converted to HSB color space?

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Yes.  We should be fairly emphatic about this.  Updated javadoc comment in latest commit:

    /**
     * NOTE:  This control has functional overlap with color and channel fader settings, and
     * could potentially cause confusing brightness behavior.
     * 
     * It may be deprecated or removed in the future and should not be used in patterns.
     *
     * @return The current value of the brightness control, by default in the range 0.0 to 1.0
     */

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher: Determined OK - discussion in slack.

jkbelcher:
ArcEdges renders fine on M2 but is max load.  When targeting 60fps actual fps jumps around between 50-70 and channel stays red.

jkbelcher (approved):

zranger1:
Re ArcEdges performance: that's about what I expected.  There are some simple optimizations I can do to reduce the GPU load for now, and an engine change that we may want to do before Burning Man that will make all the OpenGL stuff go a lot faster.  

(The bottleneck here is transfer from CPU to GPU memory.  There's a lot of per-frame stuff going on right now that can be done once, or way less frequently in the engine.  Just takes time to test and get it right.)

Gotta say though, it's amazing and cool that we have small, light laptops that can do this sort of rendering, even if it stretches them a little.

PR #190, opened 2023-04-01
Update Outrun Grid shader to use common controls

zranger1:
Outrun Grid now supports all the controls:

The globally supported controls (speed,spin,xpos,ypos, etc.) have not been changed.  They work as before.  Here's what the new ones do:

- Quantity:  Glow level
- Wow1: Curving paths
- Wow2: Bright pulses down the Z-axis on the beat
- Wow Trigger:  Boat Mode!  

Also, palette support is in, and the alpha channel now supports proper blending.

### DIFF
--- a/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -171,7 +171,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
             p = new CompoundParameter("Speed", 0.5, -4.0, 4.0)
                     .setPolarity(LXParameter.Polarity.BIPOLAR)
                     .setNormalizationCurve(BoundedParameter.NormalizationCurve.BIAS_CENTER)
-                    .setExponent(1.0)
+                    .setExponent(1.75)
                     .setDescription("Speed");
             setControl(TEControlTag.SPEED, p);
 
--- a/resources/shaders/outrun_grid.fs
+++ b/resources/shaders/outrun_grid.fs
@@ -2,13 +2,24 @@
 #define TAU PI * 2.0
 #define DEG2RAD PI/180.
 
-// Prevents flickering
-#define SUPERSAMP 8
+float horizonY = 0.15;
+float fov = 90.0;
+float farClip = -16.0;
+float boltSize = 0.25;
 
 //  rotate a point around the origin by <angle> radians
 vec2 rotate(vec2 point, float angle) {
   mat2 rotationMatrix = mat2(cos(angle), -sin(angle), sin(angle), cos(angle));
   return rotationMatrix * point;
+}
+
+// polynomial smooth min/max from
+// https://iquilezles.org/articles/smin/
+// (works as max if you invert the signs of a and b.)
+// k is the level of smoothing.  k = 0.1 is a good starting place
+float smin( float a, float b, float k ) {
+    float h = max( k-abs(a-b), 0.0 )/k;
+    return min( a, b ) - h*h*k*(1.0/4.0);
 }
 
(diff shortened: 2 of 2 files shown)


jkbelcher (commented):
These effects are cool.  Comments on their assignments to common controls:

-Wow1/Curving Paths: works great, can ease in and out of it.
-Bright Pulses on Beat: Great implementation of beat reactivity. Could imagine wanting the pulses to go either direction. (Although just remembered: for now we can't be changing the biPolar setting from one pattern to the next, it will trigger a MFT reboot which makes the controller unusable for a couple seconds. Will be fixed someday.)
-Boat mode: Cool, but triggering it causes a visual jump so it couldn't be used live.  And on the momentary switch would have to be held On the entire time it was running.
-Glow level: Also cool.  Doesn't instinctively correlate to Quantity.  Would have been a good fit for brightness if we hadn't just retracted independent interpretation of brightness.
-Speed: Could use exponent to create more finesse in the slow speeds.

If I had to guess at the control functions without testing them, the guess would be:
-Size changes width of the grid lines
-Quantity changes the number of grid squares visible on the screen, which is approximately what Size currently does.

zranger1:
Round 1 of Tuning Fixes:

- "Boat Mode" now eases in/out at higher levels of **Wow1**, along with the rest of the curvature
- Line width/glow are on the **Size** control
- Number of visible lines (zoom level) is on **Quantity**.  
- **Wow Trigger** now inverts color  (looks especially cool if you also turn up **Wow1** and **Wow2** a bit.)
- **Speed**'s response curve exponent is now 1.75.  I played with this on my hardware 'till I found a balance between good control and too much flat spot in the middle.  Let me know if it needs to be different for the midifighters.

jkbelcher:
All these mods look good!  Love the appearance of boat mode at the top of Wow1.  WowTrigger is cool.  Tuning looks good for now, expecting a round of minor tweaks pre-show.

> (And we're clearly gonna need more controls!)
You can see how this is a slippery slope to 4 controllers.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -171,7 +171,7 @@ public void buildDefaultControlList() {
             p = new CompoundParameter("Speed", 0.5, -4.0, 4.0)
                     .setPolarity(LXParameter.Polarity.BIPOLAR)
                     .setNormalizationCurve(BoundedParameter.NormalizationCurve.BIAS_CENTER)
-                    .setExponent(1.0)
+                    .setExponent(1.75)
```
jkbelcher, line 174: Exponent curves will vary by pattern, so this call should probably be in the OutrunGrid pattern constructor.

jkbelcher (changes requested):

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: I'm kind of inclined to leave this as the default curve for speed.   All the shader patterns use the same speed scaling code, inherited from the common controls, and they've all had the "hard to control low speeds" problem.

Also, these things mostly don't have accessible individual constructors -- @theyoffa built a really cool automatic mechanism that lets pattern authors just drop in GLSL shaders.   Except for my crazy hybrid patterns, all the native shader patterns use it. 

On the Java, new patterns like Spiral Diamonds uses the common speed code too, and behave better with this curve.  If we can't make a pattern use the new speed mechanism while we're updating, chances are it's Java from last year, and it'll have a constructor that will let us easily change the curve as needed.

jkbelcher: Ok sounds good.

jkbelcher (approved):

PR #192, opened 2023-04-08
Merge in updated LX snapshot

theyoffa:
Update LX lib to higher version with performance mode. Contains a few minor compatibility fixes. See email thread for more details.


src/main/java/titanicsend/pattern/pixelblaze/PixelblazePattern.java
jkbelcher, line 1: I'm getting an error on PixelblazePattern line 38 (looks like thrown by `clearPixels()` on line 35).
`Error updating points:Cannot store to int array because "this.colors" is null
`
Looks like similar situation as noted by mcslee on TEPattern constructor: colors array no longer exists at instantiation time.  This is happening with AutoVJ.lxp, so it might be getting tripped when `enableEdges` or `enablePanels` is loaded from file.

jkbelcher (changes requested):

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePattern.java
zranger1: Just commenting out the call to ```clearPixels()``` cleans up this error and has no other effect that I can see.  Doesn't affect rendering in any way.

zranger1:
Running this with AutoVJ enabled causes a hard hang (requiring power cycle) on Windows after about a minute.  100% repeatable so far.  

Runs normally with AutoVJ disabled, and I can't replicate the crash VJ'ing it manually -- fading channels, switching patterns, turning modulators on/off, etc.    Possibly I/O related?

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePattern.java
jkbelcher: Removing `clearPixels()` left stranded lit pixels after turning off the `enableEdges` or `enablePanels` parameters.  Just pushed a commit changing this this to a flag and doing the `clearPixels()` call from `runTEAudioPattern()`.

That cleans it up for now.  In the long run render zones can be controlled by views.  Once patterns are honoring views correctly these two parameters can be retired.

jkbelcher (approved):

PR #194, opened 2023-04-16
Pattern Rework 1 - Pixelblaze (and others)

zranger1:
- ```PixelblazePattern``` (javascript) &  ```PixelblazePort``` (java) patterns now support common controls.
- Some high priority Pixelblaze patterns, plus several others, have been refined and modified to work with the new controls.

Details:

**AudioTest2 (Spectrum analyzer pattern)** 
- Speed: no effect
- X/Y Offset: Yes
- Spin/Angle:  Yes, but use gently.
- Brightness: Yes
- Size: Controls overall scale
- Quantity: controls number of bands (repeated, not subdivided.)
- Wow1: Fades waveform display.  (leaving spectrum analyzer)
- Wow2: Color.  More Wow2 == more colors, more shine (and less palette compliance)
- WowTrigger: Solos waveform display in current primary color

**Phasers:**
- Speed: controls beam generator spin rate
- X/Y Offset: Yes
- Spin/Angle: Rotates canvas
- Brightness: Yes
- Size: Beam width and focus
- Quantity:  Number of beams generated by beam splitter
- Wow1: Beat reactive beam waviness.  Try it!
- Wow2: Fog intensity
- Wow Trigger: Beat reactive sweep & big flash

**Audio1:  (Use this rather than PBAudio1)**
- Color: Uses current primary-secondary gradient
- Speed: Yes
- X/Y Offset: Yes
- Spin/Angle: No
- Brightness: Yes
- Size: Scale of noise background
- Quantity: Amount of beat reactive pulse
- Wow1: Noise resolution
- Wow2: No
- WowTrigger: No

**Xorcery: (Use this rather than PBXorcery)**
- Color: Uses current primary-secondary gradient
- Speed: Yes - pattern movement and direction.  Very sensitive
- X/Y Offset: Yes, but not really useful
- Spin/Angle: Yes.
- Brightness: Yes
- Size: Scale pattern
- Quantity: No
- Wow1: Slow, beat linked "breathing" effect
- Wow2: Pattern detail modulation
- WowTrigger: No

**Follow That Star:**
- Speed: Controls star movement rate
- X/Y Offset: Yes
- Spin/Angle: Star spin speed
- Brightness: Yes
- Size: Size of stars
- Quantity:  Number of stars
- Wow1: Stars flash and dance to the beat
- Wow2: Overall glow level of stars
- Wow Trigger: No

### DIFF
--- a/resources/pixelblaze/glue.js
+++ b/resources/pixelblaze/glue.js
@@ -138,6 +138,73 @@ function getTrebleRatio() {
   return __pattern.getTrebleRatio();
 }
 
+// getter functions for the common controls
+function getRotationAngleFromSpeed() {
+    return __pattern.getRotationAngleFromSpeed();
+}
+
+function getRotationAngleFromSpin() {
+    return __pattern.getRotationAngleFromSpin();
+}
+
+function getStaticRotationAngle() {
+    return __pattern.getStaticRotationAngle();
+}
+
+function getCurrentColor() {
+    var k = __pattern.controls.color.calcColor();
+    var bri = __pattern.getBrightness();
+
+    var r = (0xff & LXColor.red(k)) * bri;
+    var g = (0xff & LXColor.green(k)) * bri;
+    var b = (0xff & LXColor.blue(k)) * bri;
+    return rgb(r, g, b);
+}
+
+function getTime() {
+    return __pattern.getTime();
+}
+
+function getSpeed() {
+    return __pattern.getSpeed();
+}
+
+function getXPos() {
+    return __pattern.getXPos();
+}
+
+function getYPos() {
+    return __pattern.getYPos();
+}
+
+function getSize() {
+    return __pattern.getSize();
+}
+
+function getQuantity() {
+    return __pattern.getQuantity();
+}
+
+/**
+ * As in Java, for for most uses, getRotationAngle() is recommended, but if you
+ * need direct access to the spin control value, here it is.
+ */
+function getSpin() {
+    return __pattern.getSpin();
+}
+
+function getWow1() {
+    return __pattern.getWow1();
+}
+
+function getWow2() {
+    return __pattern.getWow2();
+}
+
+function getWowTrigger() {
+    return __pattern.getWowTrigger();
+}
+
 /* Pixelblaze compatibility framework glue */
 
 function sentenceCase(text) {
@@ -202,12 +269,14 @@ function glueRender() {
   } else {
     r = render;
   }
+  var xOffs = __pattern.getXPos();
+  var yOffs = -__pattern.getYPos();
   var i;
   for (i = 0; i < __points.length; i++) {
     __color = 0;
     point = __points[i];
     if (point == __pattern.model.gapPoint) continue;
-    r(i, point.xn, point.yn, point.zn);
+    r(i, point.xn + xOffs, point.yn + yOffs, point.zn);
     __colors[point.index] = __color;
   }
 }
--- a/resources/shaders/phasers.fs
+++ b/resources/shaders/phasers.fs
@@ -1,14 +1,22 @@
-uniform float energy;
-uniform float iCanvasAngle;
+
+
+// Controls
+// Speed - beam rotation speed
+// Spin/Angle - canvas rotation
+// Quantity - beam splitter / number of beams
+// Size - beam width / focus
+// iWow1 - beat reactivity
+// iWow2 - glow / fog level
+// WowTrigger - center beam and strobe blast
+uniform float iCanvasAngle = 0.0;
+uniform float beamWidth;
 
 const float PI = 3.141592653589793;
 const float halfpi = PI / 2.;
-const float twopi = PI * 2.;
+const float TAU = PI * 2.;
 const float xPos1 = -0.285;
 const float xPos2 = 0.315;
 const float yCenter = 0.3;
-
-#define glow iWow2
 
 // accumulator for "Wow Trigger" effect
 float zapper = 0.0;
@@ -19,26 +27,13 @@ vec2 rotate(vec2 point, float angle) {
   return rotationMatrix * point;
 }
 
-// 2D (square) Domain repeats
-vec2 repeat2D(vec2 p, vec2 size) {
-	return mod(p + size*0.5,size) - size*0.5;
-}
-
-// polar domain repeats around coordinate origin
-// (looks like pie slices on the car)
-void modPolar(inout vec2 p, float repetitions) {
-	float angle = twopi/repetitions;
-	float a = atan(p.y, p.x) + angle/2.;
-	a = mod(a,angle) - angle/2.;
-	p = vec2(cos(a), sin(a))*length(p);
-}
-
-// 2D circle distance function
+// 2D circle signed distance function
  float circle(vec2 p, float radius){
-	return radius - length(p);
+	return length(p) - radius;
 }
 
- // rand [0,1]
+ // Probably not really that random but its
+ // distribution, whatever it be, looks nice for fog.
  float rand(vec2 p) {
  	return fract(sin(dot(p, vec2(12.543,514.123)))*4732.12);
  }
@@ -58,7 +53,7 @@ void modPolar(inout vec2 p, float repetitions) {
  float fbm(vec2 p) {
      float a = 0.5;
      float r = 0.0;
-     for (int i = 0; i < 7; i++) {
+     for (int i = 0; i < 6; i++) {
          r += a*noise(p);
          a *= 0.5;
          p *= 2.0;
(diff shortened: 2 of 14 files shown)


src/main/java/titanicsend/pattern/ben/Xorcery.java
```diff
@@ -61,10 +71,19 @@ public void render3D(int index, float x, float y, float z) {
...
-		t2 = time(.1) * PI2;
-		t3 = time(.5);
-		t4 = time(.34) * PI2;
+		t1 = 1.0 - time(.1);
+		t2 = Math.sin(t1 * PI2);
+		t3 = triangle(time(.5));
+		t4 = Math.sin(time(.34) * PI2);
+
+		scale = getSize();
+		breath = 0.05 + 0.3 * getWow1();
```
jkbelcher, line 80: It would be nice to be able to turn this effect fully off when Wow1 is down.  The appearance/disappearance of a movement is a good change to be able to make to highlight a change in the music.
As commented on another file - the pattern constructor could initialize a value while keeping the default at 0.

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
jkbelcher, line 1: Spin/Angle feel backward here too.  Expecting clockwise knob = clockwise rotation.  This is a nice variation though that the objects are rotated instead of the final rendered canvas.

jkbelcher, line 1: Again here: would be nice if Wow1 effect can be turned fully off.

jkbelcher, line 1: Wow1 creates a big change when size is 25%.  But with size at maximum there's no perceptible difference between low and high Wow1.

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePattern.java
```diff
@@ -7,12 +7,13 @@
 import heronarts.lx.parameter.LXParameter;
 import heronarts.lx.parameter.LXParameterListener;
 import titanicsend.pattern.TEAudioPattern;
```
jkbelcher, line 9: No longer needed.

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePort.java
```diff
@@ -5,13 +5,14 @@
 import heronarts.lx.parameter.BooleanParameter;
 import heronarts.lx.parameter.LXParameterListener;
 import titanicsend.pattern.TEAudioPattern;
```
jkbelcher, line 7: import TEAudioPattern no longer needed

src/main/java/titanicsend/pattern/pixelblaze/Wrapper.java
```diff
@@ -13,6 +13,7 @@
 import org.openjdk.nashorn.api.scripting.JSObject;
 import org.openjdk.nashorn.api.scripting.NashornScriptEngineFactory;
 import titanicsend.pattern.TEAudioPattern;
```
jkbelcher, line 15: import TEAudioPattern no longer needed

resources/pixelblaze/glue.js
```diff
@@ -138,6 +138,73 @@ function getTrebleRatio() {
...
+
+function getRotationAngleFromSpin() {
+    return __pattern.getRotationAngleFromSpin();
+}
+
+function getStaticRotationAngle() {
+    return __pattern.getStaticRotationAngle();
+}
+
+function getCurrentColor() {
```
jkbelcher, line 154: Can the contents of this method be simplified into `__pattern.getCurrentColor()`, which scales for brightness?

src/main/java/titanicsend/pattern/pixelblaze/Wrapper.java
```diff
@@ -106,13 +107,13 @@ public void load() throws IOException, ScriptException, NoSuchMethodException {
...
       return;
-    bindings.put("__now", System.currentTimeMillis());
+    bindings.put("__now", pattern.getTime() * 1000.0);
     bindings.put("__points", points);
     bindings.put("__colors", colors);
 
     JSObject glueBeforeRender = (JSObject) bindings.get("glueBeforeRender");
     if (glueBeforeRender != null)
-      glueBeforeRender.call(null, deltaMs, System.currentTimeMillis(), points, colors);
+      glueBeforeRender.call(null, deltaMs, pattern.getTime() * 1000.0, points, colors);
```
jkbelcher, line 116: When VariableSpeedTimer divides the time by 1000 and saves to variable, and this line retrieves the variable and multiplies by 1000, is resolution lost?

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePort.java
```diff
@@ -73,14 +77,24 @@ public void runTEAudioPattern(double deltaMs) {
...
 	}
 
+	/**
+	 * Called before adding common controls. Derived classes can override this to
+	 * change control ranges, response curves and other parameters if necessary.
+	 * <p></p>
+	 * If a pattern needs to add additional custom controls, it can do so by
+	 * implementing them in setup().
+	 */
+    public abstract void configureControls();
```
jkbelcher, line 97: Alignment?

src/main/java/titanicsend/pattern/pixelblaze/PixelblazeParallel.java
```diff
@@ -3,12 +3,13 @@
 import heronarts.lx.LX;
 import heronarts.lx.model.LXPoint;
 import titanicsend.pattern.TEAudioPattern;
```
jkbelcher, line 5: import TEAudioPattern no longer needed

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePattern.java
zranger1: Will be fixed in next commit

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePort.java
zranger1: Fixed in next commit

src/main/java/titanicsend/pattern/pixelblaze/Wrapper.java
zranger1: Fixed in next commit

resources/pixelblaze/glue.js
zranger1: Indeed it can!  Fixed in next commit.

src/main/java/titanicsend/pattern/pixelblaze/Wrapper.java
zranger1: Fixed in the latest commit. 

Notes: Yes, some precision is lost, but with a 64-bit counter, given the speed of our animations, and the relatively long timescales we're dealing with, there won't be a visible difference (Unless you run a pattern for several weeks on end, maybe.)  

Having said that, I don't like the redundant divide/multiply either.  So I've added::

- ```getTimeMs()```, which gets the full precision millisecond count from VariableSpeedTimer for things that prefer milliseconds to seconds.millis.
- ```getDeltaMs()```, which returns the number of milliseconds since the last call to ```VariableSpeedTimer.Tick()```

src/main/java/titanicsend/pattern/pixelblaze/PixelblazeParallel.java
zranger1: Fixed in next commit. Note that, sadly, this is semi-deprecated.

Ben and I did some experiments last year and found that Java's thread synchronization mechanics have so much overhead it wasn't a net performance gain.  Then, I played with it later and learned that simply spin-locking in the worker threads actually made it work pretty well.  In this case though, I didn't think it'd be a good idea to tie up resources that the rest of LX might want, so there it sits...

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
zranger1: Fixed in next commit.

zranger1: Fixed in next commit.

zranger1: Fixed in next commit.  See comments for same issue in Phasers.

src/main/java/titanicsend/pattern/jon/Phasers.java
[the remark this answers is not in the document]
zranger1: Fixed in next commit.  This is the classic - "am I inside the scene looking out, or outside looking in", mistake.

src/main/java/titanicsend/pattern/jon/Phasers.java
```diff
@@ -34,27 +27,28 @@ public Phasers(LX lx) {
...
+        // start with beam split 5 ways and spinning slowly
+        controls.setRange(TEControlTag.QUANTITY, 5, 1, 8)
+                .setUnits(TEControlTag.QUANTITY, LXParameter.Units.INTEGER);
+
+        // Speed controls background movement speed and direction
+        controls.setRange(TEControlTag.SPEED, 0.25, -1.0, 1.0);
 
-        controls.setRange(TEControlTag.SIZE,1,0.5,5);
+        // Spin controls spin rate
+        controls.setRange(TEControlTag.SPIN, 0.25, -1, 1);
```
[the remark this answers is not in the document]
zranger1: Fixed in next commit.  Defaults to 0, but has some initial spin via the constructor.   (The initial spin is there just to make it really, really obvious to newcomers what this pattern does.)

resources/shaders/phasers.fs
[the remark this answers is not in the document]
zranger1: Made a note of this, and will fix asap.  (I'll need to change it so WowTrigger runs a small state machine based on where we are in the beat instead of just conditionally doing something when pressed.)

src/main/java/titanicsend/pattern/ben/Audio1.java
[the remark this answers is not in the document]
zranger1: Fixed in next commit.   The sneaky Pixelblaze gradient patterns were getting their colors in a way that bypassed brightness.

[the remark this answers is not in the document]
zranger1: Fixed in next commit.   
- Wow1 now controls beat reactive pulse
- Wow2 now controls noise resolution

src/main/java/titanicsend/pattern/ben/Xorcery.java
zranger1: Fixed in latest commit.   It's fine for this value to be 0 - the 0.05 was just preserved from the Pixelblaze original.

zranger1:
Also added in this commit: TESparklePattern.java.   It's a first cut at wrapping an LX standard pattern for our controls.  The controls work as follows:

- Speed: Overall sparkle speed
- Quantity: Sparkle density
- Wow1: Base brightness level
- Wow2: Sharpness of sparkle curve

src/main/java/titanicsend/pattern/jon/Phasers.java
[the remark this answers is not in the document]
jkbelcher: There is no spoon.

.lxpreferences
```diff
@@ -1,6 +1,6 @@
 {
   "version": "0.4.2-SNAPSHOT",
-  "projectFileName": "AutoVJ.lxp",
+  "projectFileName": "will.lxp",
```
jkbelcher, line 3: This looks accidental.

zranger1: It is.  I'll kill it.

src/main/java/titanicsend/pattern/jon/Phasers.java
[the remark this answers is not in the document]
jkbelcher: Spin goes the correct direction now but Angle is still backwards.  And spin should be balanced so Off is in the center of the bipolar knob.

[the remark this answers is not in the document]
zranger1: Ok... this was just boneheaded of me.  Will be fixed shortly.

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
jkbelcher: With Wow1 at zero there's still a visual pulsing to the beat.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -388,34 +390,44 @@ public double getStaticRotationAngle() {
...
      * a unified mechanism for handling brightness.
-     *
      */
     public int getCurrentColorControlValue() {
         return controls.color.calcColor();
     }
 
     /**
-     *
      * @return current variable time in seconds.millis.  Note that time
```
jkbelcher, line 416: This comment looks like it should be moved to getTimeMs() below.

src/main/java/titanicsend/pattern/pixelblaze/Wrapper.java
jkbelcher: Great.  Added a comment about moving the description to the new method.

src/main/java/titanicsend/pattern/jon/TESparklePattern.java
jkbelcher, line 1: Ok.  We can add this here but we can also move it to a different PR if you want to get the rest of this through.  I'm offline today but at a quick glance:

- Speed of zero (centered) should be no movement.  And resetting speed should go to zero (ie default=0).  TBH what you've done makes sense which is setting the range 0-2.  But we're limited for now by the inability to change the biPolar setting on a control from one pattern to the next (causes a MidiFighterTwister reboot lag).  I would do the call as `controls.setRange(TEControlTag.SPEED, 0, -2, 2);`  And when you pull the value clamp it to (0,2) ie turning the speed knob to the left of center will have no action.  OR use the absolute value of speed instead of clamping it to zero.
- Wave/Min/Max parameters need a home. -> Size/X/Y?

src/main/java/titanicsend/pattern/jon/Phasers.java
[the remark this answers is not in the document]
jkbelcher: Great.  On a similar note, we should make a habit of limiting the maximum Speed and Spin values to things that are visible on the car at our framerate (60/70fps, say 60 to be safe).  It's ideal to sling a knob to maximum to get maximum effect.  But the Spin on Phasers right now when turned to maximum actually looks like it slows down and moves backwards.  It's going too far around for the motion to be perceptible.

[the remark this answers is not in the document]
jkbelcher: (That's probably worth a group announcement to everyone doing pattern refactors.  I'll post to slack.)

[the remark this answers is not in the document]
zranger1: Yes, I can do that, and it makes sense.   The idea was to make it fast enough to be able to blur and rely on the POV effect to simulate a rapidly scanning laser projector.  But without being able to test on an actual giant LED display, it's hard to know if that works.   I'll slow it down to something reasonable.

src/main/java/titanicsend/pattern/jon/TESparklePattern.java
zranger1: All this can be done easily.  Would've use abs() to start with, but I didn't know you had to reset the controller.

src/main/java/titanicsend/pattern/jon/Phasers.java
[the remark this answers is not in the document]
jkbelcher: I agree blurry fast would be good.  I think the trouble is a certain spin rate makes the pattern line up in the exact same place every frame, so it suddenly comes out of blurry into looking static.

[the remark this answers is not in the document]
zranger1: Ha!  The well known 24fps movie wagon wheel effect.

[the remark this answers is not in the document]
jkbelcher: We think our technology is so great and we still have wagon wheel problems.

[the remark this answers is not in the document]
zranger1: Fixed in latest commit.   Max rotation speed is now limited for all patterns that use ```getRotationAngleFromSpin()``` or ```getRotationAngleForSpeed()```   

There's also a new API - ```setMaxRotationSpeed(double radiansPerSecond)``` that gives pattern authors the ability to change the default maximum if they're doing something especially "interesting".

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
zranger1: Fixed -- this pattern is generally better behaved now.

src/main/java/titanicsend/pattern/pixelblaze/PixelblazePattern.java
zranger1: This is actually still necessary - though we don't use it this way, imported Pixelblaze patterns can set up their own controls.  They're built automatically from the PB javascript and added after the common controls.   I'm for preserving this option for easy porting and prototyping.

src/main/java/titanicsend/pattern/pixelblaze/Wrapper.java
zranger1: Fixed. Revised and improved the comments on all the new time-related functions.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Fixed, along with other comment improvements

src/main/java/titanicsend/pattern/jon/TESparklePattern.java
zranger1: Done!  Wave/Min/Max on Size/X/Y as requested.

src/main/java/titanicsend/pattern/jon/Phasers.java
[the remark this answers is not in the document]
jkbelcher: Awesome.  Love the management of interaction with BPM.

src/main/java/titanicsend/pattern/jon/TESparklePattern.java
jkbelcher: I don't want parameter-swapping to get too crazy, but I think this one is fine because wave/min/max can put on a hat and glasses and get in to the party as size/x/y.  The overall type of thing changed by the knob is similar.

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
jkbelcher: Agreed, behavior on this one is looking dialed.  Coffee must have worked!

jkbelcher (approved):
Looks good!

PR #196, opened 2023-04-19
Updated bouncing dots

worldveil:
Updated to use `TEPeformancePattern` as a first try using new inheritance system.


zranger1 (approved):
Code is fine, runs perfectly w/new controls!

PR #197, opened 2023-04-20
Rename color methods

jkbelcher:
Refactored the main color methods to match current LX framework naming:
`calcColor()` -> instead of `getCurrentColor()`.  This is the main one, gets "final" color, ie post-brightness.
`getColor()` -> instead of `getCurrentColorControlValue()`.  This is "pre-modification" color, ie pre-brightness.

Added `@deprecated` warnings to old gradient methods to discourage accidental use.

### DIFF
--- a/src/main/java/titanicsend/pattern/TEPattern.java
+++ b/src/main/java/titanicsend/pattern/TEPattern.java
@@ -91,7 +91,7 @@ public abstract class TEPattern extends LXModelPattern<TEWholeModel> {
   /*
    * Color methods
    */
-
+  @Deprecated
   public LinkedColorParameter registerColor(String label, String path, ColorType colorType, String description) {
     LinkedColorParameter lcp = new LinkedColorParameter(label)
             .setDescription(description);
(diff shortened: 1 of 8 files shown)


jkbelcher:
For everyone's sanity, uses the naming pattern `getColor()` & `calcColor()` established in LX's `ColorParameter` and `LinkedColorParameter`.

zranger1 (approved):
The code looks clean and straightforward, and everything worked during testing (except for a couple of older patterns we'll still need to bring up-to-date.)

I think we're going to need to re-revisit color and palette design at the next meeting though.  The new color control gives you tons of ways to play with it, but with the current mostly black palette setup, there are a lot of ways to get a mostly black display.

jkbelcher:
> I think we're going to need to re-revisit color and palette design at the next meeting though. The new color control gives you tons of ways to play with it, but with the current mostly black palette setup, there are a lot of ways to get a mostly black display.

Sounds good, happy to chat about it.  Some of the combinations enabled by the new control are not expected to be used often, so that could be distorting the average.  For example I expect Solid Color Source to *mostly* be set to Foreground.

PR #199, opened 2023-04-22
Tools for Updating FragmentShaderEffect Patterns

zranger1:
A few changes needed to simplify converting these patterns to use the common controls, plus a sample conversion of the NeonSnakeShader pattern.

This is a straightforward, very vanilla conversion, showing how to set up and use the common controls in this pattern framework.  We can add more features to this pattern later if necessary.  Controls are:

- Color: Yes
- Speed: Yes
- X/Y Offset: Y only, since the pattern moves constantly in X
- Spin/Angle:  No.
- Brightness: Yes
- Size:  Snake size
- Quantity:  No
- Wow1:  Base Glow
- Wow2:  Beat reactive snake size modulation
- Wow Trigger: No

### DIFF
--- a/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -170,7 +170,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
 
         /**
          * ** Solid-Color patterns should use this method **
-         * <p>
+         *
          * Returns the real-time value of the color, which may be different from what
          * getColor() returns if there are LFOs/etc being applied.
          * Offset has been applied to this color.
@@ -236,7 +236,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
 
         /**
          * ** Gradient patterns should use this method **
-         * <p>
+         *
          * Given a value in 0..1 (and wrapped back outside that range)
          * Return a color within the selected gradient.
          * Offset is added to lerp to create a user-shiftable gradient.
@@ -486,7 +486,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
          * To use the common controls, call this function from the constructor
          * of TEPerformancePattern-derived classes after configuring the default
          * controls for your pattern.
-         * <p>
+         *
          * If your pattern adds its own controls in addition to the common
          * controls, you must call addParameter() for them after calling
          * this function so the UI stays consistent across patterns.
@@ -672,7 +672,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
 
     /**
      * Class to support incremental rotation over variable-speed time
-     * <p>
+     *
      * The rate is tied to the engine bpm and the input time value, which is usually
      * controlled by the variable speed timer associated with the speed or spin controls.
      * (but anything with a seconds.millis timer can generate rotational angles this way.)
@@ -809,7 +809,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     /**
      * Sets the maximum rotation speed used by both getRotationAngleFromSpin() and
      * getRotationAngleFromSpeed().
-     * <p></p>
+     *
      * The default maximum radians per second is PI, which gives one complete rotation
      * every two beats at the current engine BPM.  Do not change this value unless you
      * have a specific reason for doing so.  Too high a rotation speed can cause visuals
@@ -824,14 +824,14 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
 
     /**
      * @return Color derived from the current setting of the color and brightness controls
-     * <p></p>
+     *
      * NOTE:  The design philosophy here is that palette colors (and the color control)
      * have precedence.
-     * <p></p>
+     *
      * Brightness modifies the current color, and is set to 1.0 (100%) by default. So
      * if you don't move the brightness control you get *exactly* the currently
      * selected color.
-     * <p></p>
+     *
      * At present, the brightness control lets you dim the current color,
      * but if you want to brighten it, you have to do that with the channel fader or
      * the color control.
@@ -842,11 +842,11 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
 
     /**
      * ** Instead of these two methods, use getGradientColor(lerp) which defers to TEColorParameter for gradient selection. **
-     * <p>
+     *
      * Suppress parent TEPattern gradient methods, force child classes
      * to choose solid color or gradient, keeping other choices
      * runtime-adjustable.
-     * <p>
+     *
      * TODO: remove these two methods from TEPattern to prevent confusion?
      */
     @Deprecated
@@ -949,7 +949,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     /**
      * <b>NOTE:</b> This control has functional overlap with color and channel fader settings, and
      * could potentially cause confusing brightness behavior.
-     * <p></p>
+     *
      * <b>It may be deprecated or removed in the future and should not be used in patterns.</b>
      *
      * @return The current value of the brightness control, by default in the range 0.0 to 1.0
@@ -973,7 +973,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     /**
      * Restarts the specified timer's elapsed time when called.
      * The timer's rate is not changed.
-     * <p>
+     *
      * This is useful for syncing a timer precisely to beats,
      * measures and other external events.
      *
--- a/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -599,7 +599,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     private final Rotor speedRotor = new Rotor();
     private final Rotor spinRotor = new Rotor();
 
-    protected TECommonControls controls;
+    public TECommonControls controls;
 
     protected FloatBuffer palette = Buffers.newDirectFloatBuffer(15);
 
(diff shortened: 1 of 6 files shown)


src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -599,7 +599,7 @@ void setMaxSpinRate(double radiansPerSecond) {
     private final Rotor speedRotor = new Rotor();
     private final Rotor spinRotor = new Rotor();
 
-    protected TECommonControls controls;
+    public TECommonControls controls;
```
jkbelcher, line 602: Let's do the LX style here of keeping this protected, (or better private but that would take a wide refactor), and use an accessor method for public access.  It keeps someone from replacing the variable which would leave us in a bad state if we had listeners attached.
```
private/protected TECommonControls controls;
public TECommonControls getControls() {
  return this.controls;
}
```

src/main/java/titanicsend/pattern/yoffa/effect/shaders/NeonSnakeShader.java
```diff
@@ -31,37 +24,50 @@ public class NeonSnakeShader extends FragmentShaderEffect {
...
         double[] uv = divideArrays(fragCoordinates, resolution);
-        double[] waveColor = new double[]{0, 0, 0};
+        uv[1] -= pattern.getYPos() + 0.25;  // offset y to roughly center snake vertically
+        uv = multiplyArray(3, uv); // scale pattern properly for car
 
-        uv = addToArray(-1.3, multiplyArray(2.4 * ySquish.getValue(), uv));
-        uv[1] += .1;
+        // get current calculated palette color (plus alpha, which we'll fill in later)
+        double[] waveColor = new double[4];
+        colorToRGBArray(pattern.calcColor(), waveColor);
```
jkbelcher, line 53: Ideally on a pattern rendering a solid color we'd only call pattern.calcColor() once per frame, not once per pixel.  Am I reading correctly that this will get called once per pixel?

Maybe we should modify the calcColor() method to do the math only once per frame, then we don't have to worry about it out here in the patterns.  Can save that for a different PR.

src/main/java/titanicsend/pattern/yoffa/effect/shaders/NeonSnakeShader.java
jkbelcher, line 1: There are some unused imports on this file.

src/main/java/titanicsend/pattern/yoffa/effect/shaders/FragmentShaderEffect.java
```diff
@@ -10,10 +11,12 @@
 import java.util.*;
 
 import static heronarts.lx.utils.LXUtils.clamp;
+import static java.lang.Math.cos;
+import static java.lang.Math.sin;
+import static titanicsend.util.TEMath.multiplyVectorByMatrix;
```
jkbelcher, line 16: These 3 imports are unused.

src/main/java/titanicsend/pattern/yoffa/framework/PatternEffect.java
jkbelcher, line 1: Two unused imports here.

jkbelcher (changes requested):
The controls look good on NeonSnakeShader.  Let's switch to the `getControls()` method at least for non-child classes.  Otherwise just minor cleanup comments.

src/main/java/titanicsend/pattern/yoffa/effect/shaders/NeonSnakeShader.java
zranger1: It does, and that is bad.

This is a deprecated legacy framework though, and I'm trying to invest the smallest possible amount of time to get the handful of patterns that use it running.   If these patterns show a big frame rate drop when tested, rather than fix the framework, I think we should encourage people to port to normal TEPerformancePattern patterns (or OpenGL).

And yes, in the case of calcColor()  and probably others, it'd be good to calculate only once per frame.  But only if it's relatively uncomplicated to do so since we have so little time for testing and bug fixes before EDC.

The Historical Record:
It's from before my time, and from very early in TE's development -- a fragment shader emulation framework in Java.  The idea was that you translate a shader into a ```getColorForPoint()``` function and you're basically done.  All the per frame (rather than per pixel) setup is done somewhere above this in the hierarchy.

So to add per-frame color and rotation and all the other modern conveniences, we'd have to dig in and touch some code that I'd prefer to leave alone until we can safely remove it.

src/main/java/titanicsend/pattern/yoffa/effect/shaders/NeonSnakeShader.java
zranger1: Will fix in next commit

src/main/java/titanicsend/pattern/yoffa/effect/shaders/FragmentShaderEffect.java
zranger1: Will fix in next commit.

src/main/java/titanicsend/pattern/yoffa/framework/PatternEffect.java
zranger1: Will fix in next commit.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Good idea!   Will implement in the next commit.  (Although if anybody replaces this variable, they kinda... deserve what they get.   With great power comes great responsibility!)

src/main/java/titanicsend/pattern/yoffa/effect/shaders/NeonSnakeShader.java
zranger1: Ok.   I fixed this for rotation and color - this style of pattern is CPU intensive enough that it's worth a little work to keep overall performance high.   Added functions:

- `calcColor()`
-` calcColor2()`
- `rotate2D(point,origin) `

to FragmentShaderEffect class, and added spin support to NeonSnakeShader.  (Angle might actually be useful with this pattern.  Not so sure about actually spinning that snake around.   But it works as an example.)

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -39,20 +39,20 @@ private SolidColorSource(String label) {
...
                 return this.label;
             }
-        };
+        }
 
         public final EnumParameter<SolidColorSource> solidSource =
-            new EnumParameter<SolidColorSource>("SolidSource", SolidColorSource.FOREGROUND)
-            .setDescription("For a solid color: Whether to use global TE palette (preferred), or a static color unique to this pattern");
+                new EnumParameter<SolidColorSource>("SolidSource", SolidColorSource.FOREGROUND)
+                        .setDescription("For a solid color: Whether to use global TE palette (preferred), or a static color unique to this pattern");
```
jkbelcher, line 46: Hey @zranger1 the refactor on this file now has lines using 6 and even 8 spaces for indentation.  4 was standard on the file.

zranger1: I just ran the usual automated tool on it.  Wonder what happened.

zranger1: Oh... it looks like ideaJ uses a different convention for continued lines.  At least it's consistent throughout.   I'll see if I can convince it to be a little more conservative about the number of spaces.

zranger1: Ok... should take up a little less space now.   I wonder who decides what the defaults are for this stuff...

jkbelcher: Formatting looks mostly fixed now.  Have a couple stray "<p>" lines, will flag.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -194,22 +195,23 @@ public int calcColor2() {
...
+                    return super.getColor();
             }
         }
 
         // GRADIENT METHODS
 
         /**
          * ** Gradient patterns should use this method **
-         *
+         * <p>
```
jkbelcher, line 210: Formatter blip.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -697,11 +706,11 @@ public int calcColor() {
 
     /**
      * ** Instead of these two methods, use getGradientColor(lerp) which defers to TEColorParameter for gradient selection. **
-     *
+     * <p>
```
jkbelcher, line 709: Formatter blip.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -697,11 +706,11 @@ public int calcColor() {
...
      * Suppress parent TEPattern gradient methods, force child classes
      * to choose solid color or gradient, keeping other choices
      * runtime-adjustable.
-     * 
+     * <p>
```
jkbelcher, line 713: Formatter blip.

src/main/java/titanicsend/pattern/yoffa/effect/shaders/NeonSnakeShader.java
jkbelcher: Looks good.  I can do a quick mod to TEPerformancePattern to make this safer too.

src/main/java/titanicsend/pattern/yoffa/framework/PatternEffect.java
jkbelcher: Still have 2 here.

jkbelcher (approved):
I just added the last formatting changes.  I'm away from the desk today so no need to keep holding this up over minor formatting.  Carry on!

PR #200, opened 2023-04-23
Calc solid colors only once per frame

jkbelcher:
As discussed with @zranger1 in another thread, this would protect against unnecessary repeat color calculation if a pattern calls the color methods on every pixel.  *Even with this safety mechanism we should encourage patterns to do it right when we catch them.

Could use another set of eyes on this.


zranger1 (approved):
Simple, clean, and will help keep pattern authors out of trouble.

PR #201, opened 2023-04-25
Take the gray out of gradients

jkbelcher:
Thanks to upstream assistance from mcslee, future LX releases will have improved color blending especially when blending to/from black.

This PR brings in the new blend functions so we can use them immediately.  This is a temporary patch until we upgrade to a version of LX that includes the new features.

image

### DIFF
--- a/src/main/java/titanicsend/lx/LXGradientUtils.java
+++ b/src/main/java/titanicsend/lx/LXGradientUtils.java
@@ -1,0 +1,318 @@
+/**
+ * Copyright 2016- Mark C. Slee, Heron Arts LLC
+ *
+ * This file is part of the LX Studio software library. By using
+ * LX, you agree to the terms of the LX Studio Software License
+ * and Distribution Agreement, available at: http://lx.studio/license
+ *
+ * Please note that the LX license is not open-source. The license
+ * allows for free, non-commercial use.
+ *
+ * HERON ARTS MAKES NO WARRANTY, EXPRESS, IMPLIED, STATUTORY, OR
+ * OTHERWISE, AND SPECIFICALLY DISCLAIMS ANY WARRANTY OF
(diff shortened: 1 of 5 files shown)


src/main/java/titanicsend/lx/LXGradientUtils.java
```diff
@@ -0,0 +1,318 @@
...
+      this.hue = color.getHuef() + hueOffset;
+      this.saturation = color.getSaturation();
+      this.brightness = LXColor.b(c);
+      setRGB(LXColor.hsb(this.hue, this.saturation, this.brightness));
+    }
+
+    public void set(LXDynamicColor color, float hueOffset, float saturationOffset, float brightnessOffset) {
+      int c = color.getColor();
+      this.hue = color.getHuef() + hueOffset;
+      this.saturation = LXUtils.clampf(LXColor.s(c) + saturationOffset, 0, 100);
```
zranger1, line 125: On line 124, does this
`this.hue = color.getHuef() + hueOffset;`
need to be range checked, or wrapped at 0 and 360, or is it done somewhere downstream?

zranger1 (changes requested):
Big visual improvement!

src/main/java/titanicsend/lx/LXGradientUtils.java
jkbelcher: Yeah it looks like it wants to be wrapped but this is a direct copy of LX's GradientUtils class and the two internal classes using it seem to be wrapping hue just fine.  It must be getting sorted out in the LXColor.hsb(...) call.
image

zranger1: Good enough!   I thought it'd be caught elsewhere in the framework, but just wanted to be sure.

PR #202, opened 2023-04-25
Unify Color Sources for Patterns

zranger1:
All patterns, including GLSL shaders now get their colors from the same place -- the color control via calcColor() and calcColor2() methods.

Shaders that formerly used iPalette to create their own gradients have been update to use iColorRGB and iColor2RGB instead.

This is a necessary prerequisite to updating patterns (especially shaders) that use gradients.


jkbelcher (approved):
Looking at the code this appears to be a straightforward conversion of color methods.

PR #204, opened 2023-04-26
Clean up 2022 config to provide a clean slate for 2023 build day

raldi:


worldveil (approved):

jkbelcher:
So the .tsv files were deleted in #204 and main is now broken.  Can @raldi or @worldveil properly restore these file to fix main?

PR #205, opened 2023-04-27
Add vertex-renumbering script

raldi:
...though we're decided not to run it against `main` until after EDC.


jkbelcher (approved):
Script looks good, output looks correct on the first 3 files here.

The `.tsv` files are missing so the script crashes on those.  They were deleted in #204.  Approving this script but please fix main.

PR #206, opened 2023-04-28
Remove TSVs, for real this time

raldi:


jkbelcher (approved):
Didn't have time to read this closely but it fixes the crash on main so approving PR.

PR #208, opened 2023-04-30
Update model/pattern relationship to be compatible with Views

jkbelcher:
This is conceptually a simple change.  It swaps the `TEPattern` and `TEEffect` base classes away from legacy LX classes and onto a new style which is compatible with Views.  There were a lot of classes touched by the refactor but it's nothing too crazy.

Devs should be aware that within Pattern/Effect classes, there are now TWO ways to access the model.  To reach the TE-specific stuff, use `this.modelTE` or `getModelTE()`.  For a list of points and their normalized coordinates (ie, for iterating over points in the run() loop), use `this.model` or `getModel()`.  This is necessary because when a View is applied to a channel, this.model.points will be only the points that are visible within the view.

If you want to follow the logic it may be easier to look at the 4 commits from this branch one at a time.


worldveil:
Pulled and ran, seems to work fine. LGTM

worldveil (approved):

jvyduna:
> a new style which is compatible with Views

This is REALLY exciting! I remember finding this problem hard, emailing Mark, and he also agreed it would be difficult. Thanks so much - this allows us to much more flexibly target patterns!

PR #210, opened 2023-05-02
Update WaterPanels and WaterEdges Patterns

zranger1:
They both use the same shader.  Controls are:
- Speed: Yes
- X/Y Offset: Yes
- Spin/Angle: Yes
- Brightness: Yes
- Size: Overall scale
- Quantity: Tile repeat factor
- Wow1: Iterations (old intensity 1)
- Wow2: Intensity 2
- WowTrigger: No

(To get the high contrast wavy snake line effect, set Wow1 as low as possible, and Wow2 to nearly 0.  Now try spinning, or scaling for... more snakes!)

### DIFF
--- a/src/main/java/titanicsend/pattern/yoffa/effect/shaders/WaterShader.java
+++ b/src/main/java/titanicsend/pattern/yoffa/effect/shaders/WaterShader.java
@@ -13,22 +13,8 @@ import static titanicsend.util.TEMath.*;
 //https://www.shadertoy.com/view/MdlXz8
 public class WaterShader extends FragmentShaderEffect {
 
-    public final CompoundParameter tile =
-            new CompoundParameter("Tiling", 0, 1, 4)
-                    .setDescription("");
-
-    public final CompoundParameter speed =
-            new CompoundParameter("Speed", 0.1, 0.1, 10)
-                    .setDescription("");
-
-    public final CompoundParameter intensity =
-            new CompoundParameter("Inten1", 5, 1, 20)
-                    .setDescription("Intensity");
-
-    public final CompoundParameter intensity2 =
-            new CompoundParameter("Inten2", .005, .001, .010)
-                    .setDescription("Intensity but like different though");
     private static final double TAU = 6.28318530718;
+    private static final double[] origin = new double[]{0.5, 0.25};
 
     public WaterShader(PatternTarget target) {
         super(target);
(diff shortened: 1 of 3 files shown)


src/main/java/titanicsend/pattern/yoffa/effect/shaders/WaterShader.java
jkbelcher, line 1: Quantity and size seem to be doing the same thing (?)

jkbelcher, line 1: Size & Quantity zoom from the bottom right corner.  What do you think about zooming to center/center?  It would keep the change more balanced.

jkbelcher, line 1: Default Speed to zero, init to value.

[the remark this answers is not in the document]
zranger1: They do!  (... checks ...)  well, my version, which I evidently forgot to commit, does.   Fixed in latest commit.

zranger1: They don't but, it kinda looks like it.  I took your suggestion on keeping everything centered during zoom and tile changes, and that should make the difference more apparent.

- Quantity sets the number of times the wave tile repeats across the canvas.
- Size controls the overall size of the canvas.

So when you change size, waves at the current level of detail get larger or smaller.  The tiling doesn't change.   When you change quantity, the apparent complexity of the wave pattern changes.  Set quantity high enough and it becomes pretty easy to see how the tiles are working.

It's a little weird and redundant, I know.  But this is another legacy pattern that I'm pretty sure was semi-automatically translated from Shadertoy during a set of early experiments.   If we want it to do more later, I can translate it back to GLSL and expand from there.

zranger1: Good idea - done in latest commit.

zranger1: Fixed in latest commit.

jkbelcher: Center zoom looks great

jkbelcher (approved):

PR #211, opened 2023-05-02
Rework AutoVJ enabled scaffolding

jkbelcher:
Cleans up some AutoVJ on/off behavior:

- Uses BooleanParameter internally for enabled flag
- Can't get out of sync with UI enabled button (previously this was a risk)
- Only attempts startup once, when enabled is changed to true
- If startup fails, does not re-attempt startup on every loop (as previously doing from loop())
- Captures startup failure without crash and prints stack trace to System error stream (red text in eclipse).
- *Should* cleanly restore AutoVJ enabled state on file save/open.
- Cleans up some UI wiring & listeners in TEApp.
- Increases UI button size.

### DIFF
--- a/src/main/java/titanicsend/app/TEApp.java
+++ b/src/main/java/titanicsend/app/TEApp.java
@@ -28,11 +28,9 @@ import java.util.Calendar;
 import java.util.function.Function;
 
 import heronarts.lx.LX;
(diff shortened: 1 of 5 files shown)


jkbelcher:
...now it runs!

PR #212, opened 2023-05-02
Turn on glPointCloud for UI speed gain

jkbelcher:
Thanks mcslee for pointing out this setting.  Makes the UI way faster, especially in performance mode where it is rendering two preview windows.


jkbelcher:
Recommend increasing the point size in Left Pane > Model > Camera:
image

zranger1 (approved):
I don't see any problem with this.  Changing the point size helps, and it is easy to switch back to the other renderer for comparison if needed.

PR #214, opened 2023-05-03
AutoVJ auto-loader and templating logic

worldveil:
AutoVJ now, upon enabled=true, loads channels from JSON .lxp template into the current live project. 

AutoVJ is now a portable LXGroup that can be faded in and out with a single fader. 

We align so that AutoVJ group is always the 8th channel so it is the last channel on the APC40, handy for VJs to fade in and out between automated and manual. 

Also includes a number of code quality fixes and improvements around AutoVJ.


jkbelcher (approved):
I took a look at the JSON id stripping, I actually think this is fine.  Because the LX instance already exists and is passed to every `load(...)` method, each object gets assigned an id when the method `LXComponent.setParent(...)` is called, and subsequently calls `this.lx.componentRegistry.register(this)`.

On the channels this happens when they are added to the group in TEAutoPilotMixer `group.addChannel(c)`, and for patterns it happens in from `LXChannel.load()` -> `loadPattern(...)` -> `addPattern(...)` -> `pattern.setChannel(this)`.  It's ok the patterns get their IDs before the channel too, as the ids are all coming from a central location.

Stoked on how much more friendly this will make AutoVJ to the live user.

PR #216, opened 2023-05-03
Fix pattern references in .lxps for moved patterns

jkbelcher:
Fixes the broken pattern references in .lxp project files.  A couple patterns were moved around when they were updated to the new base class.

Includes the one that was failing on AutoVJ start (loading from AUTOVJ_TEMPLATE.lxp).


zranger1 (approved):
All good!   (and someday, we might consider switching this loader to a tag-based system, or something else that isn't as tightly tied to the file structure of the source code.)

PR #217, opened 2023-05-03
Fix compiler resource leak warnings

jkbelcher:
Fixes these warnings:
image


zranger1 (approved):
Interesting and good catch!

I'd always thought that Java closed file/stream objects when they went out of scope (or at least when the GC finally caught up with them).   Apparently, it might be quite a while before they get closed, so it is better to do it yourself!.

PR #219, opened 2023-05-05
New noise patterns, Round 1

zranger1:
These two patterns show the general direction I'm taking with noise.  Rather than writing one gigantic noise pattern to rule them all, I'm doing an assortment of small, high performance patterns, with features that work with the specific type and appearance of the noise.

When testing, please be sure to turn off any LX modulators that are active on the channel. No problem using them in production if necessary, but they do tend to ~~stomp~~ obscure pattern detail.

### Turbulence Lines
Generates a value noise based turbulence field, then uses the field to draw sinuous, organic-looking lines.  Wow1 controls the noise field/line mix, Wow2 controls the foreground color/gradient mix.

Wow trigger fires a double pulse on the beat that lights the lines (not the background fog), so you can turn the lines way down in the mix, and they will be revealed as the pulse passes.

- Speed: Yes
- X/Y Offset: Yes
- Spin/Angle: Yes
- Brightness: Yes
- Size: Overall scale
- Quantity: Line density
- Wow1: Noise field/wavy line mix
- Wow2: Foreground color/gradient mix
- WowTrigger: Radial double pulse on the beat

### Simplex Posterized
Generates a cloudy simplex noise field which you can color with the current gradient, and "Posterize" - quantize to a small number of brightness levels - for a cartoon-like look.

Since static x and y offset are not useful for this sort of pattern, they are instead used to control "drift" - x and y movement speed.

Wow1 controls the posterization level, Wow2 controls the foreground color/gradient mix.

- Speed: Yes
- X/Y Offset: Yes, controls "drift" speed and directrion
- Spin/Angle: Yes
- Brightness: Yes
- Size: Overall scale
- Quantity: Noise field density
- Wow1: Posterization level
- Wow2: Foreground color/gradient mix
- WowTrigger: No

### DIFF
--- a/resources/shaders/simplex_posterized.fs
+++ b/resources/shaders/simplex_posterized.fs
@@ -1,0 +1,161 @@
+// Posterized simplex noise shader for Titanic's ENd
+//
+// Uses the Ashima Arts 2D/3D/4D simplex noise generator
+//
+// Copyright (C) 2011 Ashima Arts. All rights reserved.
+// Distributed under the MIT License. See LICENSE file.
+// https://github.com/ashima/webgl-noise
+//
+
+vec3 mod289(vec3 x) {
+    return x - floor(x * (1.0 / 289.0)) * 289.0;
+}
+
+vec4 mod289(vec4 x) {
+    return x - floor(x * (1.0 / 289.0)) * 289.0;
+}
+
+vec4 permute(vec4 x) {
+    return mod289(((x * 34.0) + 1.0) * x);
+}
+
+vec4 taylorInvSqrt(vec4 r)
+{
+    return 1.79284291400159 - 0.85373472095314 * r;
+}
+
+float snoise(vec3 v)
+{
+    const vec2 C = vec2(1.0 / 6.0, 1.0 / 3.0);
+    const vec4 D = vec4(0.0, 0.5, 1.0, 2.0);
+
+    // First corner
+    vec3 i = floor(v + dot(v, C.yyy));
+    vec3 x0 = v - i + dot(i, C.xxx);
+
+    // Other corners
+    vec3 g = step(x0.yzx, x0.xyz);
+    vec3 l = 1.0 - g;
+    vec3 i1 = min(g.xyz, l.zxy);
+    vec3 i2 = max(g.xyz, l.zxy);
+
+    //   x0 = x0 - 0.0 + 0.0 * C.xxx;
+    //   x1 = x0 - i1  + 1.0 * C.xxx;
+    //   x2 = x0 - i2  + 2.0 * C.xxx;
+    //   x3 = x0 - 1.0 + 3.0 * C.xxx;
+    vec3 x1 = x0 - i1 + C.xxx;
+    vec3 x2 = x0 - i2 + C.yyy; // 2.0*C.x = 1/3 = C.y
+    vec3 x3 = x0 - D.yyy;      // -1.0+3.0*C.x = -0.5 = -D.y
+
+    // Permutations
+    i = mod289(i);
+    vec4 p = permute(permute(permute(
+                                 i.z + vec4(0.0, i1.z, i2.z, 1.0))
+                             + i.y + vec4(0.0, i1.y, i2.y, 1.0))
+                     + i.x + vec4(0.0, i1.x, i2.x, 1.0));
+
+    // Gradients: 7x7 points over a square, mapped onto an octahedron.
+    // The ring size 17*17 = 289 is close to a multiple of 49 (49*6 = 294)
+    float n_ = 0.142857142857; // 1.0/7.0
+    vec3 ns = n_ * D.wyz - D.xzx;
+
+    vec4 j = p - 49.0 * floor(p * ns.z * ns.z);  //  mod(p,7*7)
+
+    vec4 x_ = floor(j * ns.z);
+    vec4 y_ = floor(j - 7.0 * x_);    // mod(j,N)
+
+    vec4 x = x_ * ns.x + ns.yyyy;
+    vec4 y = y_ * ns.x + ns.yyyy;
+    vec4 h = 1.0 - abs(x) - abs(y);
+
+    vec4 b0 = vec4(x.xy, y.xy);
+    vec4 b1 = vec4(x.zw, y.zw);
+
+    //vec4 s0 = vec4(lessThan(b0,0.0))*2.0 - 1.0;
+    //vec4 s1 = vec4(lessThan(b1,0.0))*2.0 - 1.0;
+    vec4 s0 = floor(b0) * 2.0 + 1.0;
+    vec4 s1 = floor(b1) * 2.0 + 1.0;
+    vec4 sh = -step(h, vec4(0.0));
+
+    vec4 a0 = b0.xzyw + s0.xzyw * sh.xxyy;
+    vec4 a1 = b1.xzyw + s1.xzyw * sh.zzww;
+
+    vec3 p0 = vec3(a0.xy, h.x);
+    vec3 p1 = vec3(a0.zw, h.y);
+    vec3 p2 = vec3(a1.xy, h.z);
+    vec3 p3 = vec3(a1.zw, h.w);
+
+    //Normalise gradients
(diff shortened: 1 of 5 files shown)


resources/shaders/turbulent_noise_lines.fs
jkbelcher, line 1: This is great.  How about default Wow1 to .5?  I turned it all the way down and then took a minute to figure out how to get the lines back.

resources/shaders/simplex_posterized.fs
jkbelcher, line 1: Good one.  I really like the x/y drift effects.  The pixelation from turning Wow1 up is cool too.

resources/shaders/turbulent_noise_lines.fs
zranger1: Done in latest commit.

resources/shaders/simplex_posterized.fs
zranger1: Drift worked out well, and I need to use it again, so I generalized the control system so other patterns can use it without duplicating code.   Just moved the CanvasTranslator class to its own file, and added a preprocessor directive that GLSL can use to disable the default xoffset/yoffset behavior.

resources/shaders/turbulent_noise_lines.fs
zranger1: Using newly revised system, added drift support to TurbulenceLines, and gave the drift "wind" some influence over line positions.  Also slightly enhanced the wind's ability to blow bits off the edges of the lines.

resources/shaders/simplex_posterized.fs
jkbelcher: Excellent!

For this case, it seems like `CanvasTranslator` could be an abstract class between `TEPerformancePattern` and `SimplexPosturized`.  But I don't know if there are other places in shader land where the base class needs to be something other than `TEPerformancePattern`.

resources/shaders/turbulent_noise_lines.fs
jkbelcher: Woah this wind effect is cool!

jkbelcher: @zranger1 Wind might deserve to be generalized to a broader scope, like an effect that could be applied to any pattern.  I could see Size and Speed in addition to the x/y directions, so you could make it big but slow and drifty.  On playa in a wind storm this would be 🔥

zranger1: Agree -- the concept is certainly useful enough to be baked into TEPerformancePattern as an API so Java patterns could use it too.   The freakin' hard part is figuring out what to call it:   getDrift()?   getTranslationOverTime()?  

I'll build it, and hopefully come up with a decent name before I'm done.

resources/shaders/simplex_posterized.fs
zranger1: Will make this available to everybody through TEPerformancePattern.

jkbelcher: Doesn't Drift only work for noise style patterns that have no x/y boundary?  For some patterns this might not translate, so maybe it's a child class of TEPerformancePattern?

resources/shaders/turbulent_noise_lines.fs
jkbelcher: I think you have two things here:
-Drift: as a child class of TEPerformancePattern, does the x/y accumulation.  `getDrift()` sounds appropriate.
-Wind effect, which is this blowing pixel distortion.  This is what I was saying could be pulled out into an `LXEffect` that could be applied to any channel, rather than being built into patterns.

resources/shaders/simplex_posterized.fs
zranger1: Added DriftEnabledPattern class - really does make it much simpler to build one of these things, in GLSL or Java.

jkbelcher:
Check #226 for a mod to this.

src/main/java/titanicsend/pattern/jon/DriftEnabledPattern.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    void resetPosition() {
+        xOffset = 0;
+        yOffset = 0;
+    }
+
+    @Override
+    protected void run(double deltaMs) {
+        computeAudio(deltaMs);
```
jkbelcher, line 53: Don't need computeAudio() here, it is called in TEAudioPattern.run().

src/main/java/titanicsend/pattern/jon/TurbulenceLines.java
```diff
@@ -0,0 +1,57 @@
...
+        // register common controls with LX
+        addCommonControls();
+
+        effect = new NativeShaderPatternEffect("turbulent_noise_lines.fs",
+            PatternTarget.allPointsAsCanvas(this));
+
+    }
+
+    @Override
+    public void runTEAudioPattern(double deltaMs) {
```
jkbelcher, line 38: This would be `protected` to match the abstract definition unless it needs to be public?  LX.run() is also `protected`.

src/main/java/titanicsend/pattern/jon/TurbulenceLines.java
```diff
@@ -0,0 +1,57 @@
...
+
+        // run the shader
+        effect.run(deltaMs);
+    }
+
+    @Override
+    // THIS IS REQUIRED if you're not using ConstructedPattern!
+    // Initialize the NativeShaderPatternEffect and retrieve the native shader object
+    // from it when the pattern becomes active
+    public void onActive() {
```
jkbelcher, line 51: Also `protected`

src/main/java/titanicsend/pattern/jon/SimplexPosterized.java
```diff
@@ -0,0 +1,53 @@
...
+
+        // run the shader
+        effect.run(deltaMs);
+    }
+
+    @Override
+    // THIS IS REQUIRED if you're not using ConstructedPattern!
+    // Initialize the NativeShaderPatternEffect and retrieve the native shader object
+    // from it when the pattern becomes active
+    public void onActive() {
```
jkbelcher, line 47: `Protected`

src/main/java/titanicsend/pattern/jon/SimplexPosterized.java
```diff
@@ -0,0 +1,53 @@
...
+
+        // register common controls with LX
+        addCommonControls();
+
+        effect = new NativeShaderPatternEffect("simplex_posterized.fs",
+            PatternTarget.allPointsAsCanvas(this));
+    }
+
+    @Override
+    public void runTEAudioPattern(double deltaMs) {
```
jkbelcher, line 34: This would be `protected` to match the abstract definition unless it needs to be public? `LX.run()` is also protected.

jkbelcher (changes requested):
Almost there, just a couple nitpicks for posterity

jkbelcher (changes requested):
Almost there, just a couple nitpicks for posterity

src/main/java/titanicsend/pattern/jon/DriftEnabledPattern.java
zranger1: Fixed in latest commit.  Sheesh... and I theoretically know better, too.

src/main/java/titanicsend/pattern/jon/TurbulenceLines.java
zranger1: Fixed in next commit.

src/main/java/titanicsend/pattern/jon/TurbulenceLines.java
zranger1: Fixed in next commit.

src/main/java/titanicsend/pattern/jon/SimplexPosterized.java
zranger1: Fixed in next commit.

src/main/java/titanicsend/pattern/jon/SimplexPosterized.java
zranger1: Fixed in next commit.

jkbelcher (approved):
Approved!

PR #220, opened 2023-05-06
View selection from midi knob

jkbelcher:
Those of you who are new to LX Studio may still be figuring this out but the state of the LX framework is constantly evolving.  If you're coming back to it from having used it a few years ago you'll notice a bunch of improvements and new features.  One of the coolest features to evolve recently - and I would say still evolving - is Views.

Views add an abstraction layer between the pattern and the point coordinates.  By using normalized point coordinates (LXPoint.xn, yn, zn) in your pattern calculations, you can apply a view to get normalized coordinates relative to all fixtures visible within the view, or even to groups of fixtures within the view (by using semicolon delimiters).

This PR takes an additional step on what has been a long history of design and experimentation on the topic.  A current limitation of views - according to this speech - is the time it takes for a user to move between them.  This PR is a test of a new way to move between them quickly.

See the video below for a demo.  The process is:
-View definitions are loaded from a text file at system startup
-One LXView instance is created for each view definition
-The singleton ViewCentral keeps a list of the LXViews
-ViewCentral monitors the mixer channels and creates a "selected view" parameter for each channel.
-As the channel's parameter is changed, the corresponding view is pulled from the list and applied to the channel.
-The channel and pattern references are pointed to the new view, but otherwise no new objects are created.
-A virtual (aka pass-through) parameter was added to the patterns, to allow view selection straight from the MidiFighterTwister.  It looks like it's a pattern parameter, but really it's a channel parameter.
-Optional selection was added to the top row of knobs on the APC40Mk2, but control is easier on the MidiFighterTwister due to support of IncrementMode.

Initial resource observations:
RAM: this project was ~900MB before the changes, and now up to 1.5GB with 15 channels x 10 patterns each x 22 views.  Seems pretty minor as most Apple architecture machines have plenty of available RAM.
CPU: no detectable impact while running, no stutter of other channels as one channel's views are rotated.

https://www.youtube.com/watch?v=AO__WHWC3L0


jkbelcher:
Note the selected view doesn't yet survive file save/load, but that can be added later.  Wanted to get this in front of people asap.

zranger1 (approved):
Looks good to add so we can start improving view support.   Runs well here, and breaks nothing, even patterns that don't yet know about views.

(Remind me to ask you if there's anything sensible for onPanic() to do with the current view.   Can we let patterns have a default starting option, even if it's just edges, panels, or whole model?)

jkbelcher:
> (Remind me to ask you if there's anything sensible for onPanic() to do with the current view. Can we let patterns have a default starting option, even if it's just edges, panels, or whole model?)

Good idea.  I'll add a default per pattern!  I think we don't want every pattern to change it onActive() though, so the initial value is whatever was already set on the channel.  But we can have the Panic and View parameters both reset to the view preferred by the pattern.

PR #221, opened 2023-05-07
Let patterns specify a preferred default View

jkbelcher:
As discussed here: https://github.com/titanicsend/LXStudio-TE/pull/220#issuecomment-1537465231

Clicking the View knob to reset or clicking Panic will go to the view preferred by the current pattern.

Had to import the MidiFighterTwister code to change the default behavior of DiscreteParameters incrementing on click instead of resetting.

Edit: A pattern can specify its default view like this:
```
@Override
public String getDefaultView() {
  return "Edges";
}
```


zranger1 (approved):
Very clean and cool functionality!

PR #222, opened 2023-05-07
Delay all file opening until after engine load

jkbelcher:
... to save 5 sec limit imposed by Processing layer.

This solves 2 problems:

1. No dual file open.  Either the most recent file is opened, OR CLI arg filename if one is passed.  Not both.
2. Greatly reduces the risk of hitting the 5 second startup limit by delaying the file open until the first engine loop.

Now when opening a large file on startup (such as AutoVJ.lxp), you will see a lag *after* the UI loads.  And nicely now no red text from the long start time.

### DIFF
--- a/src/main/java/titanicsend/app/TEApp.java
+++ b/src/main/java/titanicsend/app/TEApp.java
@@ -19,6 +19,8 @@
 package titanicsend.app;
 
 import java.io.File;
+import java.io.FileReader;
+import java.io.FileWriter;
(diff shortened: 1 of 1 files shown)


zranger1 (approved):
This is really good! 

And it very likely eliminates the need for a tactic we've been using to delay our OpenGL initialization until everything is else completely loaded.   (If fixed, it'll get us a small performance gain -- can certainly wait 'till after EDC.)

PR #226, opened 2023-05-08
Reduce pattern run loop entry points from two to one

jkbelcher:
@zranger1 TEAudioPattern was adding a second run loop without disabling the first, making it possible for child classes to call the wrong one or to call computeAudio() a second time.  What do you think about this method of forcing them onto the right track?

### DIFF
--- a/src/main/java/titanicsend/pattern/TEAudioPattern.java
+++ b/src/main/java/titanicsend/pattern/TEAudioPattern.java
@@ -67,11 +67,16 @@ public abstract class TEAudioPattern extends TEPattern {
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


jkbelcher:
Ok I'm good with that approach, run() for the management/base classes and runTEAudioPattern() for the patterns.

In that case I think there was a correction on the other PR, I'll comment it...

zranger1:
Hold onto this idea though -- I think we're going to need it to rationalize ConstructedPattern, PatternEffect, FragmentShaderEffect, etc,  when we do the big Views conversion.  In the long run, it just makes things better.

PR #227, opened 2023-05-09
Refresh Vehicle.lxp and default preferences

jkbelcher:
The setup instructions for new users direct people to Vehicle.lxp which contained old patterns that threw errors on startup.  Simplified Vehicle.lxp to 1 channel, 1 instance of TESolidPattern.  Imported colors from AutoVJ.lxp.

Changed preferences on two useful settings, and also the default file to Vehicle.lxp.  It would actually be nice in the dev environment not to wait for AutoVJ.lxp to load every time we refresh from main, as it takes a while and has Live output turned on which spams the home network.


zranger1 (approved):
Working on Windows/Ubuntu clean pull.   Funny,  I'd just gotten out of the habit of using Vehicle.lxp.   It's very handy to have something fast and simple though.

jkbelcher:
Thx for the review!

PR #228, opened 2023-05-09
Add backing opacity adjustment to TE UI controls, default to 100%

jkbelcher:
TE UI panel backings were partially see-through which was visually more apparent after switching to GLPointCloud mode.  Added a UI adjustment for opacity and defaulted it to full opaque.

Before, with default point size of 3:
image

After, with point size of 8:
image
image

*It's actually still showing through a fuzz at low angles, such as the outer edges.  But at least it's a little improved.
image


zranger1 (approved):
This helps, and having the opacity adjustable is a nice touch!

(Can't complain about the edge case see-through behavior when viewed from some angles because I know what's going on under the hood.  IMO, LX falls on the right side of the performance vs absolute accuracy divide here. )

PR #229, opened 2023-05-09
More Noise Patterns - 1 new, 1 updated

zranger1:
Two more noise patterns:

### SmokeShader
SmokeShader is a GL re-translation of Jeff's excellent and artistic Smoke pattern.  It duplicates
the look and behavior of the Java version as closely as possible, at a significantly higher frame rate. 

- Speed: Yes 
- X/Y Offset: No.  This pattern doesn't generate a continuous field.
- Spin/Angle: No
- Brightness: Yes
- Size: Overall Scale
- Quantity: Detail (number of iterations used to build the texture field.)
- Wow1: Mild bass reactivity - used with care, it's like bass reactive silly putty.
- Wow2:  Contrast
- WowTrigger: No

### Triangle Noise
Ok. So, I'm not sure "Triangle Noise" is even an official thing. But... triangles! Noisy triangles!

- Speed: Yes 
- X/Y Offset: Yes. Drift enabled
- Spin/Angle: Yes
- Brightness: Yes
- Size: Overall scale
- Quantity: Detail (number of iterations used to build the noise field.)
- Wow1: Contrast
- Wow2: Foreground color vs. gradient balance
- WowTrigger: No


resources/shaders/smoke_shader.fs
jkbelcher, line 1: This one is great.  Angle rotation on it could be a useful future add.

jkbelcher (approved):
These are both great!  Control response all looks good.

jkbelcher (approved):
These are both great!  Control response all looks good.

PR #230, opened 2023-05-10
Change public methods to protected

jkbelcher:
Fix error in main, `TEPerformancePattern.onActive()` need to be `protected`.

Also change `TEAudioPattern.computeAudio()` to `protected`, doesn't seem like it needs to be `public`.


zranger1 (approved):
Approved!  got this (but probably forgot computeAudio()) in my latest PR too.

PR #233, opened 2023-05-10
Rotation & scaling in TEGradientPattern

zranger1:
Fixed rotation and scaling.  LXParameterizedMatrix seems to want its roll, pitch and yaw to also be CompoundParameters, which we could do, but seems a little awkward for this job.

Switched to plain LXMatrix, and added a call to identity() to reset the transform at the start of every frame b/c all our controls give absolute angles and offsets rather than incremental ones.

Let me know if there's anything else this pattern needs.  Given how sparse our palettes are, we might still want to play with the parameterization a bit for best visuals.


jkbelcher (approved):
It looks great.  Thanks!

PR #234, opened 2023-05-11
Clean up pattern list in UI for performance

zranger1:
Went through the pattern list and cleaned up everything that might cause an exception, a performance issue, or a black car during a show.   Patterns moved and removed are listed in a commented section of TEApp.java, so we can go back and fix them as necessary.

### Here's what got changed, and why:

**Pattern Broken - Throws Exceptions**
- ReactiveHeart 
- SolidPanel 
- SolidEdge 
- BasicImage 
- Bubbles

**Not Currently Functional**

- HandTracker - not functional w/o extra hardware
- ShaderToy -  Not supporting random internet imports for now. 
- Planes - may be useful for test, but can turn the car black

Frame Rate Killers

- Kaleidoscope effect: huge impact on framerate -- drops mine to about 17 
- StarryOutrun - demo made w/two Java shader ports.  Frame rate killer.

### Additional Notes:

**Video Patterns:**
 These actually work if people want to provide videos and use them, but no controls, no reactivity. 

**Moved:**
- AudioTest2 pattern from Test category to Native Shaders Panels so it can be reached w/ less scrolling.  It isn't actually a test program anymore, but I can't rename it w/o breaking AutoVJ.
- Original Smoke pattern moved from Combo FG to TE Examples. It is a nice example of performance optimization of a complex Java pattern.  I'd highly recommed pattern authors take a look at this code. But it is still not quite fast enough for regular use. And we now have SmokeShader for use in perfomance.

**Unfinished Patterns:**
 (Basically everything  not flagged "done" in Notion)  These patterns work, and possibly have useful visuals, but they're non-palette compliant and don't work with the common controls.  Still, might be good for occasional use.

### DIFF
--- a/src/main/java/titanicsend/app/TEApp.java
+++ b/src/main/java/titanicsend/app/TEApp.java
@@ -52,6 +52,7 @@ import titanicsend.model.justin.ViewCentral;
 import titanicsend.output.GPOutput;
 import titanicsend.output.GrandShlomoStation;
 import titanicsend.pattern.TEEdgeTestPattern;
+import titanicsend.pattern.TEMidiFighter64DriverPattern;
 import titanicsend.pattern.TEPanelTestPattern;
 import titanicsend.pattern.ben.*;
 import titanicsend.pattern.jeff.*;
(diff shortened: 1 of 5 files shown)


jvyduna:
Looks good to me. We will have the MidiFighter64 there at EDC; do you still want to leave that one out for various reasons?

zranger1:
Thanks @jvyduna!  If the hardware's there, we should def use it.  MF64 driver restored!

jkbelcher (approved):
Looks good!

PR #235, opened 2023-05-11
Connect Shaders to Views

jkbelcher:
Momentous occasion here, wiring all the awesome shader work to Views.  Views provide normalized coordinates previously calculated in the `Dimensions` class, with the advantage they are interchangeable at runtime.  The groupings from PatternTarget have been carried forward into tags and views.

This touched a lot of patterns and some but not all have been tested.  Try changing the View parameter with one of the affected patterns.

*Note we likely need to make View selection per-pattern instead of per-channel.  That will be a later PR but it will make new patterns always default to their preferred view instead of being stuck to the channel view.  But for now, you can reset the View parameter (double click in UI) to see the pattern's default.


zranger1 (approved):
Approved.   Tested & working w/all currently usable patterns.

Great work! This is really a breakthrough for performance possibilities! 

Notes:
- I'm pushing a tiny change to the code that gets car geometry for shaders to reflect the more regular normalized geometry that view provide.   This keeps ArcEdges and Edgefall properly aligned, and sets us up for future work. 

- In the future, let's provide a way to tag a pattern "Edges only" so it can only select views with edges.   That'll save people from flipping through a lot of black panel views for patterns that use the model's information on position within an edge to render, and don't really work on panels.

jkbelcher:
> This keeps ArcEdges and Edgefall properly aligned, and sets us up for future work.

Great, I wondered about that alignment.

> In the future, let's provide a way to tag a pattern "Edges only" so it can only select views with edges. That'll save people from flipping through a lot of black panel views for patterns that use the model's information on position within an edge to render, and don't really work on panels.

Cool idea.

PR #236, opened 2023-05-12
Add per-channel color swatch selection

jkbelcher:
This is a color safety net.  In short, I'm concerned about the color situation.  No one has ever VJ'd to a strict color palette like this, and I'm wary about fading a second channel in and not being able to achieve contrast if it's the same swatch as the first channel.

This puts a "Swatch" knob on the pattern (pass-through to the channel swatch), giving you the option to spin through all the swatches in the palette.  The first item is the global/current swatch, followed by the list.

image

The brightness parameter got bumped, it is still available in code but it looks like the midi faders will be enough for brightness.

Note swatches other than current will not be oscillated.


zranger1 (approved):
Love it!

PR #237, opened 2023-05-12
Auto VJ template update

worldveil:
Just a few changes in the .lxp


jkbelcher (approved):
All self-contained to AutoVJ template and library, no problem.

PR #240, opened 2023-05-12
Gradient -> TEGradient

worldveil:
And a few other changes. Also using a few of jon's Noise patterns to get rid of builtin LX Noise pattern.

### DIFF
--- a/Projects/AUTO_VJ_TEMPLATE.lxp
+++ b/Projects/AUTO_VJ_TEMPLATE.lxp
@@ -1,6 +1,6 @@
 {
   "version": "0.4.2-SNAPSHOT",
-  "timestamp": 1683857005913,
+  "timestamp": 1683865087009,
   "model": {
     "id": 2,
     "class": "heronarts.lx.structure.LXStructure",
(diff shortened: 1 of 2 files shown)


src/main/java/titanicsend/app/autopilot/TEAutopilotMixer.java
```diff
@@ -77,6 +79,9 @@ public void setFaderTo(TEChannelName name, double faderLevel) {
      */
     public LXChannel getChannelByName(TEChannelName name) {
         if (name == null) return null;
+        for (Map.Entry<TEChannelName, LXChannel> entry : channelName2channel.entrySet()) {
+            TE.log("key: %s, value: %s", entry.getKey(), entry.getValue());
```
zranger1, line 83: You might want to disable this log message before merging.

zranger1 (changes requested):
Looks good!   Can you disable the `TE.log("key: %s, value: %s", entry.getKey(), entry.getValue());`
message in TEAutoPilotMixer before merging?

zranger1 (approved):
Nice!

PR #241, opened 2023-05-12
updating panels ip/channel config + script

theyoffa:
Changes:

- Add script to process backpacks tab of the spreadsheet into update ip/controller panel configs
- Check in updated config
- Update channels per IP to 4 (from 8) and pixels per channel to 250 (from 500) per our new controller design
- Add multi-ip/overflow functionality for panels which have over 1000 pixels

TODO:

- Investigate [this](https://titanicsend.slack.com/archives/C047AD94VCG/p1683244280798419?thread_ts=1683128931.145489&cid=C047AD94VCG) special case 
- Check with Keegan on missing panel (and edge) data


src/main/java/titanicsend/model/TEStripingInstructions.java
```diff
@@ -1,7 +1,7 @@
 package titanicsend.model;
 
 public class TEStripingInstructions {
-  public static final int DEFAULT_CHANNEL_LENGTH = 500;
+  public static final int DEFAULT_CHANNEL_LENGTH = 250;
```
theyoffa, line 4: According to this [thread](https://titanicsend.slack.com/archives/C047AD94VCG/p1683244047978589?thread_ts=1683128931.145489&cid=C047AD94VCG) it sounds we have to address having different length for channels/edges. That said, doesn't look like we have anything longer than 250 in edges.txt and doesn't seem like we use this value for edges anyway. If that's because edges.txt is currently incomplete, we can rename to panel_length.

jkbelcher: Yeah in that thread Keegan says some edges are still in the 400-500 range for pixel count.  So let's split this into something like `DEFAULT_CHANNEL_LENGTH_EDGE` and `DEFAULT_CHANNEL_LENGTH_PANEL`, and we'll need handling to use the two of them.

jkbelcher (changes requested):
Nice, great to be getting these config updates in.  I'm getting an error though...

src/main/java/titanicsend/model/TEStripingInstructions.java
raldi: Note that this constant is within TEStripingInstructions, which relates to the striping of panels. So indeed, if we wanted to be explicit, we could call this class TEPanelStripingInstructions and the constant DEFAULT_NUMBER_OF_PIXELS_TO_USE_WHEN_STRIPING_A_PANEL_BEFORE_ROLLING_OVER_TO_THE_NEXT_CHANNEL. I'll leave it up to you to decide which of those words to keep and which to omit. :D

src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -467,18 +467,26 @@ private static void loadPanels(Geometry geometry) {
...
-          socket.addPanel(p, firstChannelPixel, lastChannelPixel);
-          firstChannelPixel = lastChannelPixel + 1;
+        for (int outputIndex = 0; firstChannelPixel < p.size; outputIndex++) {
+          if (outputs.length <= outputIndex) {
+            throw new RuntimeException("Not enough ips! May require a missing overflow ip");
+          }
+          tokens = outputs[outputIndex].split("#");
+          assert tokens.length == 2 : "Bad panelType: " + outputConfig;
+          String ip = tokens[0];
+          int channelNum = Integer.parseInt(tokens[1]);
```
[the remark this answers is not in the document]
raldi: Panels always start at channel index 0, right? In other words, although we do this:

[controller] - [edge] - [edge]

...we never do this:

[controller] - [edge] - [panel]

or:

[controller] - [panel] - [panel]

...right? If so, we don't need to have ":0" in the panel config because it's always ":0" (and besides, even for edges, ":0" is implied when there's no colon.

resources/vehicle/panels.txt
```diff
@@ -1,45 +1,45 @@
-FA	89-91	89-126	91-126	91->89	unflipped	10.7.?.1#?
-FB	81-89	81-91	89-91	91->89	unflipped	10.7.?.1#?
-FSA	89-125	89-126	125-126	126->125	unflipped	10.7.?.1#?
-FSB	70-81	81-89	70-89	81->89	unflipped	10.7.?.1#?
-FSC	70-81	70-82	81-82	82->81	unflipped	10.7.?.1#?
+FA	89-91	89-126	91-126	91->89	unflipped	10.7.21.211#1:0/10.7.21.212#1:1000
```
raldi, line 1: This seems wrong; 10.7.21.212#1:1000 means, "Part of Panel FA is going to be on Channel 1 of controller 10.7.21.212, 1000 pixels into that controller's output line." That would correspond to something like:

[controller's channel 1] - [edge with 800 pixels] - [edge with 200 pixels] - [this part of Panel FA]

script/import-tsv/import-panels.py
```diff
@@ -0,0 +1,89 @@
...
+    panels_data = [line.strip().split() for line in f.readlines()]
+
+# Read TSV file
+# Assumes a direct export of the "Backpacks" tab of the "TE2 Modules, Panels, and Edges" spreadsheet
+tsv_filename = sys.argv[1]
+with open(tsv_filename, "r") as f:
+    tsv_data = [line.strip().split('\t') for line in f.readlines()]
+
+# Check if TSV begins with the appropriate header
+del tsv_data[0:3]
```
raldi, line 17: This could use a comment like:

```
# The first three lines of that tab are header notes, and we're not
# going to bother asserting that they contain any particular text
```

script/import-tsv/import-panels.py
```diff
@@ -0,0 +1,89 @@
...
+            pixel_start = row[panel_id_column + 1]
+            # current assumptions:
+            #   we always fill all the channels in an IP
+            #   there is max one overflow that starts at 1000
+            # leveraging these assumptions makes for a cleaner/more readable panels.txt, but not hard to adapt by being
+            #   explicit about every channel if necessary
+            if pixel_start == "0" or pixel_start == "1000":
+                if panel_id not in tsv_panel_to_outputs:
+                    tsv_panel_to_outputs[panel_id] = []
+                output = row[0] + "#" + str(channel_number) + ":" + pixel_start
```
raldi, line 46: As noted in the Java commentary, the number after the colon isn't supposed to be the index into the fixture; it's supposed to be the index into the controller channel.

script/import-tsv/import-panels.py
```diff
@@ -0,0 +1,89 @@
...
+
+# Create dictionaries for cross-referencing
+panels_dict = {panel[0]: panel for panel in panels_data}
+
+# Keep track of updates and non-updates
+updated_panels = []
+found_only_in_panels = []
+found_only_in_tsv = []
+
+# Cross-reference and update
```
raldi, line 60: You're doing an impressive job handling the nightmare that is automated config updating from spreadsheet dumps.

src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -467,18 +467,26 @@ private static void loadPanels(Geometry geometry) {
...
-        for (int i = 0; firstChannelPixel < p.size; i++) {
-          int lastChannelPixel = firstChannelPixel + getChannelLength(tesi, i) - 1;
-          if (lastChannelPixel > p.size - 1) lastChannelPixel = p.size - 1;
-          ChromatechSocket socket = GrandShlomoStation.getOrMake(
-                  ip, channelNum + i);
-          socket.addPanel(p, firstChannelPixel, lastChannelPixel);
-          firstChannelPixel = lastChannelPixel + 1;
+        for (int outputIndex = 0; firstChannelPixel < p.size; outputIndex++) {
+          if (outputs.length <= outputIndex) {
+            throw new RuntimeException("Not enough ips! May require a missing overflow ip");
```
raldi, line 474: I used to throw exceptions for bad output configs, but the 2023 pattern programmers have been moving them to warnings, so they can keep writing patterns even after we check in a broken output config. After some consideration, I've come to see the wisdom of this new approach.

src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -467,18 +467,26 @@ private static void loadPanels(Geometry geometry) {
...
+          }
+          tokens = outputs[outputIndex].split("#");
+          assert tokens.length == 2 : "Bad panelType: " + outputConfig;
+          String ip = tokens[0];
+          int channelNum = Integer.parseInt(tokens[1]);
+
```
raldi, line 480: Add:

```
// TODO: Mike really should've added a comment here explaining what's going on. Figuring
// out which Mike is an exercise for the reader.
```

raldi (changes requested):
This is a great first stab at a really hairy task. You're almost there, keep at it!

src/main/java/titanicsend/model/TEWholeModel.java
jkbelcher: And not just for pattern development, but you can keep sending pixels to the car for hardware debugging while you're working on the config for a single item.

src/main/java/titanicsend/model/TEStripingInstructions.java
theyoffa: made the variable name panel specific and leveraged your full name as a comment :)

@justin shouldn't need to do anything for the edges

src/main/java/titanicsend/model/TEWholeModel.java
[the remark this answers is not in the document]
theyoffa: sg, killed it

src/main/java/titanicsend/model/TEWholeModel.java
theyoffa: ugh pattern devs amiright? :)

moved to warn

src/main/java/titanicsend/model/TEWholeModel.java
theyoffa: :) added some comment

theyoffa:
added functionality for our manufacturing defects. unfortunately smaller channel size means we need to be more specific about where they are and will require manual inspection :/. Took a 50/50 shot at guessing each time for now.

raldi:
This moves the ball substantially closer to the end zone. I recommend we merge it in and address the remaining bits in the next PR.

raldi (approved):

src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -466,19 +466,32 @@ private static void loadPanels(Geometry geometry) {
...
-          ChromatechSocket socket = GrandShlomoStation.getOrMake(
-                  ip, channelNum + i);
-          socket.addPanel(p, firstChannelPixel, lastChannelPixel);
-          firstChannelPixel = lastChannelPixel + 1;
+        // keep track of which channel we're on cross-controller so we can use for striping instructions
+        int totalChannelOffset = 0;
+        for (int outputIndex = 0; firstChannelPixel < p.size; outputIndex++) {
+          if (outputs.length <= outputIndex) {
+            LX.log("Not enough ips! May require a missing overflow ip");
+            continue;
```
jkbelcher, line 480: @theyoffa This will infinite loop.

Also recommend `TE.err("...")` instead of `LX.log()`

raldi: Yeah, I think just `break` and leave the rest of the fixture dark.

theyoffa: 🤦‍♂️ good catch, fixed

jkbelcher (approved):

jkbelcher:
Next time I won't do a blocking review when there are Two Mikes on the situation

PR #245, opened 2023-05-13
Refresh remote controls for TEPerformancePattern after file load

jkbelcher:
Force refresh of pattern remote controls after file open.  Also keep the custom UI up to date with changes.

Now we don't have to re-add every pattern to every LXP any time we add/move/remove a parameter.


zranger1 (approved):
Once again, a very nice improvement!

PR #246, opened 2023-05-13
Fix: modify dev tool to add patterns on correct thread

jkbelcher:
Oversight.  Live and learn!  Thanks @worldveil for catching.


jkbelcher:
Reassurance from the `addTask()` method:
image

zranger1 (approved):

PR #248, opened 2023-05-14
2 More Patterns - PowerGrid & SpaceEdges

zranger1:
_Amost missed PowerGrid because of the name - I thought it had something to do with actual power, and didn't want to set the car on fire._   

### PowerGrid (Edges, but nice on panels w/right parameter setup)

- Speed: Yes 
- X/Y Offset: No
- Spin/Angle: No
- Brightness: Yes
- Size: Pulse width 
- Quantity: No
- Wow1: Mode - foreground color or gradient
- Wow2: Background color level
- WowTrigger: No

### SpaceExplosion

- Speed: Yes 
- X/Y Offset: Yes
- Spin/Angle: Yes 
- Brightness: Yes
- Size: No
- Quantity: No
- Wow1: 
- Wow2: Foreground color vs. gradient level
- WowTrigger: No


worldveil (commented):
SpaceExplosionEdges looks great, just tried it out. Couple ideas:
- Is it supposed to work with views? it doesn't seem to
- would be perfect if it had the Wow Trigger for the explosion tho. I could use for FX channel. I wouldn't spend forever on this, just if it's an easy fix

I don't see any changes for Power Grid in this PR?

worldveil:
Also didn't run without the cast to `BoundedParameter`, so just added that.

zranger1:
PowerGrid uses the ShimmeringEffect shader.   It only shows up as PowerGrid in the LX UI.  

The BoundedParameter cast thing is weird, since that super.reset() function explicitly returns a BoundedParameter.  Maybe mvn clean and rebuild.   It works here either way, cast or no cast.

jkbelcher:
0.4.2 returned a LXParameter.  0.4.3 returns a BoundedParameter, you can verify on your official branch here:

https://github.com/jkbelcher/LX/blob/te/src/main/java/heronarts/lx/parameter/BoundedParameter.java#L347

In fact you could probably clone this repo and add it to your IDE for easier debugging.

zranger1:
Pushed a small commit to fix a bug I just found in MatrixScroller.   The Wow1 controlled "blast radius" behavior broke when we switched to Views, and I didn't catch it 'till now.   The fix adjusts the blast dispersion math to use the normalized coordinates that Views provides.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -423,7 +423,7 @@ public BoundedParameter reset() {
             else {
                 spinRotor.addAngle(-this.getValue());
             }
-            return super.reset();
+            return (BoundedParameter) super.reset();
```
jkbelcher, line 426: Not needed.

src/main/java/titanicsend/pattern/yoffa/effect/ShimmeringEffect.java
```diff
@@ -1,27 +1,53 @@
...
+
+        @Override
+        public String toString() {
+            return this.label;
+        }
+    }
+
+    public final EnumParameter<Mode> mode =
+        new EnumParameter<Mode>("Mode", Mode.FOREGROUND)
+            .setDescription("Color Mode");
```
jkbelcher, line 37: Will you add a `setWrappable(false)` on this?  On the MidiFighterTwister that will make it easier to not overshoot.

resources/shaders/space_explosion.fs
jkbelcher, line 1: SpaceExplosion is great.  Any chance of making the explosion start on the beat but keep the travel speed adjustable with the speed knob?

jkbelcher, line 1: IMO speed range could be -1->1 instead of -4->4, with some exponent for finesse.

jkbelcher, line 1: Is the purpose of SpaceExplosionEdges as a different pattern just to allow a different default view?  It seems like once we switch to view-per-pattern that won't be needed.

zranger1: Yes, but it takes some additional code - let me do that, along with @worldveil's one-shot triggered effect in a separate PR later today.

zranger1: We have a ton of patterns like this - with separate "xxxEdges" and "xxxPanels" versions driving the same pattern code on different targets.  We should definitely clean them up for Views.   It'll make our pattern list a lot simpler.

zranger1: Set to -1.5->1.5, exponent 2 for now.  This lets you run one explosion per beat at up about 120 bpm.   

May (almost certainly will) change when I hook it directly to the beat later.

src/main/java/titanicsend/pattern/yoffa/effect/ShimmeringEffect.java
zranger1: Done in most recent commit

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Fixed in most recent commit

worldveil:
PowerGrid looks great!

SpaceExplosionEdges seems to crash my LX UI though -- see attached screenshot

Screenshot 2023-05-15 at 5 47 02 PM

Full error here:

```
[LX 2023/05/15 17:46:51] FATAL ERROR IN LXEngine.run(): Cannot add remote control that does not belong to this component: heronarts.lx.parameter.CompoundParameter@542ed9e0
java.lang.IllegalArgumentException: Cannot add remote control that does not belong to this component: heronarts.lx.parameter.CompoundParameter@542ed9e0
	at heronarts.lx.LXDeviceComponent.validateRemoteControls(LXDeviceComponent.java:140)
	at heronarts.lx.LXDeviceComponent.setCustomRemoteControls(LXDeviceComponent.java:158)
	at titanicsend.pattern.TEPerformancePattern$TECommonControls.setRemoteControls(TEPerformancePattern.java:608)
	at titanicsend.pattern.TEPerformancePattern.lambda$new$0(TEPerformancePattern.java:873)
	at heronarts.lx.LXEngine._run(LXEngine.java:1042)
	at heronarts.lx.LXEngine.run(LXEngine.java:947)
	at heronarts.lx.LXEngine.access$1400(LXEngine.java:62)
	at heronarts.lx.LXEngine$ExecutorService.runLoop(LXEngine.java:693)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

jkbelcher:
That's probably the same situation as the other thread where you can't see the new parameters.  Looks like we'll need to check parameters when loaded from file...

zranger1:
This one's definitely a typo on my part -- try it now!

theyoffa (approved):

PR #249, opened 2023-05-15
Add cast for BoundedParameter

worldveil:
Main was broken without this cast for me. This fixes


jkbelcher:
Sounds like this was an out of sync library.

PR #250, opened 2023-05-15
Add brightness back to a visible UI knob

jkbelcher:
Let there be... not so much light?


worldveil:
Hm, I'm not seeing the knob in Design mode.

zranger1 (approved):
Approved -  Working as expected here + not much code change.

Might be nice to see it somewhere in performance mode too, even if it's not on a knob, just so you can be sure the value is what you think it is.

worldveil:
@zranger1 wait are you seeing it working? I am not. I should be seeing a brightness, knob, yes?



Here I do not.

worldveil:
@zranger1 wait are you seeing it working? I am not. I should be seeing a brightness, knob, yes?

Screenshot 2023-05-15 at 5 39 35 PM

Here I do not.

zranger1:
I've got a functional brightness control.  It's a bit far off to the side, did it just get pushed out of the window?

image

worldveil:
Nice, yep, I see it, Just needed to scroll over I think (lol).

PR #252, opened 2023-05-15
Implement view-per-pattern

jkbelcher:
It turns out we want the View to be adjustable per-pattern to accommodate playlist-style channels such as those used by AutoVJ.

The problem with views per-channel is the channels become locked to various model layouts.  This is fine in the case of "channel 7 is always the outline", but doesn't work in the case of "Channel 2 is high energy patterns, not necessarily all using the same view."  Also, you can't change the view of a channel while it is faded into the mix as there's no transition.  So the channels end up permanently fixed to a view OR channel has to be faded all the way down before view can be changed.  Overall it does seem nice to unshackle channel usage from visual groupings of fixtures.

Turns out this was easier to implement than views per-channel.  I refrained from deleting all the per-channel scaffolding though because... you never know.

A couple other conveniences gained:

- Patterns initialize to their preferred view.
- View selection on each pattern survives file save/load.
- Cleaner implementation since the list of views is available at pattern instantiation time, rather than with channels where there is a delay until the pattern is added to a channel.

NOTE: For now, inserts/deletes from the middle of views.txt should be avoided pre-show as they will bump the view selections out of alignment in existing .lxps such as AutoVJ.  Maybe this can be improved later to save the string instead.

@mcslee Note this interesting evolvement of the feature.

### DIFF
--- a/src/main/java/titanicsend/model/justin/ViewCentral.java
+++ b/src/main/java/titanicsend/model/justin/ViewCentral.java
@@ -23,6 +23,7 @@ package titanicsend.model.justin;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Scanner;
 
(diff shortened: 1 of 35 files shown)


zranger1:
This is really, excellently, amazingly cool, and we really need it!   Just want to check in before approving to make sure there's a good path for AutoVJ to it working without manually reloading every pattern.   Currently, when I start AutoVJ, I can't change view at all until I reload the pattern.  Guessing this is a config file issue?

worldveil:
So yeah when I adjust the first pattern's View, everything works, even after save and re-open. 

But I can't change the views for other patterns -- on mouse down I just see a "0" and it won't change

jkbelcher:
We're not getting the new parameter through to the ui on those other patterns.  Working on it...

jkbelcher:
Taxiing, about to lift off.  That appears to fix but no time to click around.  Try it if this is holding you up!

worldveil (approved):
LGTM, works on save/load too

PR #259, opened 2023-05-17
Update LX libs to 0.4.4-SNAPSHOT

jkbelcher:
Allow modulation of modulator range and sending OSC with String parameters


theyoffa (approved):

zranger1 (approved):
Builds and runs without problems on Windows & Ubuntu here.

Complains about views.txt during startup, but I'm guessing you've got that under control.
`[LX 2023/05/16 21:26:08] Invalid number of columns in views.txt config file, found 2:...`

PR #267, opened 2023-05-17
add alpha values for remaining shaders

theyoffa:
a few of these aren't used but for the ones that are, tested in LX for functionality and layering

feels like there should be a generic way of doing this but for now, meh


zranger1 (approved):
This is fine - given how we're using color, it really is the best way to get a nice looking alpha value that plays well with LX blending. 

We could certainly do this in the fragment shader template, and with a little preprocessing can even avoid stomping shaders that calculate their own alpha - can discuss post-EDC.

worldveil:
Just ran through and tested this as well. Blending looks great.

PR #270, opened 2023-05-17
revert zero byte icon file

jkbelcher:


zranger1 (approved):
Working here!   Good to go if everything's ok at your end.

worldveil:
Just tested on the lighting-1 machine after a restart. We lose the icon in Finder, but that's it! LTGM.

PR #271, opened 2023-05-17
Turn on artnet sequence number

jkbelcher:
Recommended for use by Eli.


worldveil:
What is the actual benefit here? Reluctant to merge until we can test on hardware again.

raldi:
How does this interact with the situation where multiple are broadcasting at the same time?

raldi:
BTW, this is the sort of thing that could go here if you wanted to make it configurable:
https://github.com/titanicsend/LXStudio-TE/blob/main/resources/vehicle/general.txt

jkbelcher:
> How does this interact with the situation where multiple are broadcasting at the same time?

Well it would be hard to make that situation worse.  Behavior would depend on alignment of sequence numbers from the two sources and the controller code.  Sequence numbers in artnet go from 1 to 255 then start over.

raldi (approved):

jkbelcher:
Merging to clean up PRs.  Added a P2 software task to make this configurable in general.txt per raldi's suggestion.

PR #272, opened 2023-05-18
Fix views.txt startup error

jkbelcher:
Fix simple typo.


zranger1 (approved):
Error messages gone,  all views working!

PR #273, opened 2023-05-18
bring back the stars

theyoffa:
- BreathingDots changed to not use a million (or any) LXModulators, which LX seemed to be leaking over time (not that I wasn't abusing them). Brought back the pattern
- BreathingDots and AlternatingDots modified to respect new common quantity param
- StarryOutrun was using the legacy java-ported OutrunGrid pattern (which is super slow) instead of the native shader version. Updated and added back


worldveil:
- BreathingDots doesn't adhere to the Brightness common control. So it's impossible to turn brightness down
- AlternatingDots looks great
- I wish the stars in StarryOutrun would also spin if the spin control is used, would look super cool

worldveil:
- BreathingDots doesn't adhere to the Brightness common control. So it's impossible to turn brightness down
- AlternatingDots looks great
- I wish the stars in StarryOutrun would also spin if the spin control is used, would look super cool

zranger1 (changes requested):
In BreathingDotsEffect.java, line 30:
`int baseColor = pattern.getSwatchColor(TEPattern.ColorType.PRIMARY);`
consider using
`baseColor = pattern.calcColor()` 
instead, to use color from the new color control.

zranger1:
Also I think you're the first person to try using GLSL and Java shaders in the same ConstructedPattern.  Cool!  It works!  Nice design!

theyoffa:
> * BreathingDots doesn't adhere to the Brightness common control. So it's impossible to turn brightness down
made both adhere to brightness
> * I wish the stars in StarryOutrun would also spin if the spin control is used, would look super cool
agree. added a todo for now, prolly not highest pri for edc, but would like to add

zranger1 (approved):

PR #295, opened 2023-05-19
bring back excessive electric dispersion

theyoffa:
it was hooked up to quantity but not enough to get it to fully dissipate in (IMO) the way that looks really cool. when you get it there IMO it's more suited for wow factor so i swapped it for wow2 which ups the haze


zranger1 (approved):
This looks fine - just make sure AutoVJ is ok with the new parameter setup before merging.

(FWIW, when away from actual LEDs, I'm generally super conservative about parameter ranges -- if you're looking at the car and different values make it look cooler, then that's what we need to do!)

theyoffa:
double checked with @worldveil, auto vj should be fine

PR #297, opened 2023-05-19
Set OSC destination address at startup

jkbelcher:
For iPads, now that LaserTask is defaulted to off.


worldveil (approved):
I wonder if we could just do it after, not before. For the next show...

PR #299, opened 2023-05-19
Fix main

jkbelcher:
Fix for space explosion


zranger1 (approved):
Ha!  Just opened a PR for the same thing, w/ same solution.   Yours is first in.

PR #300, opened 2023-05-19
Fixes for views handling changes

zranger1:
Views behavior evolved a bit while this was pending review - fixed to reflect current version.   Really surprised that git merge didn't catch this.


zranger1:
Duplicate of #299

PR #303, opened 2023-05-21
Update NeonTriangles

zranger1:
Add angle, spin and color control support.

Wow2 mixes the rainbow colors back in, just in case!

(we can always do something else later.)


worldveil (approved):
LGTM, I tested

PR #305, opened 2023-05-21
set alpha to one unless black

theyoffa:
We're realizing the blending was actually pretty good even with setting alpha to one and it's not worth losing the brightness of the lowered alpha. But to help out blending a bit, let's set alpha to 0 for black space.


zranger1 (approved):
Code looks fine.  Approving for expediency, but for GL shaders, it would've been easier to just set finalColor.a = 1.0 as the last line of the shader template! 

(Also, I haven't had time to thoroughly test, so keep an eye on these things when you fire 'em up.   There may be a few shaders lurking  that actually used alpha *as* brightness. )

theyoffa:
> it would've been easier to just set finalColor.a = 1.0 as the last line of the shader template!

wouldn't this step on shaders that are actually using alpha appropriately?

zranger1:
Not 100% sure, but don't think it would affect our stuff.  (except maybe having zero brightness == 0 alpha)

Shadertoy effectively ignores alpha - you can't use it as brightness, and we're all using it as a model, so all our shaders multiply color to get final output level and alpha is mostly afterthought.    I'll do the experiment later if I still remember!  :-)

PR #307, opened 2023-05-21
Fix Alpha/Brightness for last few shaders.

zranger1:
Alpha/brightness PR turned smokeshader to solid color.  One-line fix makes everything OK again!

Edit:  After testing also found that RadialSimplex and SimplexPosterized shaders were also affected.  Added fixes to this PR.

Only the GLSL shaders, have changed.   Once merged and pulled, you should be able to just delete and reload innstances of the affected shaders in their channels to get this working, even mid-show.  (Or just reload the .lxp)  No need to restart or rebuild LX.


worldveil:
Just tested these -- they look way better! Excellent PR and catch. Merging.

worldveil (approved):

PR #308, opened 2023-05-21
Extra Bonus Pattern : RainBands

zranger1:
I thought we could use something with simple shapes and strong vertical and horizontal movement, so I made this.   Check it out if you've got time tonight - video link also on slack.

https://www.youtube.com/watch?v=KbOG8PgmH6I

Finished while delayed at airport last night, on craptastic laptop, so I'll trust you guys on whether the graphics are ok.

### RainBands
Rain, rockets, stripes, horizontal, vertical and angled bands...

- Speed: Yes
- X/Y Offset: Yes
- Spin/Angle: Yes
- Brightness: Yes
- Size: Overall Scale - height & width of bands
- Quantity: Band density
- Wow1: Relative y scale - "tail" length
- Wow2: Foreground color vs. gradient level
- WowTrigger: No


worldveil:
One request / question: when the Size knob is moved, is there a way to not have it re-adjust all the positions? 

One modulation we found to be super nice was to attach the bass EQ band to parameters like Size, but here it would cause a lot of stutter-iness.

zranger1:
Scaling behavior improved in latest commit.   See the video at:

https://www.youtube.com/watch?v=tnJZezJcBkE

theyoffa (approved):
very cool! i feel like there's a lot we could do with beat reactivity here if we wanted. eg alternating bands extending and retracting on each beat. could also make it into an equalizer, though maybe too tacky.

PR #315, opened 2023-06-04
Move GLAutoDrawable to later to avoid startup crash

jkbelcher:


zranger1 (approved):
Fine for now.  We need to figure out what really happened with this though.

PR #316, opened 2023-06-06
Change dev tool to add only TEPerformancePatterns

jkbelcher:
Also sorts alphabetically.  Useful for setting up for show.


zranger1 (approved):
Code looks fine, tested on a single new channel, and on EDC showcase w/8 other channels running faders full up.  Added channel, added patterns, deleted channel,  added patterns multiple times, deleted channel then added it again...  Basically thrashed the heck out of it.   

Rendering stops briefly when loading, but otherwise no problems encountered.

PR #319, opened 2023-06-20
Added beacon

raldi:


src/main/java/titanicsend/model/TEWholeModel.java
```diff
@@ -499,18 +500,28 @@ private static void loadLasers(Geometry geometry) {
...
       String line = s.nextLine();
+      if (line.startsWith("#")) continue;
       String[] tokens = line.split("\t");
       assert tokens.length == 4 : "Found " + tokens.length + " tokens";
 
       String id = tokens[0];
       int x = Integer.parseInt(tokens[1]);
       int y = Integer.parseInt(tokens[2]);
       int z = Integer.parseInt(tokens[3]);
-      String[] tags = new String[] { id };
```
jkbelcher, line 509: Bringing the IDs through as tags is useful for filtering with views.  I would leave this part in.

src/main/java/titanicsend/model/TELaserModel.java
```diff
@@ -15,12 +15,14 @@ public class TELaserModel extends TEModel {
   public int color;
   public String id;
   public LaserControl control;
+  public int thickness;
 
-  public TELaserModel(String id, double x, double y, double z, String ... tags) {
-    super("Laser", makePoint(x, y, z), tags);
```
jkbelcher, line 20: Would keep `tags` in here.

src/main/java/titanicsend/pattern/justin/TESolidPattern.java
```diff
@@ -25,6 +25,9 @@ protected void runTEAudioPattern(double deltaMs) {
 
       colors[p.index] = color1;      
     }
+
+    // Draw lasers in the GUI with the new colors
+    this.updateVirtualColors(deltaMs);
```
jkbelcher, line 30: A better home for this would be `TEAudioPattern.run(deltaMS)`, as a last line after `runTEAudioPattern(deltaMs)`.  Then this code won't need to be replicated across every pattern.

resources/vehicle/lasers.txt
```diff
@@ -1,12 +1,15 @@
...
-HP4	-1219200	6387435	-5202305
-AS1	-1346200	3962400	7366000
-AS2	-1286736	3965520	3851275
-AS3	-1178926	5080000	3185822
-AS4	-1219200	5080000	2123000
-AS5	-1219200	5080000	-2123000
-AS6	-1219200	5080000	-3185822
-AS7	-1286736	3965520	-3851275
-AS8	-1346200	3962400	-7366000
+# Beacon
```
jkbelcher, line 1: Fine for getting started but next up the config for beacon(s) should have their own `beacons.txt` file.  There will be other columns to add that are beacon-specific and not relevant to lasers.

src/main/java/titanicsend/app/TEVirtualOverlays.java
jkbelcher, line 1: Can we go ahead and add BooleanParameters to this file for `lasersVisible` and `beaconVisible`, maybe default beacon visibility to false until the orientation is meaningful.

jkbelcher (changes requested):
The visual preview looks great.  Useful too for choosing a placement on the car and setting range limits to avoid shining into the frame.

Added a couple fine-tuning comments.  Could use a merge main to fix the compile error.

raldi:
Ryan, let me know when you're free for a tour of this code; I'd like your input on whether you'd like to keep going with this PR or go a different path.

resources/vehicle/lasers.txt
worldveil: like what? just curious

jkbelcher: Yaw, pitch, roll, range constraints, artnet settings

src/main/java/titanicsend/app/TEVirtualOverlays.java
jkbelcher: Disregard this one, TEVirtualOverlays got moved around in Chromatik

jkbelcher:
Most of this is outdated but it could be useful to reference Mike's math when adding the beacon previews into Chromatik UI.

PR #320, opened 2023-06-20
Created channel_report.py

raldi:


raldi:
I added a column to panels.txt with the number of pixels (previously just calculated by LX on startup), added a check to LX to print an error if it doesn't match its calculation (which is a quick way to calculate the value for new panels), and wrote a script called channel-report.py that prints output showing what's on each controller channel, e.g.:

10.7.6.20:
  1: None
  2: None
  3: None
  4: 39-101, 39-121, 44-101
10.7.6.21:
  1: None
  2: SUA_1000
  3: SUA_1250
  4: SUF_1000

You can pass a module number on the command line to just see the controllers in that module, or leave it blank to see them all.

script/vehicle-testing/channel-report.py
```diff
@@ -0,0 +1,83 @@
...
+    num_pixels = int(num_pixels_str)
+    num_channels = math.ceil(num_pixels/PIXELS_PER_PANEL_CHANNEL)    
+    if '?' in outputs_str:
+      continue
+    outputs = list(outputs_str.split("/"))
+    ip = None
+    channel = 5
+    for i in range(num_channels):
+      if channel > 4:
+        ip, channel_str = outputs.pop(0).split("#")
```
fakeplant, line 67: Do you plan to add support for complex mapping after this is merged?

raldi: That PR is standing by for pushing as soon as this one’s merged.

fakeplant: Sorry, to clarify, I don't see complex panel mapping working with this script, and don't see any code in the other branch. But from an order of operations, i was able to run this branch fine. so LGTM.

Just need to add hyphen support in other branch once this is in.

raldi: Can you make that LGTM official? I can’t merge until someone clicks Approve.

fakeplant (approved):
LGTM.

PR #321, opened 2023-06-21
Support for complex channel configs

raldi:


fakeplant:
@raldi this probably needs a rebase w/ main and the new channel-report.py script needs support for new panels.txt config (channel ranges)

raldi:
We don't actually have any channels configured yet that make use of the new functionality; I'd like to get this PR in and then do the channel-report.py updates in the next PR, to keep any one PR from getting too complex.

fakeplant (approved):

PR #324, opened 2023-06-24
Fix context leak warning

zranger1:
This should fix the context leak warning on startup, and speed up lxp load times considerably.


jkbelcher (approved):
Fixes the context leak messages and EDC_Showcase.lxp now loads without issues.  Nice!

PR #332, opened 2023-07-01
RFC: Codifying non-implemented params

andrewlook:
@theyoffa explained to me that the task [Codifying patterns with non-implemented parameters](https://www.notion.so/titanicsend/Codifying-patterns-without-implemented-parameters-2d3005944c43486c9445c5b00772bc89?pvs=4) is partly about tracking which patterns have non-implemented params so we can go back and update them, and partly about signalling to the UI that a pattern isn't used so that the VJ can see clearly that a parameter won't work on a given pattern.

He mentioned that there's a standard order/layout to the basic params on the midifighter controllers, so we probably don't want to fully delete those controls, but we could prefix the controls with an `[X]` or something similar so it's at least visible in the UI.

I picked a pattern at random and noticed that the `iSpeed` parameter wasn't implemented in the fragment shader file, so updated the label on that one.

Here's an example of what the UI looks like afterwards:
Screen Shot 2023-07-01 at 2 25 56 PM

I just wanted to check that this approach of modifying the label wasn't totally off-base for what we're going for here - I'm not sure if there's a better approach to signal TouchOSC or to hide the params, happy to adjust if necessary.

If this looks like I'm on the right track, I can start going through more thoroughly and cataloging the actual params that aren't implemented. If it's helpful to compile those into a spreadsheet so we can project-manage the process of actually changing the patterns to support those parameters, I can do that too. Since `setLabel` is a new function, it should be easy enough to find the occurrences and aggregate them.

Side note: `setLabel` is copy-and-pasted from `setRange` - there's probably a cleaner way to de-duplicate that code, and I'm happy to change that if it's preferable. Just wanted to make sure I was on the right track first.


theyoffa:
Thanks for picking this up!

- In terms of the user facing outcome, this is exactly what I had in mind.
- One way to achieve this is to go through the list and update as appropriate. Another way that might be easier would be a would be to dynamically detect whether the parameter has been implemented. Eg for all shaders without iSpeed present in their .fs file, update the speed label. Could also leverage to autogen the TODO list.
- Yeah, seems there's no way to update the range or label after creating the object. I guess could abstract out a 'copyParamsFromControl' method.

zranger1:
Note:  Don't flag shaders for lacking iSpeed.  Unless they're doing something unusual, they don't use it.

The shader system uses the speed control to calculate a variable rate time, which gets passed to shaders as iTime.  This gives smooth control over pattern speed, forward and backward, without the shader having to have any additional code to support that.

src/main/java/titanicsend/pattern/jon/Electric.java
```diff
@@ -16,6 +16,7 @@ public class Electric extends TEPerformancePattern {
     public Electric(LX lx) {
         super(lx, TEShaderView.DOUBLE_LARGE);
 
+        controls.setLabel(TEControlTag.SPEED, "[X] Speed");
```
zranger1, line 19: Not needed.  This pattern supports the speed control via the normal path for shaders.

worldveil:
Yep, to crystalize here, you mentioned two requirements:

1. `tracking which patterns have non-implemented params so we can go back and update them`
2. `signalling to the UI that a pattern isn't used so that the VJ can see clearly that a parameter won't work on a given pattern`

but one more is

3. whatever we do here can be picked up in @jkbelcher's TouchOSC code

src/main/java/titanicsend/pattern/jon/Electric.java
andrewlook: got it, thank you for clarifying!

andrewlook:
> Yep, to crystalize here, you mentioned two requirements:
> 
> 1. `tracking which patterns have non-implemented params so we can go back and update them`
> 2. `signalling to the UI that a pattern isn't used so that the VJ can see clearly that a parameter won't work on a given pattern`
> 
> but one more is
> 
> 3. whatever we do here can be picked up in @jkbelcher's TouchOSC code

Thanks for adding that note - is there anywhere I can learn about what it takes to hook into touchOSC?

andrewlook:
> Thanks for picking this up!
> 
> * In terms of the user facing outcome, this is exactly what I had in mind.
> * One way to achieve this is to go through the list and update as appropriate. Another way that might be easier would be a would be to dynamically detect whether the parameter has been implemented. Eg for all shaders without iSpeed present in their .fs file, update the speed label. Could also leverage to autogen the TODO list.
> * Yeah, seems there's no way to update the range or label after creating the object. I guess could abstract out a 'copyParamsFromControl' method.

This is great, thank you - I think it should be straightforward to take a pass through the shaders and flag them, maybe a simple python script could do the trick. I'll take @zranger1's  notes into account (specifically that speed + spin/angle aren't expected to be in the shaders).

@theyoffa To be clear, is the scope of this work only to look at shaders, or should I look at Java patterns too?

src/main/java/titanicsend/pattern/jon/Electric.java
zranger1: Cool, and thanks for doing this.   One more thing I forgot:  shaders also automatically get X offset and Y offset as the vec2 uniform ```iTranslate```.   I think that's all of 'em that don't directly correspond to controls!

zranger1:
I'd say look at Java patterns too, especially given that we want to tag unused parameters on the iPads.   There are lots - probably the majority - that don't support all the common controls.

theyoffa:
> simple python script

i think ideally we would use java to update it at build time rather than manually marking a one-off. This would make it forward proof. Happy to help with this if you need.

jkbelcher:
Ideally a mark is a boolean flag on / associated with a parameter.  Second best option is a list of unused parameters returned by the pattern.  The reason #2 is less ideal is it requires comparing the two lists to get the outcome.

The UI label (and TouchOSC behavior) would respond to the flag, rather than the source of truth being the [X] in the label.

script/shader-mgmt/palette_usage.py
```diff
@@ -0,0 +1,147 @@
...
+    # iSpeed is handled by the system - time is modulated before being passed in as a uniform
+    "iSpeed",
+    "iRotationAngle",
+    "iSpin",
+    "iTranslate",
+]
+
+STANDARD_SHADERTOY_UNIFORMS = [
+    "iTime",
+    "iResolution",
```
zranger1, line 35: iResolution is also optional, although most shaders use it to rescale the incoming coordinates.    
iMouse is always vec4(0.0) -- it's not used.  We could probably remove any code in shadertoy-derived shaders that uses it.

script/shader-mgmt/palette_usage.py
```diff
@@ -0,0 +1,147 @@
...
+    "iWow2",
+    "iWowTrigger",
+]
+
+TE_CONTROLS_IMPLICIT = [
+    # iSpeed is handled by the system - time is modulated before being passed in as a uniform
+    "iSpeed",
+    "iRotationAngle",
+    "iSpin",
+    "iTranslate",
```
zranger1, line 30: For most shaders, setting x and y offsets w/iTranslate is automatically handled by the framework.  So unless a shader is doing something special (like the handful of patterns derived from the DriftEnabledPattern class) you probably won't see iTranslate in shader code.

script/shader-mgmt/palette_usage.py
```diff
@@ -0,0 +1,147 @@
...
+    "iColorRGB",
+    "iColorHSB",
+    "iColor2RGB",
+    "iColor2HSB",
+]
+
+TE_CONTROLS_EXPLICIT = [
+    "iScale",
+    "iQuantity",
+    "iBrightness",
```
zranger1, line 19: iBrightness is handled automatically by the framework, and shouldn't normally appear in shader code.

src/main/java/titanicsend/pattern/jon/Electric.java
andrewlook: thanks for clarifying this - if I understand correctly, some patterns can ignore `iTranslate` if they `#define NOTRANSLATE`, is that right? I'll ignore those for now though unless that's a problem.

script/shader-mgmt/palette_usage.py
andrewlook: got it, thanks - I was making an exhaustive list of the controls in my script just to understand the shaders better, but these ones weren't flagged. (I only searched/replaced `TE_EXPLICIT_CONTROLS`)

script/shader-mgmt/palette_usage.py
andrewlook: that's good to know, thank you

script/shader-mgmt/palette_usage.py
andrewlook: OK thanks, I'll remove this from the list since I actually had this down as one of the flagged explicit controls.

Just curious, where does that brightness adjustment happen?

src/main/java/titanicsend/pattern/jon/Electric.java
zranger1: Yes, patterns derived from the java class DriftEnabledPattern use the offset controls to control movement in X and Y, not just offset.   This is only used by a few "drifting through a noise field"-type patterns at the moment.   So all shaders do *something* with iTranslate whether it's in their code or not.

zranger1:
Sorry to drop all this pattern code lore on you all at once, but also glad we get a chance to explain it to someone!

andrewlook:
> Ideally a mark is a boolean flag on / associated with a parameter. Second best option is a list of unused parameters returned by the pattern. The reason #2 is less ideal is it requires comparing the two lists to get the outcome.
> 
> The UI label (and TouchOSC behavior) would respond to the flag, rather than the source of truth being the [X] in the label.

@jkbelcher That makes sense to me. I was assuming that I'd call some function affecting touchOSC in relation to the parameter in question, rather than having it programmatically look for an `[X]`.

I was imagining something along the lines of:
```
controls.setLabel(TEControl.WOWTRIGGER, "[X] WowTrigger");
tellTouchOscThatControlIsUnused(TEControl.WOWTRIGGER);
// or
tellTouchOscThatControlIsUnused(controls.getValue(TEControl.WOWTRIGGER));
```

But if I'm understanding you correctly, it'd be better to subclass Parameter or add a boolean flag to the base class, and we could configure TouchOSC to look for that flag?

andrewlook:
> I'd say look at Java patterns too, especially given that we want to tag unused parameters on the iPads. There are lots - probably the majority - that don't support all the common controls.

that's good to know - I'm assuming you mean only things subclassing TEPerformancePattern (or ConstructedPattern) since those classes are the ones expected to have the common controls? It seems like there are some patterns that, say, directly extend TEAudioPattern and just have 3 controls.

Either way maybe I can address the Java patterns in a future PR since there's already a fair amount going on here?

andrewlook:
> > simple python script
> 
> i think ideally we would use java to update it at build time rather than manually marking a one-off. This would make it forward proof. Happy to help with this if you need.

Ah okay. 

Two questions:
1. Is this a blocker to merging / should I drop the scripting approach entirely?
2. Do you mean "build time" as in maven build/java compile time? Or like shader parsing/compilation-time after java runtime is active?

If you have a pointer to where this might go / or want to help that'd be great. My only real question is how I could get the information from shader parser to a context where we can update the Pattern+Controls (would it be inside `NativeShaderPatternEffect` where there is a reference to `this.pattern` available?).

zranger1:
On Java patterns: There are a few java leftovers that we could probably convert to TEPerformance patterns, and I'll make a low priority work item to go through those as time permits, but generally, yes - only TEPerformance patterns and things derived from them use the common controls.   

From a VJ UI perspective, it'd be super nice to have inactive controls in the Java patterns [x]'ed out so the VJ knows they won't work.  And this is definitely scriptable too, but for now, we'll probably have to do it more on a case-by-case basis.

zranger1:
On when to do this:  It gets really complicated if we try to do it at run time during or post startup.  My strong preference would be to make it a build-time script of some sort.

zranger1:
I like the script approach to finding missing controls -- just brainstorming implementation here.  Maybe...

- we add a function that runs as part of common control initialization (```AddCommonControls()```) that reads the list of missing controls from a text or JSON file and changes the labels.

That way, we could run the script to generate that file whenever -- either during build, or at startup time, and every channel that used a given pattern would automatically get the correct label changes.

andrewlook:
> I like the script approach to finding missing controls -- just brainstorming implementation here. Maybe...
> 
> * we add a function that runs as part of common control initialization (`AddCommonControls()`) that reads the list of missing controls from a text or JSON file and changes the labels.
> 
> That way, we could run the script to generate that file whenever -- either during build, or at startup time, and every channel that used a given pattern would automatically get the correct label changes.

That approach makes sense to me. I'm signing off for the evening now, but if there's no major objections I can tackle that soon, next time I have a spare cycle.

andrewlook:
> > Ideally a mark is a boolean flag on / associated with a parameter. Second best option is a list of unused parameters returned by the pattern. The reason #2 is less ideal is it requires comparing the two lists to get the outcome.
> > The UI label (and TouchOSC behavior) would respond to the flag, rather than the source of truth being the [X] in the label.
> 
> @jkbelcher That makes sense to me. I was assuming that I'd call some function affecting touchOSC in relation to the parameter in question, rather than having it programmatically look for an `[X]`.
> 
> I was imagining something along the lines of:
> 
> ```
> controls.setLabel(TEControl.WOWTRIGGER, "[X] WowTrigger");
> tellTouchOscThatControlIsUnused(TEControl.WOWTRIGGER);
> // or
> tellTouchOscThatControlIsUnused(controls.getValue(TEControl.WOWTRIGGER));
> ```
> 
> But if I'm understanding you correctly, it'd be better to subclass Parameter or add a boolean flag to the base class, and we could configure TouchOSC to look for that flag?

@jkbelcher I looked through the code a bit more.

It looks like `TEControl` will be the easiest place to add a boolean flag marking a param as used or not used, without making a new subclass of `LXListenableNormalizedParameter`

Pros: we can set this for most of the common controls that are relevant for shader patterns at least, and I can programmatically update the label to include `[x]` without much extra code.

Cons: won't work for `TePerformancePatter.{viewParameter, color, swatchPerChannel}` - for those we can update the labels but I'm not clear on how we'd have some boolean variable easily accessible by TouchOSC built directly into the parameter/control. I could keep a separate hashmap on the pattern like `TEControlTag -> boolean`, but then TouchOSC would have to check that explicitly.

Thoughts?

andrewlook:
@zranger1 @jkbelcher I think I got something working that addresses your latest round of feedback on this PR.

It's a little messy in `addCommonControls` but I can refactor that out if we've decided the approach is worthwhile.

So far, I have a python script that scans all native shaders, and saves a file called `resources/pattern/missingControls.json`. This has a field `uses_palette` which is true if the shader uses at least one of the `iColor*` uniforms. There is also a list of `TEControlTag` values for which the control is not used by the shader.

`TEPerformancePattern.addCommonControls` loads up this JSON file, compares the classname of the pattern to the list of patterns related to native shaders, and makes the necessary control updates. For controls that have a `TEControlTag`, there's a hashmap called `unusedControls` that can be referenced programmatically (by TouchOSC) to check if a control is used.

Then `unusedControls` tells us whether to set the `[x]` label for all `TEControlTags`.  For example, here's `NeonBlocks`:
image
image


Caveat: `color` doesn't have a `TEControlTag`, so for that one I just set the label (I'm open to suggestions about how to make that mark easily accessible to TouchOSC). For example, here's `NeonHeartNative`:
image
image


Side note: at some point, I hit a compilation error about `setUnits` and `setExponent` not taking a first argument for `TEControlTag` (see this commit https://github.com/titanicsend/LXStudio-TE/pull/332/commits/bedfdf0c0c7e0a139445ed2dc3487aaab599e6bf ). I'm not sure if this was due to an LX library version, or if it was something about the `TEPerformancePattern.setRange/updateParam` code. If someone else could check that this code works for them, that'd be really helpful.

jkbelcher:
Sorry for the delay here, have been focused on other code and there has been a lot of action in this PR.  Looks like this has been a good orientation to TEPerformancePattern!

I think the map lookup is good.  I'd like to make some adjustments with how it comes through to the user though.  Will follow up when I have time, hopefully in the next day or 2.

jkbelcher:
PR #354 shows a way to hide the UI controls without so much impact on the parameters and pattern classes themselves.  Here's how it looks with the Size control hidden:
image

jkbelcher:
Apologies again this is coming to you late.  Just haven't had time to get eyes on this.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -540,43 +564,63 @@ public TECommonControls setGetterFunction(TEControlTag tag, _CommonControlGetter
...
             LXListenableNormalizedParameter newControl;
             if (oldControl instanceof CompoundParameter) {
-                newControl = (CompoundParameter) new CompoundParameter(oldControl.getLabel(), value, v0, v1)
-                    .setNormalizationCurve(((CompoundParameter)oldControl).getNormalizationCurve())
-                    .setExponent(oldControl.getExponent())
-                    .setDescription(oldControl.getDescription())
-                    .setPolarity(oldControl.getPolarity())
-                    .setUnits(oldControl.getUnits());
+                newControl = (CompoundParameter) new CompoundParameter(label, value, v0, v1)
+                        .setNormalizationCurve(((CompoundParameter)oldControl).getNormalizationCurve())
```
jkbelcher, line 571: It looks like a refactor tool incorrectly added extra spaces to a bunch of lines here.  Let's try to avoid running refactors on big files mixed in with code changes... it hides the real changes.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -610,20 +663,67 @@ public TECommonControls setUnits(TEControlTag tag, LXParameter.Units units) {
...
+            String packageAndClassname = pat.getClass().getName();
+            if (packageAndClassname.contains("$")) {
+                String[] parts = packageAndClassname.split("\\$");
+                currClassname = parts[parts.length - 1];
+            } else {
+                String[] parts = packageAndClassname.split("\\.");
+                currClassname = parts[parts.length - 1];
+            }
+
+            // load the missing controls file
```
jkbelcher, line 678: `addCommonControls()` is called for every pattern constructor.  That happens both at project file open and during runtime when a new pattern is added to the channel.  We don't want to read from file every time this happens.  MissingControls.json only needs to be read once at application startup.  It should live somewhere global for reference, not within a pattern instance.

src/main/java/titanicsend/pattern/jon/ElectricEdges.java
```diff
@@ -16,6 +16,9 @@ public class ElectricEdges extends TEPerformancePattern {
     public ElectricEdges(LX lx) {
         super(lx, TEShaderView.ALL_EDGES);
 
+        markUnusedControl(TEControlTag.WOWTRIGGER);
+
+
```
jkbelcher, line 21: Reduce these to single blank lines to match existing styling.  Pretty uncommon for us to use double blank lines within a method.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -492,6 +515,7 @@ public double getValue(TEControl cc) {
 
 
         private final HashMap<TEControlTag, TEControl> controlList = new HashMap<TEControlTag, TEControl>();
+        private final HashMap<TEControlTag, Boolean> unusedControls = new HashMap<TEControlTag, Boolean>();
```
jkbelcher, line 518: Looks like this could be a List<> instead of a Map<,>.  Currently the only boolean value you're saving into it is True.

jkbelcher (commented):
Ok lots of good work here, appreciate you diving in.  Check the PR... I think a lot of this can be isolated in the UI layer.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
andrewlook: thanks! I extracted it into `titanicsend.util.MissingControlsManager` (since it looked like `titanicsend.pattern.util` was for actual "utility patterns" not utility code related to patterns). Happy to move it somewhere else if that's better.

Also, I made it a threadsafe singleton - I wasn't sure if LX will load patterns in parallel, but it seemed possible, so I wanted to make sure we're not reading the same JSON file twice in parallel. LMK if that guess was wrong and I can adjust as needed).

src/main/java/titanicsend/pattern/TEPerformancePattern.java
andrewlook: thanks for catching that, that annoys me too when it makes the code harder to review. Do you know if there are any preferred linting settings for the codebase? If not I can just try to double-check my changes and make sure no extraneous whitespace is getting in the way.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
andrewlook: Cool, done - I went with a `HashSet` since it seems like we just need a quick check for existence.

andrewlook:
> PR #354 shows a way to hide the UI controls without so much impact on the parameters and pattern classes themselves. Here's how it looks with the Size control hidden: image

This PR was really helpful, thanks - here's what I did to integrate the idea:
- replaced `Set<TEControlTag> unusedControls` with `Set<LXParameter> unusedParams` since `UITEPerformancePattern` seems to speak params rather than `TEControlTag`. Also, not every LXParameter we use has a `TEControlTag` - I'm not exactly clear on why, but using `LXParameter` in the `Set` seems more flexible (as long as the `LXParameters` don't change between when they are set and when they are looked up - but I think they're immutable and that all of the `setLabel` or `setRange` calls happen before we mark the parameter as unused.
- `UITEPerformancePattern` checks this set when calling `hideDeviceControls`

A couple questions I had about how to proceed:
1. If we're using this code to hide the UI controls, **can I get rid of the code that sets the `[x]` labels on the controls?** just double-checking, since I'm not sure if there are other ways the UI might show or up or if both approaches are still necessary.
2. I used `Set<LXParameter> unusedParams` instead of `Set<LXNormalizedParameter>` so that I could add `TEColorParam` to the list as well. But when I run it, I still see Color parameter with an `[x]` in the label. I'm assuming that `UITEPerformancePattern` doesn't deal with color or somehow treats it differently. **Should I stick with the `[x]` label for color or is there a better way that I could make the color parameter `null` as we did for the other controls?**
3. This question isn't a blocker, but I'm curious about what `setRemoteControls()` is doing - does this affect only TouchOSC or does it affect the regular LX UI as well? It seems that `UITEPerformancePattern` is listening for changes in the remoteControls, so I'm wondering if that their main purpose or if it does something else too?

andrewlook:
A couple other small notes:
- I added some exception handling in `MissingControlsManager`, just as a sanity check so things don't crash if the JSON file was invalid for whatever reason.
- I added a maven exec step, in case we ever wanted to automate updating `missingControls.json` as part of a CI build or something. `mvn exec:exec@MarkUnusedControls`

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher: None written down.  Indentation is either 2 spaces or 4 spaces, whatever matches the file.  We decided earlier this year the standard should be 2 spaces which matches LX convention, but there are a ton of legacy files using 4.

jkbelcher:
Yes the list of LXNormalizedParameter is more convenient than TEControlTag, for direct comparison to the list of remote controls.  Good.

BTW - avoid `force-push` once reviews have started or anyone else has pulled your code.  Two reasons:
1. It breaks the links from existing&prior github conversations to the locations in Files Changed.  If you scroll through files changed, you'll see that open conversations from before the force-push are now not visible.
2. Risk of getting out of sync. If any of us have pulled the branch, we can't update with a `git pull`.  We have to delete the branch and start over.

Yes, if you're on board with this method to hide controls, then it's safe to remove the code to set labels to "[x]..".

> I used Set<LXParameter> unusedParams instead of Set<LXNormalizedParameter> so that I could add TEColorParam to the list as well. But when I run it, I still see Color parameter with an [x] in the label. I'm assuming that UITEPerformancePattern doesn't deal with color or somehow treats it differently. 
Stick with <LXNormalizedParameter> for the list because it matches the remote controls list type.  If you look at `TEPerformancePattern.setRemoteControls()` you can see the first two parameters are not color but rather subparameters of color.   They are `color.offset` and `color.gradient`.  If you include those subparameters in the unused list then they can be correctly matched to the remote controls.  I'd be careful hiding those... seems like most patterns should be pulling a color.

jkbelcher:
> This question isn't a blocker, but I'm curious about what setRemoteControls() is doing - does this affect only TouchOSC or does it affect the regular LX UI as well? 

It sets the list of remote controls on the pattern, which is mainly intended for consumption by midi surfaces.  However in TE we're doing an implementation that I've come to prefer on personal projects, which is to make the UI elements line up with the MidiFighterTwister knobs.  For a VJ-oriented setup it makes things easier to follow if the UI matches the hardware.

> It seems that UITEPerformancePattern is listening for changes in the remoteControls, so I'm wondering if that their main purpose or if it does something else too?

It's complicated.  When a pattern is loaded from a saved project file the custom remoteControls get applied After the pattern is loaded.  I hope to modify the LX approach to this in the future and allow pattern classes to modify the default remote controls so the custom remote controls can be left to the user.

These are great questions.  Thanks for absorbing the project in such detail.

andrewlook:
Thanks for the heads-up re: force-push - the convention totally makes sense, especially if reviews are in progress / other people have pulled. I'm happy to squash and rebase whenever we're getting ready to merge so the list of commits on `main` is less hectic, but I'll hold off on that until you give the OK.

I'm glad the `LXNormalizedParameter` choice makes sense. For the everything except `color`, I'll remove the `[x]`. If hiding color/gradient/swatch is too aggressive, I can just keep the "x" on those params for now - I commented out the lines to hide those controls, but lmk how you'd like to proceed there. Here's the commit containing those 2 tweaks: https://github.com/titanicsend/LXStudio-TE/pull/332/commits/2dcfda2c5ce15d3bfd6b9ceb410f74467a423406

Re: color usage - here's a list of all the shaders that don't use color palette. Seems like 6 registered patterns total.

|shaderName|patternClasses|usesPalette|
|------------ | --------------| -----------|
|neon_heart.fs|['NeonHeartNative']|False
|pulsing_heart.fs|['PulsingHeart']|False
|warp.fs|['Warp']|False
|synth_waves.fs|['SynthWaves']|False
|jet_stream.fs|['JetStream']|False
|neon_cells.fs|['NeonCells']|False

There are also 5 more shaders that don't use color palette, but they're not mapped to any Java patterns as far as I can tell.

|shaderName|patternClasses|usesPalette|
|------------ | --------------| -----------|
|combustable_veronoi.fs|[]|False
|spectrum_analyzer1.fs|[]|False
|audio_test.fs|[]|False
|testshader.fs|[]|False
|circuitry.fs|[]|False

(there are more shaders that don't map to pattern classes - I'm assuming it can be helpful to keep that code around for reference, but happy to make a list if we'd like to park those under `resources/shaders/unused` or something)

andrewlook:
> > This question isn't a blocker, but I'm curious about what setRemoteControls() is doing - does this affect only TouchOSC or does it affect the regular LX UI as well?
> 
> It sets the list of remote controls on the pattern, which is mainly intended for consumption by midi surfaces. However in TE we're doing an implementation that I've come to prefer on personal projects, which is to make the UI elements line up with the MidiFighterTwister knobs. For a VJ-oriented setup it makes things easier to follow if the UI matches the hardware.
> 
> > It seems that UITEPerformancePattern is listening for changes in the remoteControls, so I'm wondering if that their main purpose or if it does something else too?
> 
> It's complicated. When a pattern is loaded from a saved project file the custom remoteControls get applied After the pattern is loaded. I hope to modify the LX approach to this in the future and allow pattern classes to modify the default remote controls so the custom remote controls can be left to the user.
> 
> These are great questions. Thanks for absorbing the project in such detail.

Ok I think this makes sense, thanks.

Maybe part of my question was "what is a remote control?" and I think the answer is that it's the midi/osc handle to an LX Parameter, which also shows up as a UI control. I see the appeal of having the LX UI match up to the midifighter layout - maybe part of why I got confused here is how closely related they are and there seem to be parallel chunks of code doing similar things. Just to make sure I understand, is it true that every remote control maps to one LXParameter?

jkbelcher:
> Thanks for the heads-up re: force-push - the convention totally makes sense, especially if reviews are in progress / other people have pulled. I'm happy to squash and rebase whenever we're getting ready to merge so the list of commits on main is less hectic, but I'll hold off on that until you give the OK.

There are better git experts here than I (@theyoffa), and they are welcome to weigh in.  My take is: rebasing is good on your local machine before you add your branch to the repo and PR, that brings it up to date with main at the time PR is created.  But after that don't rebase since it changes the history.  When a PR is ready to merge, if there are no merge conflicts you can do it all in github.  If there are merge conflicts, do this on your local machine:

```
git checkout main
git pull
git checkout featureBranch
git merge main
(fix any merge conflicts)
git push
(Then the PR can be merged on github)
```

jkbelcher:
> Re: color usage - here's a list of all the shaders that don't use color palette. Seems like 6 registered patterns total.

I'm unaware of the original directives in this project.  Seems like a list in Notion is the place for these, linked from Software Tasks.

src/main/java/titanicsend/util/MissingControlsManager.java
```diff
@@ -0,0 +1,83 @@
...
+            missing_control_tags = new ArrayList<>();
+            pattern_classes = new ArrayList<>();
+        }
+    }
+
+    private final Map<String, MissingControls> classToMissingControls = new HashMap<>();
+
+    private static MissingControlsManager instance;
+
+    public static synchronized MissingControlsManager getMissingControlsManager() {
```
jkbelcher, line 37: Unwritten LX convention would simplify this method name to `get()`.  The context is already known because you're calling it on the MissingControlsManager class, so repeating the name in the method just makes things longer to read.  IE the call would be `MissingControlsManager.get()` instead of `MissingControlsManager.getMissingControlsManager()`.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -23,11 +23,14 @@
...
 import titanicsend.pattern.yoffa.framework.TEShaderView;
+import titanicsend.util.MissingControlsManager;
 import titanicsend.util.TE;
 import titanicsend.util.TEColor;
 
 import java.nio.FloatBuffer;
-import java.util.HashMap;
+import java.util.*;
+
+import static titanicsend.util.MissingControlsManager.getMissingControlsManager;
```
jkbelcher, line 33: Shouldn't need to import a method, only classes.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
jkbelcher: Patterns will all load on the engine thread, not in parallel.

jkbelcher: >  I extracted it into titanicsend.util.MissingControlsManager 

That's a good place for it.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -610,20 +634,40 @@ public TECommonControls setUnits(TEControlTag tag, LXParameter.Units units) {
          * controls, you must call addParameter() for them after calling
          * this function so the UI stays consistent across patterns.
          */
-        public void addCommonControls() {
-            registerColorControl();
+        public void addCommonControls(TEPerformancePattern pat) {
```
jkbelcher, line 637: Hmm will have to think about this one.

jkbelcher (commented):
This is a partial review

src/main/java/titanicsend/util/MissingControlsManager.java
```diff
@@ -0,0 +1,83 @@
...
+        if (instance == null) {
+            instance = new MissingControlsManager();
+        }
+        return instance;
+    }
+
+    private MissingControlsManager(){
+        try {
```
jkbelcher, line 45: Let's just fire this once early, when all the other file reading is happening, rather than triggering it from within a Pattern load.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
andrewlook: done

src/main/java/titanicsend/util/MissingControlsManager.java
andrewlook: done

src/main/java/titanicsend/pattern/TEPerformancePattern.java
andrewlook: OK, let me know what you come up with

src/main/java/titanicsend/util/MissingControlsManager.java
andrewlook: that sounds good to me - what would be a good place to do that? I'm not super familiar with the init process before patterns get loaded. Would calling it at the top of `TEApp.init()` be a good place, or somewhere else?

andrewlook:
> > Thanks for the heads-up re: force-push - the convention totally makes sense, especially if reviews are in progress / other people have pulled. I'm happy to squash and rebase whenever we're getting ready to merge so the list of commits on main is less hectic, but I'll hold off on that until you give the OK.
> 
> There are better git experts here than I (@theyoffa), and they are welcome to weigh in. My take is: rebasing is good on your local machine before you add your branch to the repo and PR, that brings it up to date with main at the time PR is created. But after that don't rebase since it changes the history. When a PR is ready to merge, if there are no merge conflicts you can do it all in github. If there are merge conflicts, do this on your local machine:
> 
> ```
> git checkout main
> git pull
> git checkout featureBranch
> git merge main
> (fix any merge conflicts)
> git push
> (Then the PR can be merged on github)
> ```

Got it, thanks. I just merged main and pushed btw, no conflicts yet.

src/main/java/titanicsend/util/MissingControlsManager.java
andrewlook: just checking in here, i think this is the last outstanding piece of feedback (pending any other reviewers). LMK if there's a more convenient place to put this and I'm happy to move it.

jkbelcher: Whew thanks for your patience!  Let's do it [here](https://github.com/titanicsend/LXStudio-TE/blob/main/src/main/java/heronarts/lx/studio/TEApp.java#L138) in the TEApp.Plugin.constructor where we initialize all the other singletons.

andrewlook: @jkbelcher No worries, you've been doing a lot of amazing work!

So I made a fix here ( https://github.com/titanicsend/LXStudio-TE/pull/332/commits/9c0a5e302ebf7df0a657f8cbe2594c958f8f04fa ) to initialize the singleton once during TEApp initialization.

If I understand correctly, this should make sure that there won't be any file-reading at pattern load time. The only downside I see is that just calling `singleton.get()` doesn't make it super clear that a file is getting loaded, but it seemed simpler than anything else I could think of - so I just added a comment explaining what's happening. LMK if you think there's a better way to go about this and I'm happy to make the change.

jkbelcher: I think this is fine.  If you were looking for a way to make it more clear you could have broken off the first call into a static initialize() method.  This will work... importantly the file access is surrounded with a try/catch so the constructor completes and the get() method will only try the file load once.

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
```diff
@@ -18,7 +18,7 @@ public FollowThatStar(LX lx) {
         super(lx, TEShaderView.ALL_POINTS);
 
         controls.setRange(TEControlTag.QUANTITY, 5, 1, 10)
-                .setUnits(TEControlTag.QUANTITY, LXParameter.Units.INTEGER);
+                .setUnits(LXParameter.Units.INTEGER);
```
jkbelcher, line 21: Per above, I'd say we stick with how these were from the pattern classes.

src/main/java/titanicsend/pattern/jon/TEControlTag.java
```diff
@@ -31,4 +31,20 @@ public String getPath() {
 
         return path;
     }
+
+    public String getLabel() {
```
jkbelcher, line 35: This looks good.  If we're adding it here though we should use this in `TECommonControls.buildDefaultControlList()` to avoid duplication of these labels.

jkbelcher (changes requested):
We're through the thick of it.  Found a few more edits.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -548,43 +552,63 @@ public TECommonControls setGetterFunction(TEControlTag tag, _CommonControlGetter
             return this;
         }
 
-        public TECommonControls setRange(TEControlTag tag, double value, double v0, double v1) {
-            // copy data from previous tag
-            LXListenableNormalizedParameter oldControl = getLXControl(tag);
+        public static LXListenableNormalizedParameter updateParam(LXListenableNormalizedParameter oldControl, String label, double value, double v0, double v1) {
```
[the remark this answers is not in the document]
andrewlook: got it! done.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -548,43 +552,63 @@ public TECommonControls setGetterFunction(TEControlTag tag, _CommonControlGetter
...
             } else {
-                TE.err("Unrecognized control type in TE Common Control " + tag);
-                return this;
+                TE.err("Unrecognized control type in TE Common Control " + oldControl.getClass().getName());
+                return oldControl;
             }
+            return newControl;
+        }
 
+        public LXListenableNormalizedParameter setRange(TEControlTag tag, double value, double v0, double v1) {
```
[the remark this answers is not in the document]
andrewlook: that makes sense, especially now that I have more context on why TECommonControls is here. Thanks for explaining the rationale.

made the change just now.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -548,43 +552,63 @@ public TECommonControls setGetterFunction(TEControlTag tag, _CommonControlGetter
...
+            LXListenableNormalizedParameter oldControl = getLXControl(tag);
+            LXListenableNormalizedParameter newControl = updateParam(oldControl, oldControl.getLabel(), value, v0, v1);
             setControl(tag, newControl);
-            return this;
+            return newControl;
+        }
+
+        public LXListenableNormalizedParameter setLabel(TEControlTag tag, String newLabel) {
```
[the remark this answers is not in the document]
andrewlook: done

src/main/java/titanicsend/pattern/TEPerformancePattern.java
andrewlook: ah yes, just needed a way to do the lookup on the parent pattern but didn't want to pass it in via the constructor just for this.

src/main/java/titanicsend/pattern/jon/FollowThatStar.java
andrewlook: ok done. made the change here and in all patterns i'd touched in this PR:
- FollowThatStar.java
- Phasers.java
- SpiralDiamonds.java
- TESparklePattern.java
- BouncingDots.java

src/main/java/titanicsend/pattern/jon/TEControlTag.java
andrewlook: got it. done.

andrewlook:
@jkbelcher thanks for the detailed review! I think I made all the relevant changes (and I tried to batch them into manageable commits since there's a lot here), but lmk if I missed anything!

jkbelcher (approved):
Approving! Thanks @andrewlook for sticking with this to get us a new system for hiding unused controls.

Extension into TouchOSC code can be added later, is dependent on the touchosc layout being updated as well.

jkbelcher:
I've run this and tested with patterns that have hidden controls and patterns that have no hidden controls.  Merging!

andrewlook:
awesome, thanks @jkbelcher !

PR #333, opened 2023-07-01
Starting a list of patterns that don't respect color palette

andrewlook:
I just wanted to start a [list of patterns that do not respect palette](https://www.notion.so/titanicsend/Catalog-list-of-patterns-that-do-not-respect-palette-59c5a538668f4b4fbffe3b02e43556a6?pvs=4), so adding comments was the easiest way to start a draft PR while I work.

Next steps:
- I'll take a second pass through all the patterns and see if I missed any patterns
- For each of the patterns in question, I'll see if there's a clean way to swap in our palette (might ask for feedback as I go)

### DIFF
--- a/resources/pixelblaze/fireworknova.js
+++ b/resources/pixelblaze/fireworknova.js
@@ -18,6 +18,7 @@ export function render3D(index, x, y, z) {
   z -= 0.5
   //get pixel distance from center
   r = sqrt(x*x + y*y + z*z) * scale
+  // TODO: respect color palette
   //make colors
   h = r * 3
   //blast wave - a triangle's peak moving based on the center
--- a/resources/shaders/jet_stream.fs
+++ b/resources/shaders/jet_stream.fs
@@ -362,6 +362,7 @@ void mainImage( out vec4 fragColor, in vec2 fragCoord )
         arclight += exp(abs(arcz - tunRef.p.z) * -0.3) * fract(sin(arcseed) * 198721.6231) * arcint;
     }
     vec3 arccol = vec3(0.9, 0.7, 0.7);
+    // TODO: respect color palette
     col += arclight * arccol * 0.5;
     col = mix(col, arccol, clamp(arcv, 0.0, 1.0));
     col = pow(col, vec3(1.0, 0.8, 0.5) * 1.5) * 1.5;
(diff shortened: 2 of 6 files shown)


theyoffa:
Similar to the other one, if we can dynamically create that list, should be super easy to extend for detecting `iColorRGB`. We could plug it into updating the `palette` param, since that also won't work in this case and we could similarly autogen. This would future proof and keep in sync.

zranger1:
I've fixed two of the weirder ones, fire.fs and neontriangles.fs, in PR #341.

(We did a last-second EDC change to neon triangles to use iColorRGB for its main triangle color. But it also had a complicated 3D dithering pass for generating bloom, with its own color generator, that we just didn't get around to fixing on the moment.  The moral is, I guess, watch out for glow and other secondary color effects when testing these!)

worldveil:
+1 to autogen of some sort that is future proof as @theyoffa mentioned

andrewlook:
> Similar to the other one, if we can dynamically create that list, should be super easy to extend for detecting `iColorRGB`. We could plug it into updating the `palette` param, since that also won't work in this case and we could similarly autogen. This would future proof and keep in sync.

Autogenerating the list generation makes sense to me.

What do you mean by "plug it into updating the `palette` param, since that also won't work in this case"?

theyoffa:
so there's a palette common parameter that switches which colors are used. not sure if you were thinking about that or the main palette, but if one doesn't work, that other won't either. since that's not working in this case, we can update the palette param with an [x] just like your other PR.

andrewlook:
Quick update: I'll avoid touching the control labels till I have a better sense for how we want to signal touchOSC

I did make a list of the (shader-only) patterns that don't use at least one of `iColorRGB,iColorHSB,iColor2RGB,iColor2HSB` followed by the Shader classes that reference them (note there's a few fragment shaders with no Java patterns referencing them)

```
testshader.fs	[]
circuitry.fs	[]
neon_heart.fs	['NeonHeartNative']
audio_test.fs	[]
pulsing_heart.fs	['PulsingHeart']
combustable_veronoi.fs	[]
warp.fs	['Warp']
spectrum_analyzer1.fs	[]
synth_waves.fs	['SynthWaves']
jet_stream.fs	['JetStream']
neon_cells.fs	['NeonCells']
```

andrewlook:
closing in favor of #332

PR #334, opened 2023-07-04
new util pattern for visually targeting specific pixels

theyoffa:
[Notion task](https://www.notion.so/titanicsend/f5264e3538bf417e86956d87124d1a89?v=e45c0f187eb84ecabbae9c00a27f0fd2&p=aca54d6af197448e95bf3ebbedd40a3b&pm=s)

Movable bullseye with a single pixel in the middle. Can stamp multiple places in the car at once. Still doesn't wrap around the car appropriately (similar to hand tracker) but the rest of the functionality is ready to go. If you see the middle pixel disappearing, it will be fixed with proper wrapping (it appears on the other side of the car). This will be fixed with [HandTracker Improvements](https://www.notion.so/titanicsend/f5264e3538bf417e86956d87124d1a89?v=e45c0f187eb84ecabbae9c00a27f0fd2&p=5d2923a030d448948fc568c28930f239&pm=s)

Middle pixel doesn't cycle colors, lmk if you think we need this.

### DIFF
--- a/src/main/java/titanicsend/app/TEApp.java
+++ b/src/main/java/titanicsend/app/TEApp.java
@@ -72,6 +72,7 @@ import titanicsend.pattern.justin.TESolidPattern;
 import titanicsend.pattern.mike.*;
 import titanicsend.pattern.pixelblaze.*;
 import titanicsend.pattern.tom.*;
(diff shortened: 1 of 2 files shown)


raldi (approved):
I can’t wait to play with this — let’s merge it quick!

PR #335, opened 2023-07-04
utility effect to make minor adjustments panel locations at runtime

theyoffa:
[Notion task](https://www.notion.so/titanicsend/f5264e3538bf417e86956d87124d1a89?v=e45c0f187eb84ecabbae9c00a27f0fd2&p=4b2129404fa84ff7b835f9a7a7495af3&pm=s)

fully featured with runtime UI and save functionality.

note if we're good on this approach, we'll have to add it to the project files similar to the NoGapEffect.


raldi (approved):
It’ll be hard to test this until the end of Build Week, but I’m feeling optimistic.

src/main/java/titanicsend/effect/PanelAdjustEffect.java
```diff
@@ -0,0 +1,183 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.lang.reflect.Type;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+
+@LXCategory("Utility Patterns")
```
jkbelcher, line 29: This shows up in the Effects list, so it might be more appropriate to name the category just "Utility"

src/main/java/titanicsend/effect/PanelAdjustEffect.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    @Override
+    public void onParameterChanged(LXParameter parameter) {
+        if (parameter == this.panelIdParameter) {
+            String panelId = this.panelIdParameter.getString();
+            currentAdjustment = PANEL_ID_TO_ADJUSTMENT.get(panelId);
+            if (currentAdjustment == null) {
+                currentAdjustment = new TEPanelModel.Adjustment();
+                if (isValidPanelId(panelId)) {
```
jkbelcher, line 92: This could be checked immediately after the `String panelId = ...` line.  No need to create a `TEPanelModel.Adjustment` for an invalid panel.

src/main/java/titanicsend/effect/PanelAdjustEffect.java
theyoffa: sure can drop `patterns` in a future pr

src/main/java/titanicsend/effect/PanelAdjustEffect.java
theyoffa: practically speaking, there's no real difference

PR #337, opened 2023-07-04
Update label of NoGapEffect toggle

raldi:
This was getting truncated to just "Show G" in the UI

### DIFF
--- a/src/main/java/titanicsend/effect/NoGapEffect.java
+++ b/src/main/java/titanicsend/effect/NoGapEffect.java
@@ -15,7 +15,7 @@ public class NoGapEffect extends TEEffect {
 
   public NoGapEffect(LX lx) {
     super(lx);
-    addParameter("Show", showGapPixels);
+    addParameter("Show Gaps", showGapPixels);
   }
   
   @Override
@@ -23,7 +23,7 @@ public class NoGapEffect extends TEEffect {
     // Friendly reminder this will be most effective on master channel
     LXComponent parent = this.getParent();
     if (parent == null || !parent.equals(this.lx.engine.mixer.masterBus)) {
-     LX.log("Warning: NoGap Effect is running on a channel that is not master. Gap pixels may not get filtered.");
+     LX.warning("NoGap Effect is running on a channel that is not master. Gap pixels may not get filtered.");
     }
     super.onEnable();
   }
--- a/src/main/java/titanicsend/effect/NoGapEffect.java
+++ b/src/main/java/titanicsend/effect/NoGapEffect.java
@@ -15,7 +15,7 @@ public class NoGapEffect extends TEEffect {
 
   public NoGapEffect(LX lx) {
     super(lx);
-    addParameter("Show Gaps", showGapPixels);
+    addParameter("Show", showGapPixels);
   }
   
   @Override


src/main/java/titanicsend/effect/NoGapEffect.java
```diff
@@ -15,15 +15,15 @@ public class NoGapEffect extends TEEffect {
...
+    addParameter("Show", showGapPixels);
   }
   
   @Override
   protected void onEnable() {
     // Friendly reminder this will be most effective on master channel
     LXComponent parent = this.getParent();
     if (parent == null || !parent.equals(this.lx.engine.mixer.masterBus)) {
-     LX.log("Warning: NoGap Effect is running on a channel that is not master. Gap pixels may not get filtered.");
+     LX.warning("NoGap Effect is running on a channel that is not master. Gap pixels may not get filtered.");
```
jkbelcher, line 26: As-is this doesn't print to console but you could use "TE.err(" which prints to std error (red text in eclipse) but doesn't actually throw an error.

raldi: We're configured to print `log` but not `warning`? I thought warning was higher priority than general logs. We have some important-looking LX.warning() logs throughout the source, so maybe we should turn this loglevel on?

theyoffa: the idea of error log is to log issues that should be addressed. you don't want to pollute it with non-issues or you will obscure the real problems.

raldi: Sure, but the log levels, in decreasing priority, are: error, warning, log. We're already logging error (in red) and warning (in gray), so it makes sense to also log warnings (in gray, with "<WARNING>" prefixed).

theyoffa: gap pixels showing might arguably qualify, but things like [this](https://github.com/titanicsend/LXStudio-TE/pull/335/files#diff-0656472eb7a4c5cdf49517710e975c11cc9b36e5b0a59763eb29c2cb0e4f53daR62) would not.

raldi: What would you recommend be done when the adjustments file cannot be saved? Popup, silence, LX.log(), warning, or error?

theyoffa: we are agreeing :) i was just saying we should show warnings rather than dump everything into error

theyoffa (approved):

src/main/java/titanicsend/effect/NoGapEffect.java
theyoffa: if user facing functionality is not working as intended, i would definitely log an error. if possible, also letting the user know some way or another would be ideal.

theyoffa: if we're talking about the PR referenced above, in that case i am throwing a runtime exception which will hit error logs. if i was less lazy, it would include a descriptive message and warning to the user.

jkbelcher: Turning on warnings logging is the right solution.  Good work everybody.

PR #338, opened 2023-07-04
Animate the bullseye through a cycle of colors and make the active target a different color than the saved ones

raldi:


src/main/java/titanicsend/pattern/util/TargetPixelStamper.java
```diff
@@ -45,21 +56,36 @@ public TargetPixelStamper(LX lx) {
...
         addParameter("Stamp", this.stamp);
         addParameter("Clear", this.clear);
+        this.totalMsec = 0.0;
     }
 
     @Override
     public void run(double deltaMs) {
-        showSavedStamps();
-        showCurrentTarget();
+        this.totalMsec += deltaMs;
```
theyoffa, line 64: i like storing a timestamp so you don't have to worry about maintaining the correct time, but this also works fine.

src/main/java/titanicsend/pattern/util/TargetPixelStamper.java
```diff
@@ -45,21 +56,36 @@ public TargetPixelStamper(LX lx) {
...
+        int phase = (int)(this.totalMsec / MSEC_PER_COLOR) % 4;
+        int bullseyeColor;
+        if (phase == 0) {
```
theyoffa, line 67: can use a switch or even just put these in an array and retrieve the index of phase

theyoffa (approved):

src/main/java/titanicsend/pattern/util/TargetPixelStamper.java
raldi: Done

PR #343, opened 2023-07-09
Allow multiple ggg for gap pixels again, allow framerate to be specified as an arg, fix bug in virtual fixture involving white pixels

raldi:


jsohi11 (approved):
Discussed updates in Slack

PR #344, opened 2023-07-12
Added new script to monitor SK memory usage

worldveil:
WIP for a script to monitor SK and kill and restart if goes too high.

### DIFF
--- a/script/beat_linking/monitor_sk.py
+++ b/script/beat_linking/monitor_sk.py
@@ -1,0 +1,51 @@
+import psutil
+import time
+
+
+POLLING_WAIT_TIME_SECS = 5
+SK_PROCESS_NAME = "xxx"
+MEMORY_GB_LIMIT = 15
+
+def is_process_sk(process):
+    """
+    Returns True if this is the SK process, False otherwise
+    """
+    return "rekordbox" in p.name().lower()
+
+def is_process_taking_too_much_memory(process):
+    """
+    Uses psutil's `process.memory_info()` to check for memory usage
+    https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info
+
+    I have found this number to be a little low. For example, I was seeing
+    a process report as 1.81 GB in ActivityMonitor, but report 780 GB using psutil,
+    so we may want to have a fudge factor or add another form of memory to it. If
+    we can't get the field we may need to run as "sudo" to get it.
+    """
+    try:
+        gb_using = p.memory_info().rss / (1024.0 ** 3)
+        return gb_using > MEMORY_GB_LIMIT
+    except psutil.AccessDenied:
+        return False
+
+def kill_process(process):
+    # TODO: need to kill a process given a PID
+    pass
+
+def restart_sk():
+    # TODO: need to run the application for SK to boot it up
+    # need this to be a separate process that isn't dependent on
+    # this script, ideally
+    pass
+
+while True:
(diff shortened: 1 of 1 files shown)


script/beat_linking/monitor_sk.py
```diff
@@ -0,0 +1,51 @@
+import psutil
+import time
+
+
+POLLING_WAIT_TIME_SECS = 5
+SK_PROCESS_NAME = "xxx"
```
raldi, line 6: Do you want to do `SK_PROCESS_NAME = sys.argv(1)`?

raldi: Or is this a vestigial constant?

script/beat_linking/monitor_sk.py
```diff
@@ -0,0 +1,51 @@
...
+    we can't get the field we may need to run as "sudo" to get it.
+    """
+    try:
+        gb_using = p.memory_info().rss / (1024.0 ** 3)
+        return gb_using > MEMORY_GB_LIMIT
+    except psutil.AccessDenied:
+        return False
+
+def kill_process(process):
+    # TODO: need to kill a process given a PID
```
raldi, line 32: Three ways:

```
os.system(f"kill {pid}")

os.system(f"pkill {name}")

def kill_process_by_name(name):
    for proc in psutil.process_iter(['pid', 'name']):
        if proc.info['name'] == name:
            psutil.Process(proc.info['pid']).terminate()            
```

script/beat_linking/monitor_sk.py
```diff
@@ -0,0 +1,51 @@
...
+
+while True:
+    pids = psutil.pids()
+    for i, pid in enumerate(pids):
+        p = psutil.Process(pid)
+        if is_process_sk(p):
+            should_kill = is_process_taking_too_much_memory(p)
+            if should_kill:
+                kill_process(process)
+                restart_sk()
```
raldi, line 49: Do you want to place a limit on the script where it'll only do a restart at most once every N seconds?

titanicsend: Good idea

script/beat_linking/monitor_sk.py
```diff
@@ -0,0 +1,65 @@
...
+def is_process_taking_too_much_memory(process):
+    """
+    Uses psutil's `process.memory_info()` to check for memory usage
+    https://psutil.readthedocs.io/en/latest/#psutil.Process.memory_info
+
+    I have found this number to be a little low. For example, I was seeing
+    a process report as 1.81 GB in ActivityMonitor, but report 780 GB using psutil,
+    so we may want to have a fudge factor or add another form of memory to it. If
+    we can't get the field we may need to run as "sudo" to get it.
+    """
```
andrewlook, line 29: @worldveil in case this is helpful, I looked around a bit after you mentioned this last night during the software meeting.

Virtual vs actual memory usage numbers can vary across processes - looking at total "free" memory might be a simpler way of foreseeing a crash.

here's a pointer to some python code that shells out to `ps`/`vm_stat`: https://apple.stackexchange.com/a/4296

There's also the pip library `psutil` and the script [meminfo.py](https://raw.githubusercontent.com/giampaolo/psutil/master/scripts/meminfo.py), which I came across here: https://apple.stackexchange.com/questions/4286/is-there-a-mac-os-x-terminal-version-of-the-free-command-in-linux-systems/4289#4289

PR #345, opened 2023-07-12
Blending Fixes

zranger1:
### Enhanced Alpha Blending

Allows smooth, consistent blending between channels, regardless of channel order.  The main part of the work
is in ```shaders/framework/template.fs```, with a handful of adjustments to individual patterns.  There's a similar
fix for Java FragmentShader patterns like MatrixScroller and NeonSnake.   Most Java patterns do not set alpha directly and so will not be affected by this PR.

See comments in ```shaders/framework/template.fs``` for detailed notes on how it works, and the various options for
turning it on/off and setting a threshold for opacity in individual shaders if necessary. 

Note that in case of emergency, you can disable the new method entirely and return to the blending we had at EDC by using
```#define TE_NOALPHAFIX``` at the top of ```shaders/framework/template.fs```, and restarting LX.

### Testing
The best way to see what's going on is by playing with faders on 3 different channels.  In the default "Add" blend mode,
channel order should make little difference in the output brightness. In general, dimmer pixels should be more transparent,
brighter pixels less so.

### Also includes a couple of pattern fixes:
- Neon Triangles was looking a little dull after getting palette compliance.  It now has its **NEON** back.  Glow is
adjustable with the Wow1 control, foreground color/gradient mix w/Wow2.
- RainBands didn't get merged in time for last week's speed fix.  Its speed now goes in the proper direction.

### CODE REVIEW

File: src/main/java/titanicsend/pattern/will/shaders/RhythmicFlashingStatic.java [hunk 1]
```diff
@@ -22,7 +22,7 @@ public class RhythmicFlashingStatic extends FragmentShaderEffect {
 
     public RhythmicFlashingStatic(PatternTarget target) {
         super(target);
-        outColor = new double[4];
+        outColor = new double[3];
```
jkbelcher, line 25: This is causing line 85 to throw a fatal exception:

image

jkbelcher:
I ran this with 3 channels of patterns including the modified ones.  Confirmed blending looks great and channel order does not impact the blend.  I merged main into this branch to make sure it works with latest.  Also restored the Wow1 dial but am not sure if editing the `missingControls.json` was the correct way to do that - @andrewlook can you confirm?

There is one error in RhythmicFlashingStatic, otherwise this can be approved for merge.

src/main/java/titanicsend/pattern/will/shaders/RhythmicFlashingStatic.java
zranger1: Weirdest thing ever!   I think you were the first person to try this pattern since before we added views.   To get it working, I had to:
- fix the obvious array size problem
- un-normalize the incoming coordinates, b/c it expected a much larger range
- fix a threading/memory access problem.  (This pattern type is from before my time - it's TE's first attempt to do anything shader-like.   It uses a bunch of wrapper functions to do shadery-y things per-pixel using a thread pool.  Don't think I'd looked closely enough before to see that it was multi-threaded.  Anyway, changed some temporary object allocations to be thread-safe.

zranger1:
I also restored the WowTrigger control for Edgefall by editing the JSON file.   Looks like it runs OK.

For patterns with code both in Java and GLSL, we need to back up the automated script with a human sanity check.  I would avoid re-running the script while on site at events, where time and pressure make this difficult.   A few non-working controls are easier to deal with than missing important ones.

src/main/java/titanicsend/pattern/will/shaders/RhythmicFlashingStatic.java
jkbelcher: @zranger1 I still get a crash running this pattern.  Does it run for you?  The `outColor` array is initialized to length 3 and then line 85 tries to access `outcolor[3]`

resources/shaders/framework/template.fs
```diff
@@ -51,6 +84,14 @@ void main() {
...
-    // force black pixels to full transparency, otherwise use shader provided alpha
+    // Post-processing: Make sure we've got optimal color and alpha values for brightness
+    // and blending. define TE_NOALPHAFIX in your shader code if you need
+    // to disable this feature.
+    #ifndef TE_NOALPHAFIX
+    finalColor = _blendFix(finalColor);
+    #else
+    // Old EDC blending - force black pixels to full transparency, otherwise
+    //use shader provided alpha
     finalColor.a = ((finalColor.r + finalColor.g + finalColor.b) == 0.0) ? 0.0 : finalColor.a;
```
andrewlook, line 95: not sure if this is relevant, but: what if r+g+b=0.001 - should we still make it mostly transparent? putting it another way, is there some threshold below which we'd want to still set alpha to a low value, even if it's not absolute black?

andrewlook, line 95: not sure if this is relevant, but: what if r+g+b=0.001 - should we still make it mostly transparent? putting it another way, is there some threshold below which we'd want to still set alpha to a low value, even if it's not absolute black?

src/main/java/titanicsend/pattern/will/shaders/RhythmicFlashingStatic.java
andrewlook: @zranger1 I also still get a crash running this pattern

zranger1: Something went badly wrong w/git.   I think the changes to that pattern got partly reverted somehow.  Fixing... (I hope)

resources/shaders/framework/template.fs
zranger1: At this point, we were arbitrarily declaring 0 transparent and everything else fully opaque.  It was done like, minutes before a show, and we just wanted to make sure everything would show up as bright as possible!    It worked, but it made blending kind of clunky, which is why we updated it here.

zranger1: see above... probably a duplicate.   Hopefully, we'll never have to use this blending mode again, but it's there for emergencies.

zranger1:
Had a git meltdown in there somewhere.   Correct version of RhythmicFlashingStatic should be there now.

resources/shaders/framework/template.fs
andrewlook: ah that makes sense, thanks for explaining! After thinking about it some more I can see how what you did above would have the same effect as what I was wondering about.

andrewlook:
> I also restored the WowTrigger control for Edgefall by editing the JSON file. Looks like it runs OK.
> 
> For patterns with code both in Java and GLSL, we need to back up the automated script with a human sanity check. I would avoid re-running the script while on site at events, where time and pressure make this difficult. A few non-working controls are easier to deal with than missing important ones.

Agreed. I'll hold off on running the missing controls script at all unless we need to - for now, just adding the name of the uniform in a comment seems like a functional workaround until we can find a better solution after the burn.

src/main/java/titanicsend/pattern/will/shaders/RhythmicFlashingStatic.java
andrewlook: just re-tested and the fix looks good 👍

andrewlook (approved):
Tested before/after the blending fixes made and can confirm that color blending looks better, and many patterns seem brighter than they did before the changes in this branch.

Also double-checked that `RhythmicFlashingStatic` was fixed based on the latest changes.

PR #347, opened 2023-07-15
chromatik

andrewlook:
add exec:exec goal ExecMain, with -Djava.awt.headless integrated.

```
mvn clean package
mvn exec:exec@ExecMain
```


jkbelcher (approved):

raldi:
> mvn exec:exec@ExecMain

Is this what we need to do every time now? Is there a way to teach IntelliJ to do it automatically when we click the ▶️ button?

PR #352, opened 2023-07-19
fix shaders on the sides of the car

theyoffa:
well, here's an inelegant way of doing it. i mean it works.

that said, now that we're using view instead of models, i have a hard time finding metadata in the view objects. I think i have the idea on how they are created in used, but with love to chat with @jkbelcher at some point to better understand how to handle them.

for this case, is there an easy way of understanding which model a point comes from?

### DIFF
--- a/src/main/java/titanicsend/model/TEWholeModel.java
+++ b/src/main/java/titanicsend/model/TEWholeModel.java
@@ -15,6 +15,7 @@ import heronarts.lx.transform.LXVector;
 import titanicsend.lasercontrol.MovingTarget;
 import titanicsend.output.ChromatechSocket;
 import titanicsend.output.GrandShlomoStation;
+import titanicsend.pattern.TEPattern;
 import titanicsend.util.TE;
 
 public class TEWholeModel extends LXModel {
@@ -26,6 +27,7 @@ public class TEWholeModel extends LXModel {
   public HashMap<LXVector, List<TEEdgeModel>> edgesBySymmetryGroup;
   public HashMap<String, TEPanelModel> panelsById;
   private final HashMap<TEPanelSection, Set<TEPanelModel>> panelsBySection;
+  private final HashMap<TEPanelSection, Set<Integer>> pointIndexesBySection;
   public HashMap<String, List<TEPanelModel>> panelsByFlavor;
   public HashMap<String, TELaserModel> lasersById;
   public List<LXPoint> edgePoints; // Points belonging to edges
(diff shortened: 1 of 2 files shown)


zranger1:
Let's talk about this at tonight's meeting if possible.   There's an experiment I've been meaning to do at the point right before views are created.  Here's the outline:

- make a temporary copy of the model's normalized points
- at that point, it's easier to get just edges and panels on the ends of the car.
- for end pixels, we leave nx and ny alone, but add the magnitude of nx to nz, in the direction that nz is going.  
- this creates a distorted geometry which if rendered, will give the car a weirdly pointy nose and tail.
-  but if it works, Chromatik will use the "real" x,y,z values to draw the car in the UI, and the distorted geometry will let us texture map in the simple way we've already been doing it.
- doing it this way won't disturb x, and so will have minimal impact on the few java patterns that actually make use of all 3 dimensions.  
- the only trick is to make sure that aspect ratios in views don't get too distorted by the extra added z.

theyoffa:
yeah, let's chat at the meeting. one thing to note is this is only for native shaders so java patterns should be unaffected.

if i understand you correctly, it think the result might be similar to the other option i was considering, which would be to use a projection to rotate the side panels by 90 degrees (virtually, not in UI) and have them painted as an extension of the canvas. however i think you miss out on some of the cooler parts of the pattern this way.

zranger1:
Rotating them 90 degrees works too.   What I'm proposing winds up with a little less rotation.   The effect is almost identical, it just distorts aspect ratio a bit less, even if we don't do anything else.   If we're only doing it for shader-style effects, it might even be possible to leave the existing normalized points alone and add (u,v) normalized texture coordinates to the model, spccifically for this purpose.

zranger1:
My other wild idea will have to wait 'till after burning man, but it would make the shader system fully 3D once and for all.

What we do is this:
Instead of rendering to a quad, we pass the normalized 3D coords of the current view, plus the point's index in the TE color buffer, to the vertex shader as point geometry.  (The point geometry can be made persistent in GPU memory until the view changes, so we won't lose time loading it on every frame. )

It would then call the fragment shader with the full 3D coordinates for every point in the view. 

When it got a color back, instead of painting to a background image and reverse mapping onto the car, we'd have  a buffer full of colors exactly the right size, in the right order, ready to copy to the "real" color buffer.

Another advantage to doing something like this is that it eliminates texture sampling/rescaling issues, and might even result in faster overall rendering, since we'd be rendering only points that actually exist on the car.

PR #353, opened 2023-07-19
beat reactive explosion for all shaders

theyoffa:
take the effect from matrix scroller and make it a common param for all native shaders

i loved this effect from edc and thought it would be cool to have it on more patterns. that said, i am conscious of diluting our base parameter set. can discuss at meeting, i'm open to making it into an effect rather than a default param.


jkbelcher:
I merged main into this branch for the essence of time, hope that's ok.  Then a quick commit to fix a compile error where a new switch statement did not include your enum value.

jkbelcher:
It would be Total Chaos to insert a new control in the middle of the list and move the Wow controls to the right.  Went ahead and fixed this before considering further.

jkbelcher (commented):
I agree this effect is cool.  Comments:

- The last parameter space had been reserved for a Shift parameter that never got built and likely won't happen before BM.  In which case I think it's fine to fill it for now.
- Overall this feels more like an effect than a pattern parameter.  For the immediate event however running an effect wouldn't put it on a knob, so maybe this is ok for now.  Would you be open to also making this as an effect for the sake of longevity/portability?  Not blocking the merge for it, just a nice to have.
- The effect appears to move the pattern slightly down and to the right as it comes on.  You can see it well on NeonTriangles. @theyoffa do you think you could modify that?

zranger1:
Just pushed fix for directional bias + performance and axis orientation fixes.  Should be good to go now.   This is dead cool!

Note that it only works on GLSL shader patterns, so VJs will still have to be a bit alert.  To make it universal, we really will have to do it as an effect.

jkbelcher:
Good point.  Can we hide this knob on all the java patterns?

zranger1:
Pretty sure we can talk @andrewlook into modifying the missing controls script to disable TEControlTag.EXPLODE for everything that doesn't have an associated shader.  That'd be the simplest fix.

jkbelcher:
Great.  Let's do that as a separate PR to keep the queue moving.  Merging this in.

jkbelcher (approved):

PR #354, opened 2023-07-19
Hide unused controls from UI without impacting midi controllers

jkbelcher:
Here's a clean way to help the VJ see which controls are active for the current pattern.  By hiding the unused controls you wouldn't need to change their labels.


jkbelcher:
@andrewlook Closing, looks like you implemented these changes manually.

PR #355, opened 2023-07-20
ModelBender - Improved Texture Mapping 

zranger1:
This PR improves texture mapping on the fore and aft ends of the car.   It does this by adjusting the z coordinates of the model before views are created.   This lets textures wrap (mostly) undistorted around the sides and ends of the car, with mirroring around a vertical plane at x == 0.

For details, see comments in jon/ModelBender.java.  
Edit: Demo/comparison video is here:  https://www.youtube.com/watch?v=Dy8-ausP1xo 

This method changes the coordinates presented to patterns, but is done in such a way that it should not interfere with true 3D patterns -- distances precomputed by LX are not altered, and the relationship between z and x is preserved.

If you do find something weird though, please let me know asap!


src/main/java/titanicsend/pattern/jon/ModelBender.java
```diff
@@ -0,0 +1,77 @@
...
+
+        // adjust the z coordinates of the end points to taper them
+        // with decreasing x. This gives the model a slightly pointy
+        // nose and makes texture mapping easier and better looking.
+        ArrayList<LXPoint> endPoints = getEndPoints(model);
+        for (LXPoint p : endPoints) {
+            // kick out gap points or they'll break normalization.
+            if (model.isGapPoint(p)) continue;
+
+            double zOffset = endXMax - Math.abs(p.x);
```
andrewlook, line 60: so if I'm understanding correctly, when we hit the side of the car, actual `z` remains fixed, but we use the `x` value to effectively extend the `z` value that we pass to the pattern, so that the texture continues onto the side of the car (rather than have each horizontal row of pixels on the side of the car always having the exact same color / resulting in horizontal striping).

is that right?

src/main/java/titanicsend/model/justin/ViewCentral.java
```diff
@@ -127,16 +129,24 @@ protected void initialize() {
...
         v.setModel(LXView.create(this.lx.getModel(), v.viewSelector, v.viewNormalization));
+
       } else {
         LX.warning("View Definition is not a filter.");
         v.setModel(this.lx.getModel());
       }
     }
+
+    // restore original model geometry
+    mb.restoreModel((TEWholeModel) this.lx.getModel());
```
andrewlook, line 149: Just asking for my own understanding: so we modify the model in-place and then revert it after view definition - what would happen to the current channel if we didn't restore original geometry?

I'm assuming we modify the model in-place (rather than creating a copy) to save on memory. But I'm unclear on where in the render lifecycle this fits - are we restoring the original coordinates in case another java pattern is using the model downstream of this one, for instance?

andrewlook (approved):
Tested this locally and the result is a big improvement - sides of the car are looking much better across all the patterns I tested 🙌 

I asked a couple of clarifying questions in the comments, but apart from that this one looks good to me.

src/main/java/titanicsend/model/justin/ViewCentral.java
zranger1: This video shows exactly what it does to the model -- kind of stretches out the nose and tail a little bit.

We actually never touch the underlying geometry at all - just slightly adjust the normalized, per-view coordinates.  The effect is rarely visible on java patterns, and then only if you know what to look for.   (And it actually looks better on Java patterns that are doing calculations in 2D, which is most of them.)

https://www.youtube.com/watch?v=WKGQpwakJT4

src/main/java/titanicsend/pattern/jon/ModelBender.java
zranger1: Yeah, you've got the right idea!   We "cheat" on the ends of the car by adding delta x to the z coord, so that z moves along with x at a roughly pixel-density-correct rate.   This way, z/y texture mapping works as it should across the ends without us having to explicitly calculate or store extra texture coordinates.   The visual result is to give the model, if rendered with these "adjusted" coordinates,  a slighty tapered pointy nose, shown in the video linked below.

andrewlook: excellent! thanks for explaining

src/main/java/titanicsend/model/justin/ViewCentral.java
andrewlook: thanks for the clarification that this only affects how views are generated at startup time 👍

jkbelcher:
This was a clever way to modify the normalized coordinates only during view creation.  Looks good.

PR #356, opened 2023-07-22
Finish PDB

jsohi11:
~See Slack for next steps for PFC: https://titanicsend.slack.com/archives/C02L0MDQB2M/p1689173793633949~

PDB has been completed (partially coded previously)


resources/vehicle/striping-instructions.txt
```diff
@@ -62,7 +62,7 @@ PED
 PEE
 PFA
 PFB 42 R . . . . +. .- . . . .- +. . . . g .- . . +. g .- . . . g +.- . . . . .- +. g . . . g .- . +. . g .- . . .
-PFC
+PFC 30 R . +. . . -------------------. . +. . +. . +. . +. . +. g ++++++++++++++++++. . +. . . +. . . +. . . +. . . +. . . +. . . . +. . +. . . . +. .
```
raldi, line 65: See Slack; I'm afraid this will have to be redone after the row with all the pluses. But that should be mostly popping work, because I believe the only change you'll need to make here is to remove the "." row after the one with all the pluses.

raldi (changes requested):

resources/vehicle/striping-instructions.txt
kryptonat: No longer relevant in this PR. PFC was fixed in another PR https://github.com/titanicsend/LXStudio-TE/commit/1f5f041835a89e8ff20dbf670865f6a3aff021f4

kryptonat (approved):

raldi (approved):

PR #357, opened 2023-07-25
Midifighter64 Refactor

zranger1:
### Midifighter64 subsystem improvements for 2023
- simplified code paths.  
- greatly reduced memory footprint
- improved performance
- "full polyphony" - multiple patterns blend, more patterns support multiple colors... Basically, you can press as many buttons simultaneously as you like and *something* interesting and different will happen.
- Automatic palette-derived color.  Overlay colors automatically change to harmonize with system palette.  (Original bright primaries are still available as an option.)
- new "Hearts" pattern. (All buttons on the first page now draw overlays. Second page buttons still run the logger.)

### To Test
- create and enable a channel, and add TEMidiFighter64DriverPattern to it.
- set the MF64 channel's blending mode to "normal"  (it defaults to "add", which doesn't work as well for this purpose)
- enable MIDI on the channel and in the system "OSC+MIDI" settings.
- press buttons on the MF64 in pretty much any combination

**Notes**  
- If palette derived color is enabled, the color on the first (leftmost) button will always be a very bright version of the palette primary and the last button will always be white.  The remaining colors will be automatically generated. 
- When the system palette changes, the MF64 palette will not be recalculated until all buttons are released (to avoid obvious visual discontinuity.)
- You actually can get a board and press all 64 buttons simultaneously.  Not the greatest look, but it will work.


jkbelcher (approved):
Ran without midi hardware, confirmed doesn't crash with no hardware.

@zranger1 I added a commit that sets the channel blend and enables midiFilter when pattern becomes active, then restores the state to whatever it was when pattern becomes inactive.  It's hard to remember special settings in the chaos of playa so I thought it'd be nice to not worry about remembering these.  If it looks good to you feel free to merge.

PR #360, opened 2023-08-02
Audio-reactive pattern: SigmoidDance

andrewlook:
https://github.com/titanicsend/LXStudio-TE/assets/236018/83fee741-2160-4102-894d-cd1f4cd603e3

2 patterns included here:
1. `SigmoidDanceAudioLevels` - average (and normalize/scale) FFT bands to distill one param (or do lower half of FFT bands as 1, upper half of FFT bands as another param). This effect is modulated by `iWow1`. Use these values for `a` and `w` params controlling sigmoid shape (described below).
  - `iQuantity` controls the number of overlapping sigmoid shapes - 1, 2, or 4.
3. `SigmoidDanceAudioWaveform` - use raw wave data proportional to X coordinate - this breaks the S shape, but the audio wave is really visible for small variations in the audio (the scale of this effect is modulated by `iWow2`).

The basic idea for this is based on a shape made of 2 overlapping logistic sigmoid functions, parametrized by `w` (width offset between the sigmoids) and `a` (steepness of the sigmoid curve).

This has some interesting properties:
- when `a` is really low (near `-1.0`), the shape looks like an `X`
- when `a` is higher (near `1`), the shape looks like two parallel vertical lines `| |`
- when `a` is moderate (`-0.7` - `0.7`), it's made of symmetrical S shapes that dance with small variations.

Screen Cast 2023-08-01 at 8 53 43 PM


src/main/java/titanicsend/pattern/look/SigmoidDanceAveraged.java
```diff
@@ -0,0 +1,98 @@
...
+        float scaledLo = normLo * 2 - 1;
+        float scaledHi = normHi * 2 - 1;
+//        System.out.printf("scaledLo = %s, scaledHi = %s\n", scaledLo, scaledHi);
+
+//        shader.setUniform("iWow1", avgLo);
+//        shader.setUniform("iWow2", avgHi);
+//        shader.setUniform("iWow1", normLo);
+//        shader.setUniform("iWow2", normHi);
+        shader.setUniform("iWow1", scaledLo);
+        shader.setUniform("iWow2", scaledHi);
```
zranger1, line 83: You don't have to re-use the control uniforms here.  In fact, I wouldn't recommend it - you might want to use those controls later.   You can create as many uniforms as you need, of whatever type, as long as the name and type are the same in both Java and GLSL.

For example, consider using ```shader.setUniform("scaledLo",scaledLo);``` in the java, paired with declaring ```uniform float scaledLo``` at the top of the shader code.

resources/shaders/sigmoid_dance.fs
```diff
@@ -0,0 +1,107 @@
...
+vec2 rotate(vec2 st, float a) {
+    st = mat2(cos(a), -sin(a), sin(a), cos(a)) * (st-.5);
+    return st + .5;
+}
+
+void mainImage(out vec4 fragColor, in vec2 fragCoord) {
+    vec2 st = fragCoord.xy/iResolution.xy;
+
+    st = rotate(st, iSpin * iTime);
+
```
zranger1, line 64: use ```rotate(st,iRotationAngle)``` here instead.   iSpin * iTime will produce discontinuous visuals if either of these values changes by more than a hair. 

The iRotationAngle thing was set up to get around this by tracking rotation and changing it smoothly and incrementally with control changes.

src/main/java/titanicsend/pattern/look/SigmoidDanceAveraged.java
andrewlook: Great advice, thank you! Especially since I might want to sue wow1/2 as sensitivity params, as you mentioned in slack. Just changed the names.

resources/shaders/sigmoid_dance.fs
andrewlook: Thanks for explaining that! I hadn't traced through the code to see exactly where that was happening, so it's nice to understand why that param is there. Fixed now.

resources/shaders/sigmoid_dance_audio.fs
```diff
@@ -0,0 +1,73 @@
...
+    st -= iTranslate;
+
+    float index = mod(st.x * TEXTURE_SIZE * 2.0, TEXTURE_SIZE);
+    float wave = (0.5 * (1.0+texelFetch(iChannel0, ivec2(index, 1), 0).x)) - 0.25;
+    float p = floor(index / pixPerBin);
+    float tx = halfBin+pixPerBin * p;
+    float freq  = texelFetch(iChannel0, ivec2(tx, 0), 0).x;
+
+    float norm_x = pow(wave, 1. / iWow2);
+    float norm_y = pow(freq, 1. / iWow1);
```
andrewlook, line 66: TODO: find a better way to parametrize this for audio-sensitivity

zranger1: Consider sending avgVolume in as a normal
```
shader.setUniform("avgVolume",avgVolume.getValuef());
```
   and playing with something like this:
```
float norm_x = (wave / avgVolume); norm_x *= norm_x;
float norm_y = (wave / avgVolume); norm_y *= norm_y;
```
it generates a lot of movement, even in quiet passages.

src/main/java/titanicsend/pattern/yoffa/shader_engine/NativeShader.java
```diff
@@ -210,6 +210,11 @@ private void setStandardUniforms(PatternControlData ctl) {
         setUniform("bassLevel", (float) ctl.getBassLevel());
         setUniform("trebleLevel", (float) ctl.getTrebleLevel());
 
+        // added by @look
+        setUniform("bassRatio", (float) ctl.getBassRatio());
+        setUniform("trebleRatio", (float) ctl.getTrebleRatio());
+        setUniform("volumeRatio", ctl.getVolumeRatiof());
+
```
zranger1, line 217: If you add these guys in this PR, please go ahead and comment out iMouse here too!

zranger1 (approved):
Approved!  Can check in when you're ready -- or I'll re-review if you want to make any more changes.

PR #361, opened 2023-08-04
Pattern Improvements and Fixes

zranger1:
A few bug fixes and control enhancements, along with palette and basic control support for some less-frequently-used patterns.  (Some of which might now become more frequently used!)

## Fixes
**SpiralDiamonds** - now correctly supports view switching
**TESparklePattern** - fixed threading-related crash when switching views

## Improvements

**MatrixScroller**
- added rotation support.  (Wow, this could make people fall over!)

**NeonHeartNative**
- palette support
- basic scale, rotation, angle support
- Quantity controls heart segment length

**PulsingHeart**
- palette support
- basic scale, rotation, angle support
- Wow2 controls background color level

**NeonCells**
- palette support
- basic scale, rotation, angle support
- Wow2 mixes in additional layers

**NeonBarsEdges (and panels)**
- palette support

### DIFF
--- a/resources/shaders/neon_heart.fs
+++ b/resources/shaders/neon_heart.fs
@@ -14,12 +14,16 @@
 #define POINT_COUNT 8
 
 vec2 points[POINT_COUNT];
-float speed = -{%speed[0.5,0.1,2.5]};
-float len = {%len[.25,.01,.5]};
-float scale = {%scale[.012,.006,.024]};
+float scale = 0.012;
 float intensity = 3-{%haze[1.7,0,2]};
 float thickness = {%thickness[0.0035,0.001,0.01]};
 float radius = 0.012;
+
+//  rotate a point around the origin by <angle> radians
+vec2 rotate(vec2 point, float angle) {
+    mat2 rotationMatrix = mat2(cos(angle), -sin(angle), sin(angle), cos(angle));
+    return rotationMatrix * point;
+}
 
 //https://www.shadertoy.com/view/MlKcDD
 //Signed distance to a quadratic bezier
@@ -95,7 +99,7 @@ float getGlow(float dist, float radius, float intensity){
 // Changes in here
 float getSegment(float t, vec2 pos, float offset){
 	for(int i = 0; i < POINT_COUNT; i++){
-        points[i] = getHeartPosition(offset + float(i)*len + fract(speed * t) * 6.28);
+        points[i] = getHeartPosition(offset + float(i)*iQuantity + fract(t) * 6.28);
     }
 
     vec2 c = (points[0] + points[1]) / 2.0;
@@ -127,8 +131,9 @@ void mainImage( out vec4 fragColor, in vec2 fragCoord ){
     vec2 pos = centre - uv;
     pos.y /= widthHeightRatio;
     //Shift upwards to centre heart
-    pos.y += 0.03;
+    //pos.y += 0.03;
 
+    pos = rotate(pos, iRotationAngle) * iScale;
     float t = iTime;
 
     //Get first segment
@@ -140,7 +145,7 @@ void mainImage( out vec4 fragColor, in vec2 fragCoord ){
     //White core
     //col += 10.0*vec3(smoothstep(0.006, 0.003, dist));
     //Pink glow
-    col += glow * vec3(1.0,0.05,0.3);
+    col += glow * iColorRGB;
 
     //Get second segment
     dist = getSegment(t, pos, 3.4);
@@ -149,7 +154,7 @@ void mainImage( out vec4 fragColor, in vec2 fragCoord ){
     //White core
     //col += 10.0*vec3(smoothstep(0.006, 0.003, dist));
     //Blue glow
-    col += glow * vec3(0.1,0.4,1.0);
+    col += glow * iColorRGB;
 
     //Tone mapping
     col = 1.0 - exp(-col);
--- a/resources/shaders/rain_noise.fs
+++ b/resources/shaders/rain_noise.fs
@@ -44,7 +44,7 @@ void mainImage(out vec4 fragColor, in vec2 coord) {
     float ypos = coord.y * iWow1;
 
     // speed variance between cells
-    float time = iTime / -16.;
+    float time = iTime / 16.;
     ypos += (time + noise(time * .2 + hash(vec2(id, 0.0)), id) * SQRT2) * 2.;
 
     // length variance
@@ -68,6 +68,6 @@ void mainImage(out vec4 fragColor, in vec2 coord) {
         bri = length(vec2(xdist,ydist));
     }
     // iWow2 controls the foreground/gradient mix
-    vec3 col = bri * mix(iColorRGB,mix(iColorRGB,iColor2RGB,mod(id,4.) / 4.),iWow2);
-    fragColor = vec4(col,1.0);
+    vec3 col = mix(iColorRGB,mix(iColorRGB,iColor2RGB,mod(id,4.) / 4.),iWow2);
+    fragColor = vec4(col,bri);
 }
--- a/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -1094,7 +1094,7 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     }
 
     public int getGradientColor(float lerp) {
-        return controls.color.getGradientColor(lerp);
+        return TEColor.setBrightness(controls.color.getGradientColor(lerp),(float) getBrightness());
     }
 
     /**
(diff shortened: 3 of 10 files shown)


resources/shaders/rain_noise.fs
```diff
@@ -68,6 +68,6 @@ void mainImage(out vec4 fragColor, in vec2 coord) {
         bri = length(vec2(xdist,ydist));
     }
     // iWow2 controls the foreground/gradient mix
-    vec3 col = bri * mix(iColorRGB,mix(iColorRGB,iColor2RGB,mod(id,4.) / 4.),iWow2);
-    fragColor = vec4(col,1.0);
+    vec3 col = mix(iColorRGB,mix(iColorRGB,iColor2RGB,mod(id,4.) / 4.),iWow2);
+    fragColor = vec4(col,bri);
```
andrewlook, line 72: just curious - does this relate at all to your "blending fixes" PR #345 , which seems to deal with alpha values of native shaders?

and in general, do we want to use alpha for brightness as opposed to multiplying it into the RGB values? That's something I've been wondering about, so this change caught my eye.

src/main/java/titanicsend/pattern/jon/SpiralDiamonds.java
```diff
@@ -44,7 +44,7 @@ public void runTEAudioPattern(double deltaMs) {
         int color = calcColor();
         double squareocity = getQuantity();
 
-         for (LXPoint point : modelTE.panelPoints) {
+         for (LXPoint point : getModel().getPoints()) {
```
andrewlook, line 47: So this change made SpiralDiamonds handle view changes correctly?

I wasn't clear on exactly how the view changes interacted with Java patterns but from this change, I understand that `getModel().getPoints()` gives coordinates correctly normalized for the current view - is that right?

It seems like this will also use all points on the model (edges + panels) as opposed to just panels - I'm assuming that's desired but just wanted to double-check.

resources/shaders/rain_noise.fs
zranger1: (I think this is a copy of rain_noise from the other PR that I forgot to erase, but it can go in this PR anyway!)

You can actually write color & alpha either way, whichever is more convenient for you.  It will work now and with the blending fix when it arrives.  

The blending fix slightly improves automatic calculation of transparency in cases where alpha is clamped to 1.0.  And it doesn't have a huge visual impact when running in the "add" blending mode we normally use for our channels.   It just makes it slightly easier to see what's going on when multiple channels are mixed, and slows down the inevitable color pileup that ends in white!

src/main/java/titanicsend/pattern/jon/SpiralDiamonds.java
zranger1: Yeah, modelTE is the "real" entire model, and getModel().getPoints() returns the current view as determined by the control.  The pattern can set the default in code, but the channel settings and the VJ actually control what points the pattern sees.

resources/shaders/neon_heart.fs
```diff
@@ -95,7 +99,7 @@ float getGlow(float dist, float radius, float intensity){
 // Changes in here
 float getSegment(float t, vec2 pos, float offset){
 	for(int i = 0; i < POINT_COUNT; i++){
-        points[i] = getHeartPosition(offset + float(i)*len + fract(speed * t) * 6.28);
+        points[i] = getHeartPosition(offset + float(i)*iQuantity + fract(t) * 6.28);
```
andrewlook, line 102: should `iQuantity` here be replaced by `iWow2` (just based on your PR description, it seems like that was the intention). Quantity does kinda make sense in this context too FWIW, just wanted to call it out since its the only mismatch I found between params in the PR description vs testing out the patterns.

zranger1: Good catch.   I could go either way here too, but let's leave it with iQuantity and I'll change the description.

andrewlook:
I went through and tested all the patterns before and after this PR. Sharing notes below:

## Fixes

### SpiralDiamonds

- [x] now correctly supports view switching

verified - this one looks awesome hitting panels and edges, especially with rotation! Love it.

### TESparklePattern

- [x] fixed threading-related crash when switching views

reproduced the crash on `main`:
TESparkle_crash

verified that it's fixed on this PR. ✅ 

## Improvements

### MatrixScroller

- [x] added rotation support. (Wow, this could make people fall over!)

verified.

Only note: the pattern seems to have gotten darker around the center, and the brightness parameter doesn't seem to change this at all.

### NeonHeartNative

- [x] palette support
- [x] basic scale, rotation, angle support
- [ ] Wow2 controls heart segment length
  - **iQuantity** controls segment length, when I tested it. Wow2 doesn't have any effect.

### PulsingHeart

- [x] palette support
- [x] basic scale, rotation, angle support
- [x] Wow2 controls background color level

### NeonCells

- [x] palette support
- [x] basic scale, rotation, angle support
- [x] Wow2 mixes in additional layers

### NeonBarsEdges (and panels)

- [x] palette support

one thing I noticed here - the overall brightness seemed to vary a lot between palette colors, without changing any params. Don't think it's a problem, just wanted to point it out as FYI.

Screen Shot 2023-08-04 at 6 35 30 PM
Screen Shot 2023-08-04 at 6 35 15 PM

zranger1:
@andrewlook, you found an interesting bug.  Brightness is actually supposed to work w/MatrixScroller and it doesn't.  Apparently we missed a method (getGradientColor()) back when we did the original color control refactor.  Fixed now!

zranger1:
Oh, and you have to resolved mark conversations as resolved -- it'll perma-block the merge if I do it.

zranger1:
On MatrixScroller's brightness:   It's super subject to the swatch and gradient.  Also the default parameterization is really weird in EDC_showcase.   Press the Panic button to get it positioned back in the right place, then play with the color controls to see what it's doing.  Brightness should work after the last comit.  (Also, try rotatating it w/the "Panels:Sections" view up -- that's the one that's going to hypnotize people!)

resources/shaders/rain_noise.fs
andrewlook: oh cool! i hadnt thought about the impacts when blending multiple patterns (but I'd noticed that white-pileup effect). thanks for explaining! 👍

src/main/java/titanicsend/pattern/jon/SpiralDiamonds.java
andrewlook: that makes sense, thanks! I'd been wondering where exactly the View was changing the coordinate space, but I just noticed `TEPerformancePattern.getModel()` referencing `TEPerformancePattern.model` and saw how `viewPerPatternListener` swaps that out when the parameter is changed. 👍

andrewlook:
OK, marked conversations resolved (thanks for that, I wasn't aware that would block if the submitter resolves them).

Glad that spotting the MatrixScroller brightness bug was useful - I'll re-test shortly.

src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -1094,7 +1094,7 @@ public int getColor() {
     }
 
     public int getGradientColor(float lerp) {
-        return controls.color.getGradientColor(lerp);
+        return TEColor.setBrightness(controls.color.getGradientColor(lerp),(float) getBrightness());
```
andrewlook, line 1097: nice catch! 👍

andrewlook (approved):
OK this looks good - I verified that matrixScoller works better with something besides the default gradient value set! Looks awesome now. (non-PR-blocking idea: I wonder if we can make the pattern default to one of the appropriate gradient values, the way we set per-pattern defaults for size/speed/etc, so that it doesn't default to the darker gradient combo?)

Screen Cast 2023-08-04 at 8 10 44 PM 2

src/main/java/titanicsend/pattern/TEPerformancePattern.java
zranger1: Glad you tried the control!  Can't believe we missed that, but as you've seen, TEPerformancePattern.java got complex really, really quickly.   

(And all default parameter settings are saved per pattern, per channel, with the project file.   I'd bet that this one just got created in a huge hurry as the EDC deadline approached.  Hopefully we'll have time to do a new, even cooler setup for the burn.)

Also, easy parameterization of patterns is a huge issue for the future.  Our real-time, reactive control setup is way state of the art, and it's definitely not how most people control lights.  We need to make it simpler.

The patterns have a huge diversity of behaviors depending on how you set those knobs, and the only way to figure it out is to play with them, which nobody has time to do during a performance.   So people stick pretty close to the project file setup and change a few things they're familiar with.  It makes shows less interesting than they probably could be. 

We should think about adding something like a "patch" control that lets users move quickly between a range of stored variants at performance time, same way as they do views.  (And ideally, you could play with the patterns away from the car and store and edit your own patches too.)

PR #362, opened 2023-08-05
DMX Engine for Beacon and DJ Lights

jkbelcher:
First version of DMX Engine for controlling DMX fixtures including beacons and dj lights.

The dmx engine runs alongside the normal LXMixerEngine, keeping most of the dmx values outside of the scope of channel faders.  It blends dmx-aware patterns from any channel, passing the LX channel fader levels through to only the relevant dmx values (usually Dimmer).

The patterns here are basic for testing but can be expanded once an end-to-end test works.

Still some fine tuning and distilling to be done, but going ahead and making this available for hardware testing and first reviews.  ~~Of note this did not get tested on the beacon yet but will have a chance to do so on Tuesday.~~  Edit: Confirmed this works on a beacon.

Demo of how to load and test a pattern:
https://youtu.be/g-7c93XCbSg


src/main/java/titanicsend/dmx/pattern/DjLightsEasyPattern.java
```diff
@@ -0,0 +1,123 @@
...
+    int strobeSpeed = this.strobeSpeed.getValuei();
+    boolean mirror = this.mirror.getValueb();
+    double dimmer = this.dimmer.getNormalized();
+    double focus = this.focus.getNormalized();
+
+    int color = this.linkedColor.calcColor();
+    int r = ((color >> 16) & 0xff);
+    int g = ((color >> 8) & 0xff);
+    int b = (color & 0xff);
+    int w = (r < g) ? ((r < b) ? r : b) : ((g < b) ? g : b);
```
worldveil, line 89: haha whaaaat

zranger1: (This is one of the "normal" ways of extracting the "white" level for driving RGBW devices.   The amount of white light in a color is taken as roughly equal to the minimum of the other 3 channels. )

zranger1 (approved):
@jkbelcher, this is mighty!   Wow, those beacons have a lot of features!  

Can verify that this sends valid DMX data over the net, and that it keeps working through multiple project load/reload cycles. (No real DMX lights here, so can't validate the output data.  It looks like it's in the proper ranges though. )  

Haven't seen it interfere with or change any other existing system.   Approving so we can start working with it.

PR #364, opened 2023-08-07
Audio-reactive pattern: TriangleInfinity

andrewlook:
Controls Common to `TriangleInfinityLevels` and `TriangleInfinityWaveform`:
- scale, rotation, speed
- `iQuantity`: number of bands/layers to add
- `iWow`: reactivity to volume/bass/treble levels

Controls Specific to TriangleInfinityWaveform:
- `iWow2`: how much the raw audio waveform modifies the shape

### `TriangleInfinityWaveform`

https://github.com/titanicsend/LXStudio-TE/assets/236018/4d47cd36-b2a9-49e5-9a97-859b6872e675

### `TriangleInfinityLevels`

https://github.com/titanicsend/LXStudio-TE/assets/236018/7b33dacf-3e61-4625-a747-143716b22a7d

## details

This shader is based on the idea of a triangle distance field, with a neon effect added and a noise function to make some random-seeming offshoots.

https://github.com/titanicsend/LXStudio-TE/assets/236018/d95fb5fc-56ea-4814-ba09-d6c3ccdc2c17

The "quantity" param mentioned above controls how many layers of overlapping triangles get drawn - the higher the quantity, the more wispy small layers show up.

https://github.com/titanicsend/LXStudio-TE/assets/236018/8f8637e9-026a-448a-bfc1-1db532d9bfe3

The interesting shapes come from fractalizing and morphing the space at each layer (this param is hooked up to audio levels to make the pattern dance):

https://github.com/titanicsend/LXStudio-TE/assets/236018/c71d81ea-c348-4da3-807e-cb40bc1c90fd


src/main/java/titanicsend/pattern/yoffa/shader_engine/NativeShader.java
```diff
@@ -210,6 +210,11 @@ private void setStandardUniforms(PatternControlData ctl) {
         setUniform("bassLevel", (float) ctl.getBassLevel());
         setUniform("trebleLevel", (float) ctl.getTrebleLevel());
 
+        // added by @look
```
zranger1, line 213: This is good!  We should've done it sooner.  While you're in this file, can you go ahead and comment out ```setUniform("iMouse"...)``` above?  (Another thing I've been meaning to do.)

We don't support it, and GL will automatically initialize it to zero and leave it there, so it's 4 32-bit values per frame we really don't need to format and pass to the GPU.

resources/shaders/triangle_infinity_radial_waveform.fs
```diff
@@ -0,0 +1,141 @@
...
+        uv.y *= -1.;
+        // uv0.y *= -1.;
+        // uv.y *= -1. * i;
+        // uv.y += 0.2;
+        // uv.y += 0.02 * abs(sin(noise(i, iTime*0.02)));
+        // uv.y += 0.02 * noise(i, iTime*0.02);
+
+        float d = 1.;
+        d *= sdEquilateralTriangle(uv, 0.9) + iWow2*wave;
+        d *= exp(-sdEquilateralTriangle(uv0, 0.9));
```
zranger1, line 115: This (```d *= exp(-sdEquilateralTriangle(uv0, 0.9));```) doesn't change and can be done once outside the loop.

resources/shaders/triangle_infinity_radial_waveform.fs
```diff
@@ -0,0 +1,141 @@
...
+        d += 0.3*noise(i*d, iTime*0.1);
```
zranger1, line 116: Consider eliminating both calls to this noise function -- it's just adding randomly colored pixels, which are what's really causing that "static" that you're trying to eliminate on line 129 below.    (Or maybe use the noise to perturb the coordinates rather than directly coloring pixels.)  

Your call -- I tested w/o the noise and last static reduction step though, and it's quite beautiful.
Chromatik — Titanic's End 8_8_2023 4_17_20 PM

zranger1:
This is looking great!   Just a couple of comments on triangle_infinity_radial_waveform.fs that might apply to all the shaders.

resources/shaders/triangle_infinity.fs
```diff
@@ -85,6 +85,7 @@ void mainImage(out vec4 fragColor, in vec2 fragCoord) {
     // normalize coordinates
     vec2 uv = fragCoord.xy / iResolution.xy;
     uv -= 0.5;
+    uv -= vec2(0., {%defaultYOffset[-0.17,-0.5,0.5]});
```
zranger1, line 88: In general, I'd recommend doing this sort of thing as a constant instead of as a uniform.   Each uniform adds to the per-frame setup cost, so it's best to use as few as you can get away with.

src/main/java/titanicsend/pattern/yoffa/shader_engine/NativeShader.java
andrewlook: excellent, done!

resources/shaders/triangle_infinity_radial_waveform.fs
andrewlook: good idea, thanks! I made this change in waveform/radial waveform versions (where the static was more chaotic)

resources/shaders/triangle_infinity_radial_waveform.fs
andrewlook: great catch! made this change

resources/shaders/triangle_infinity.fs
andrewlook: ah yes thank you. I was considering using a uniform to make some "preset" versions of the pattern by just making different patterns that set their own defaults (when these defaults fall outside the regular values - in this case, I noticed that "panic" button always resets xpos/ypos to `0` regardless of the pattern's default control values. But I ended up not using this uniform, so I went ahead and just made this a `const float` 👍

andrewlook:
> This is looking great! Just a couple of comments on triangle_infinity_radial_waveform.fs that might apply to all the shaders.

wonderful! thanks for the review here @zranger1 , this was very helpful

zranger1 (approved):
These look really good!

PR #368, opened 2023-08-11
Log number of beacons+DJ lights, warn if none found

jkbelcher:
Added stats for the DMX fixtures to TE's summary startup message.  Added warning if none were found, which would be the case if someone forgets to update the config files.


worldveil:
Tested, works OK. doesn't affect anything without the beacons plugged in, anyway

worldveil (approved):

PR #369, opened 2023-08-11
Move Autopilot button to top, organize other custom UI sections

jkbelcher:
Moved Autopilot button to the top for quick access.  Moved the other custom UI sections too... 

The Global pane contains things that would be actively used during a run session:
image

The Model pane now contains things that are generally set once such as output controls:
image


worldveil:
Tested it out, looks good!

worldveil (approved):

PR #370, opened 2023-08-11
Add BeaconEverything pattern

jkbelcher:
This is the pattern that you *actually* want to use controlling a beacon.  It distills out all the DMX junk and provides a normal set of controls.  Doesn't require any special knowledge of shutter settings.

image


worldveil (approved):
Going to merge since it works in LXstudio without crashing. Will test with beacon soon

PR #371, opened 2023-08-12
Update Views to support port side panel sections

zranger1:
Now that the port side of the car is completely lit, it should be symmetrical with the starboard side when possible.  This PR updates the TEPanelSections and views so that the "Panel Sections"  view works correctly on both sides of the car. 

It also fixes a couple of patterns that broke when the simple "TEPanelSections.PORT" descriptor got more complicated.


resources/vehicle/views.txt
```diff
@@ -6,7 +6,7 @@ Edges	true	Edge
...
 Panels	true	Panel
 Panels;	true	FA;FB;FSA;FSB;FSC;FPA;FPB;FPC;AA;AB;ASA;ASB;ASC;APA;APB;APC;SFA;SFB;SEA;SFC;SFD;SEB;SEC;SED;SDA;SDB;SDC;SEE;SCA;SCB;SCC;SBA;SBB;SBC;SBD;SBE;SAA;SAB;SAC;SAD;SUF;SUA;PAA;PAB;PBA;PAC;PAD;PBB;PBC;PBD;PCA;PCB;PCC;PBE;PDA;PDB;PDC;PEA;PEB;PEC;PED;PEE;PFA;PFB;PFC;PFD;PUF;PUA
-Panel Sections	true	fore;aft;port;starboard-fore;starboard-fore-single;starboard-aft;starboard-aft-single
+Panel Sections	true	fore;aft;starboard-fore;starboard-fore-single;starboard-aft;starboard-aft-single;port-fore;port-fore-single;port-aft;port-aft-single;
```
jkbelcher, line 9: Doesn't need a semicolon after the last item

src/main/java/titanicsend/model/TEPanelModel.java
jkbelcher, line 1: Is this an alternate to Yoffa's changes to TEWholeModel in #352 ?

jkbelcher (commented):
This looks good from reading the code.  One typo and one question...

resources/vehicle/views.txt
zranger1: Extra semicolon removed!

src/main/java/titanicsend/model/TEPanelModel.java
zranger1: It works that way, though it wasn't intended as such, and it has the virtue of simplicity.

I'd actually recommend not taking 352 because the texture mapping on the car ends doesn't come out right.  But in the longer run, we might to use Yoffa's approach to associating model data w/section tags.

jkbelcher (approved):
I ran this, confirmed the port side panels look correct.  Read the code changes, they also look correct.

PR #376, opened 2023-08-12
Backside Config

fakeplant:
Includes:
- New config for backside edges and panels.
- Fixes to existing edge/panel config for components that were not lit at EDC.

Config changes should match latest hardware mounting documentation.
[TE - Modules, Panels, Edges - Routing Guide and Visual Documentation.pdf](https://github.com/titanicsend/LXStudio-TE/files/12391684/TE.-.Modules.Panels.Edges.-.Routing.Guide.and.Visual.Documentation.pdf)


raldi:
Looks like this covers all panels, and all edges except three:

33-38
36-37
45-46

Any idea what's up with those three?

jkbelcher (approved):
Hopefully I’m not stepping on Mike’s toes by approving this. Just don’t want Eli held up on playa waiting on a stamp.

jkbelcher:
Also I’m away from the house and have not tested this.

raldi:
Yep, go ahead and merge. I was just curious about the question above.

fakeplant:
@raldi
Those are all horizontal cross beams between front and back side modules on bridge and peak.
We'll need to do a little bit of work on playa to figure out 1) if we even have edges on them 2) where we can add them to existing chains.

fakeplant:
@raldi
I've created new config for the horizontal edges.
We have natural extension spots for each on existing chains, so it's easy on two of the three.

The current problem is 36-37. For now I will leave commented out until we solve on playa.

It is not possible to add it to the closet chain because of the length of the entire chain becoming 506.
LX throws an error message if over 505 pixels.

**Q:** 
I see a `MAX_PIXELS_PER_CHANNEL` set to 505.
Can you explain reasoning for that? Is it a power draw thing? Or is it based on artnet?
If its based on artnet, why is it not 510 (ie. 170 pixels per universe * 3 layered universes per strip)


Options:
1. Change software to allow for 506. This will end up being longest chain on car and could be a power issue, but easiest from install.
2. Add a brand new power backpack just for this edge. It solves any power issues, and we have spare boxes to support, only downside is finding a spot for it that isn't hanging upside down in peoples faces.
3. Rewire backside of car. We haven't actually wired anything on backside so it's not ton of rework, however, having un-mirrored pixels is less than ideal, and not clear we'd find any chain solutions that fix our problem (would require more work to find as well)

Screenshot 2023-08-21 at 6 12 53 PM
c86e">

raldi:
It was originally max 500, but then some strands were manufactured with 51px instead of 50, so we had a couple runs with 503, so I bumped it to 505.

510 should be fine.

raldi:
There’s also option 0: Skip these three edges. None of them are adjacent to a panel, and the car might look better without em. But if we do, we should remove their lines from the config altogether.

PR #378, opened 2023-08-13
Audio-reactive pattern: TriangleCross

andrewlook:
There are 2 patterns in here:
- `TriangleCrossAudioLevels`
- `TriangleCrossAudioWaveform`

image

Both patterns use audio levels in the following way (these are modulated by `iWow1` - if it's set to zero, these effects stop entirely):
- volumeRatio: controls outer triangle size
- bassLevel. / trebleLevel: control dimensions of inner cross, making them pulse with the music.

`TriangleCrossAudioWaveform` has the outer triangle perturbed by the raw audio waveform (modulated by `iWow2`), with the waveform mapped radially around the triangle.

Size/spin/xpos/ypos work as expected.

When the color swatch has 2 non-black colors, the inner cross will bleed out onto the outer triangle for a cool effect :)

image


resources/shaders/framework/template.fs
```diff
@@ -13,6 +13,10 @@ uniform float sinPhaseBeat;
 uniform float bassLevel;
 uniform float trebleLevel;
 
+uniform float volumeRatio;
+uniform float bassRatio;
+uniform float trebleRatio;
```
andrewlook, line 18: note: this is duplicated from #364 , will dedupe whichever from merges first

zranger1 (approved):
Leaving figuring out the merge to you on this one!

PR #382, opened 2023-08-16
Dev Switch

jkbelcher:
This is a new tool designed to fix and prevent a lot of problems, particularly showtime playa headache.  It adds a panel in the left pane called Dev Switch that has buttons for all the things you may want to turn on and off:

image

It also has two buttons across the top, **Developer** and **Production**.  Clicking either one will set the appropriate settings AND hold them when a new project file is opened **AND save it to the local machine and apply the settings on the next run**.  So for example we can lock Lighting-1 to Production and everything will always turn on no matter what.  And we can set our laptops to Developer and no longer have to race to turn off Live output after pulling from main.

If you change any of the settings within LX the DevSwitch will turn off the highlighted Developer/Production switch to indicate the state is out of sync.  But the most recent of Dev/Prod will be applied again on the next run.

**Motivation:**

As you may be aware, there are a lot of components than can be turned on and off in this setup.  This is tricky in a few places:

1.  It can be hard to remember them all and find them under the pressure of showtime.  Most of them save to the project file but if they were overlooked when the project file was created then you can be racing to find them at showtime.
2. In the case of MIDI hardware, it can be impossible to get all the midi devices enabled in the LXP show file when they're physically not on hand.
3. Some of these components never had user-facing toggles to disable them, such as the LED pixel output and the DMX lights.  For example I've had to spam my home network with LED data while trying to test the beacons.
4. At EDC we were fighting ongoing interference from developer laptops which upon opening LX would start sending OSC and pixel data.  Some of the pieces, especially the custom ones, were hard to turn off.  But you want to be able to run dev laptops on the production network without causing interference.
5. At home we're always having to race to turn off Live output after pulling from main.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ classes/
 Projects/project.lxp
 resources/shaders/cache/*.bin
 
+.devSwitch
--- a/src/main/java/heronarts/lx/studio/TEApp.java
+++ b/src/main/java/heronarts/lx/studio/TEApp.java
@@ -44,6 +44,8 @@ import titanicsend.app.TEOscListener;
 import titanicsend.app.TEUIControls;
 import titanicsend.app.TEVirtualOverlays;
 import titanicsend.app.autopilot.*;
+import titanicsend.app.dev.DevSwitch;
+import titanicsend.app.dev.UIDevSwitch;
 import titanicsend.dmx.DmxEngine;
 import titanicsend.dmx.pattern.ExampleDmxTEPerformancePattern;
 import titanicsend.dmx.pattern.BeaconDirectPattern;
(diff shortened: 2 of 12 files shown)


zranger1 (approved):
Can confirm:
- disabling various types of output stops traffic to the applicable address ranges.      
- save and restore of Dev/Production mode works as specified.

Approved.  This is an elegant and useful addition!  (Request for next year - can we have it optionally save local dev/production settings, since you've already got gitignore'ed file set up?)

jkbelcher:
> (Request for next year - can we have it optionally save local dev/production settings, since you've already got gitignore'ed file set up?)

Good idea.  Wouldn't want it to happen automatically, ie just because someone turns off beacons doesn't mean we don't want them again tomorrow.  But the option to customize Dev/Production per computer would be nice.

PR #383, opened 2023-08-17
Disable Explode on Non-shader Patterns

zranger1:
For now, "Explode" effect is only available on GL shader patterns - this PR disables the control on all Java patterns.


andrewlook (approved):
@zranger1 thanks for making this fix!

Clever use of the missingControls file to find all non-shader patterns.

I tested this out locally, didn't exhaustively go through all java patterns but every java pattern I tested had the explode knob hidden (and all shaders still had the explode knob).

PR #385, opened 2023-08-19
Don't reset beacon position when channel fader down

jkbelcher:
Silly oversight, beacon position was resetting to default with the channel fader down.  Different from LEDs, the dmx frame needs to be blended even if the color is black.


zranger1 (approved):
Good catch!

PR #386, opened 2023-08-19
Allow CrutchOSC control of parameter by name

jkbelcher:
Simple change, allows OSC messages coming to the focused pattern to access parameters by path in addition to index.  Enables a work-in-progress on the TouchOSC side to get buttons for each view.


zranger1 (approved):
Looking forward to view buttons!

PR #387, opened 2023-08-21
Intel Mac Support

fakeplant:


jkbelcher (approved):
Sorry this held you guys up.  This looks like the right fix.

The newest libs are 0.4.2.TE.6-SNAPSHOT, though it won't hurt to fix the older 0.4.5-SNAPSHOT pom as well.

PR #397, opened 2023-09-30
Panel & Vertex Labels + other small 3d improvements.

zranger1:
#### Features and Improvements:
- Panel and Vertex labels are back! 
- Added generic 3D text capability, so it is relatively easy to add labels as needed.
- 3D model display should be **much brighter** now due to backing panel rendering adjustments.
- Backing panels are now really, really opaque.  

#### Notes 
- Includes FontMaker tool, in case we ever want to change or add 3d fonts.  It converts a TrueType font into a small,
alpha-only texture atlas for use by the 3D text renderer.  (it's in ui/text3d/FontMaker.java -- just run it from the IDE.)

- Added basic support for TE-custom BGFX shaders in 3D model display. This is used by the
3D text engine, but can eventually help render lasers, beacons, etc.

I've provided tools and shell scripts to help with the somewhat awkward process of compiling BGFX shaders for
every supported platform. Fortunately, recompilation should be a rare event.


jkbelcher (approved):
Awesome work here @zranger1 !

I read through the code, all looks reasonable and well formatted.  Ran the UI on macOS and the labels look great.  The black backgrounds on panel labels are nice for keeping them readable when pixels are lit.  The semi-transparent backgrounds on vertex labels are good too.  Backing panel adjustments look good.

I did not test the shader compilation scripts.  Assuming they are fine or can be fixed later if needed.  I pushed a small commit to remove some accidental paragraph breaks.

Extra points for the multi-platform compilation support. ⭐

Merging!

PR #398, opened 2023-10-08
Move to CC BY-NC-SA 4.0 license

jkbelcher:
In March 2023 the TE software team decided to change licenses to the CC BY-NC-SA 4.0 license.  It was understood that work up until that point would remain under the MIT license but everything moving forward would be under the new license.  This PR updates the LICENSE.md file which had not been updated to reflect the intended change.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -15,7 +15,7 @@ The contents of:
     /src/main/java/titanicsend
     /test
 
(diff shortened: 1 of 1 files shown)


jvyduna:
Thank you so much. This has been on my Todo forever and it was much simpler than I imagined.

jvyduna (approved):

PR #400, opened 2023-10-11
Update Chromatik libs to TE.7-SNAPSHOT

jkbelcher:
Latest chromatik libraries.  Someone please pull and test, and be sure to do a `mvn clean`.  Note the linux lwjgl profiles are commented out here.


zranger1:
Tested & working on Windows and Ventura.   Will approve, but just want to bring one weirdness to your attention:   The default behavior of vehicle.lxp is different on Mac and Windows.   Mac came up perfectly normal after clean build, but in Windows the axis display is enabled for all views, and the cue/aux 3d views are aligned very differently (as seen in screenshot below). 

After adjusting this stuff, saving the .lxp and reloading, everything's fine, and I can definitely live with it.   It's just... weird.

image

zranger1 (approved):
Working as expected on Mac and PC.  Minor startup differences between Mac/PC can be fixed along the way, if necessary.

jkbelcher:
> The default behavior of vehicle.lxp is different on Mac and Windows. Mac came up perfectly normal after clean build, but in Windows the axis display is enabled for all views, and the cue/aux 3d views are aligned very differently (as seen in screenshot below).

Interesting.  The axes visibility is saved in the .lxp.  Surprised it would come through as a different default.

PR #402, opened 2023-10-12
DMX Modulators

jkbelcher:
Added modulators: DmxDualRange modulator can be used for the Strobe input, where 0=off, 1-127 is range1, and 128-255 is range2.  DmxColorModulator can be used to route RGB values to a color in the global palette.  The existing "DMX" modulator can be used for the global dimmer.

image

image

### DIFF
--- a/src/main/java/titanicsend/modulator/dmx/Dmx16bitModulator.java
+++ b/src/main/java/titanicsend/modulator/dmx/Dmx16bitModulator.java
@@ -1,0 +1,37 @@
+package titanicsend.modulator.dmx;
+
+import heronarts.lx.LXCategory;
+import heronarts.lx.dmx.DmxModulator;
+import heronarts.lx.dmx.LXDmxEngine;
+import heronarts.lx.modulator.LXModulator;
+
+/**
+ * A DMX modulator consuming two bytes of DMX data for high resolution
+ *
+ * @author Justin K. Belcher
+ */
+@LXModulator.Global("DMX 16-bit")
+@LXModulator.Device("DMX 16-bit")
+@LXCategory("DMX")
+public class Dmx16bitModulator extends DmxModulator {
+
+  public Dmx16bitModulator() {
+    this("DMX 16-bit");
+  }
+
+  public Dmx16bitModulator(String label) {
+    super(label);
+    this.channel.setRange(0, LXDmxEngine.MAX_CHANNEL - 1);
+  }
+
+  @Override
+  protected double computeValue(double deltaMs) {
+    int universe = this.universe.getValuei();
+    int channel = this.channel.getValuei();
+
+    byte byte1 = this.lx.engine.dmx.getByte(universe, channel);
+    byte byte2 = this.lx.engine.dmx.getByte(universe, channel + 1);
+
+    return (((byte1 & 0xff) << 8) | (byte2 & 0xff)) / 65535.;
+  }
+}
--- a/src/main/java/titanicsend/modulator/dmx/DmxDualRangeModulator.java
+++ b/src/main/java/titanicsend/modulator/dmx/DmxDualRangeModulator.java
@@ -1,0 +1,100 @@
+/**
+ * License notes: Expecting to contribute these DMX modulators back to LX upstream
+ */
+
+package titanicsend.modulator.dmx;
+
+import heronarts.lx.LXCategory;
+import heronarts.lx.dmx.DmxModulator;
+import heronarts.lx.modulator.LXModulator;
+import heronarts.lx.parameter.BooleanParameter;
+import heronarts.lx.parameter.CompoundParameter;
+
+/**
+ * Modulator receiving two ranges of input from one DMX channel.
+ * Input     Output
+ *  0        Output1 = 0    Output2 = 0
+ *  1-127    Output1 = 0-1  Output2 = 0
+ *  128-255  Output1 = 0    Output2 = 0-1
+ *
+ * @author Justin K. Belcher
+ */
+@LXModulator.Global("DMX Dual Range")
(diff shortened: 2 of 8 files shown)


jkbelcher:
Added 16-bit modulator for high resolution DMX input.  We'll use that for Speed.

jkbelcher:
Added GrandMA.lxp demonstrating all the DMX inputs mapped, except for random strobe and speed.  The brightness input maps to a Hue+Saturation effect on the master channel and the 1-127 strobe input maps to a Strobe effect on the master channel.

src/main/java/titanicsend/modulator/dmx/Dmx16bitModulator.java
```diff
@@ -0,0 +1,37 @@
...
+
+  @Override
+  protected double computeValue(double deltaMs) {
+    int universe = this.universe.getValuei();
+    int channel = this.channel.getValuei();
+
+    byte byte1 = this.lx.engine.dmx.getByte(universe, channel);
+    byte byte2 = this.lx.engine.dmx.getByte(universe, channel + 1);
+
+    return (((byte1 & 0xff) << 8) | (byte2 & 0xff)) / 65535.;
```
andrewlook, line 35: Just curious, is the bit-masking `& 0xff` necessary, since `byte` is already an 8-bit representation? (If i understand correctly, `a & 0xFF` == `a` for any value of `byte a`)

src/main/java/titanicsend/modulator/dmx/DmxDualRangeModulator.java
```diff
@@ -0,0 +1,100 @@
...
+      this.output2.setValue(0);
+      return 0;
+    } else if (dmx < 128) {
+      double value = (dmx - 1) / 126.;
+      this.isZero.setValue(false);
+      this.range1active.setValue(true);
+      this.range2active.setValue(false);
+      this.output1.setValue(value);
+      this.output2.setValue(0);
+      return value;
```
andrewlook, line 85: why does this `value` get returned for only this part of the conditional? it seems like `output1` gets set here, which is the value we expect to get used?

src/main/java/titanicsend/modulator/dmx/DmxDualRangeModulator.java
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      double value = (dmx - 128) / 127.;
+      this.isZero.setValue(false);
+      this.range1active.setValue(false);
+      this.range2active.setValue(true);
+      this.output1.setValue(0);
+      this.output2.setValue(value);
+      return 0;  // modulator value is output1
```
andrewlook, line 93: should this comment read `output2`?

src/main/java/titanicsend/ui/modulator/UIDmxColorModulator.java
```diff
@@ -0,0 +1,50 @@
...
+    .addChildren(
+        row("Universe", new UIIntegerBox(PARAM_WIDTH, 16, modulator.universe)),
+        row("Channel", new UIIntegerBox(PARAM_WIDTH, 16, modulator.channel)),
+        row("Byte Order", new UIDropMenu(PARAM_WIDTH, modulator.byteOrder))
+        )
+    .addToContainer(uiModulator);
+
+    UI2dContainer.newVerticalContainer(70, 2)
+    .addChildren(
+        new UILabel(70, 16, "Palette Color").setTextAlignment(Align.LEFT, Align.MIDDLE),
```
andrewlook, line 34: Maybe a name like "Global Palette Index" or something would be a bit more descriptive?

andrewlook (approved):
Looks great to me, and everything worked when I tested it. Left a few small comments / questions but they're non-merge-blocking

src/main/java/titanicsend/modulator/dmx/Dmx16bitModulator.java
jkbelcher: Good question.  I am always wary of issues with type conversions and signed/unsigned.  In this case I was looking at the [LXColor.rgba()](https://github.com/heronarts/LX/blob/dev/src/main/java/heronarts/lx/color/LXColor.java#L283) method for guidance.

src/main/java/titanicsend/modulator/dmx/DmxDualRangeModulator.java
jkbelcher: Every modulator has a built-in normalized single output value. It is the source that gets mapped if you use the icon on the top of the modulator:
image
In this case there are two outputs.  The correct way to map (either, but particularly the second) output it to create a new global mapping, select the first or second range (meter) as the source, then select a destination.  However if someone created a mapping with the modulator's built-in Map button, I am making the output look exactly like Output1 instead of jumping between Output1 and Output2.

src/main/java/titanicsend/modulator/dmx/DmxDualRangeModulator.java
jkbelcher: Per above, it is always output1's value.

src/main/java/titanicsend/ui/modulator/UIDmxColorModulator.java
jkbelcher: Yeah I started with "color index" but felt like seeing 0-based values in the UI felt weird, and 1-based values look better as they match the OSC addresses.  The word "index" feels 0-specific so I took that word out.  I'm open to ideas here.

jkbelcher:
> Left a few small comments / questions but they're non-merge-blocking

Unresolved conversations are merge-blocking.  Resolving them now to merge but feel free to keep discussing.

PR #404, opened 2023-10-13
Global Speed control (for DMX integration)

zranger1:
Allows a single modulator or external source to control the speed of all currently running patterns.   

To test, load the project file, " global control test.lxp".   This will load several channels of patterns, and add the effect "GlobalPatternControl" to the main channel, with global speed control enabled and driven by  a modulator. 

Note that:
- all running patterns speed up and slow down together with the modulation source
- speed is appropriately scaled according to each pattern's original control range
- you can still use the individual pattern speed controls to change a pattern's movement direction (foreward/backward) if you wish.  The global control handles speed, but respects individual direction settings.
- disabling global control with the "enable" button in the effect will return all control to individual channels.   The patterns will return to running according to their local control settings.
- the effect can be used with any modulation source, or you can drive it manually w/the knob
- we can easily extend this mechanism to other common controls for even better DMX integration.  Can they give us another channel or two?

Short video at:
https://www.youtube.com/watch?v=p_mGl36Q4rk

### DIFF
--- a/src/main/java/heronarts/lx/studio/TEApp.java
+++ b/src/main/java/heronarts/lx/studio/TEApp.java
@@ -36,13 +36,7 @@ import heronarts.lx.mixer.LXChannel;
 import heronarts.lx.pattern.LXPattern;
 import heronarts.lx.pattern.form.PlanesPattern;
 import heronarts.lx.pattern.texture.NoisePattern;
-import titanicsend.app.GigglePixelBroadcaster;
-import titanicsend.app.GigglePixelListener;
-import titanicsend.app.GigglePixelUI;
-import titanicsend.app.TEAutopilot;
-import titanicsend.app.TEOscListener;
-import titanicsend.app.TEUIControls;
-import titanicsend.app.TEVirtualOverlays;
+import titanicsend.app.*;
 import titanicsend.app.autopilot.*;
 import titanicsend.app.dev.DevSwitch;
 import titanicsend.app.dev.UIDevSwitch;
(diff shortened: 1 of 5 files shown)


src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -1206,7 +1211,13 @@ public double getDeltaMs() {
     }
 
     public double getSpeed() {
-        return controls.getValue(TEControlTag.SPEED);
+        BoundedParameter speedControl = (BoundedParameter) controls.getControl(TEControlTag.SPEED).control;
+        double spd = speedControl.getValue();
+        if (globalControls.useGlobalSpeed.isOn()) {
+            double g = globalControls.globalSpeed.getValue();
```
jkbelcher, line 1217: Could do `getNormalized()` here, to avoid problems if someone messes with the range on the `globalSpeed` definition.

src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -137,6 +132,10 @@ public Plugin(LX lx) {
       // Saved options for UI overlays
       lx.engine.registerComponent("virtualOverlays", this.virtualOverlays = new TEVirtualOverlays(lx));
 
+      // set up global control values object so all patterns can potentially be controlled
+      // by a single external input (e.g. DMX controller )
+       lx.engine.registerComponent("globalPatternControls", new TEGlobalPatternControls(lx));
```
jkbelcher, line 137: Accidental indent space 🔎

jkbelcher (approved):
Looks great!  Perfect placement for a global speed override honored by every pattern.  This will save a ton of setup work, no need to manually map to every pattern instance.  Nice!

I made a handful of annoying professor-like comments just to be thorough and maybe educational to anyone following along.  Nothing blocking, feel free to mark conversations as resolved and merge when you're ready.

jkbelcher:
Actually I'm going to mark these conversations as resolved and merge so I can use this from @andrewlook 's PR.

PR #405, opened 2023-10-13
Delete ColorCentral and swatchPerChannel

jkbelcher:
IMO swatch-per-channel was a failed experiment.  Having each channel point to a different swatch was chaos when 2+ channels were running and the user was attempting to queue up a 3rd channel and find a matching swatch.  I think the future is to have A/B master swatches at the global palette level.

Making this change available to merge if we want to commit.


zranger1 (approved):
Code looks fine!  Tested, confirmed that per-channel swatch control is gone, global palette switching works as expected, and resource use (and the rest of Chromatik) is stable.   Approved - merge when ready.

Hang on to this code though!  It may be that we can redesign our palettes and controls a little and come up with a per-channel/per-pattern color scheme that is simpler to control.   I know it's hard to run in real-time, but there are some really cool looks we can get w/contrasting per-pattern colors.

jkbelcher:
> It may be that we can redesign our palettes and controls a little and come up with a per-channel/per-pattern color scheme that is simpler to control. I know it's hard to run in real-time, but there are some really cool looks we can get w/contrasting per-pattern colors.

Agreed, there could be another good solution.

PR #407, opened 2023-10-14
Add Random Strobe effect

zranger1:
This variant of the "normal" Chromatik strobe effect strobes panels and edges at the specified rate, but with each element offset by a random amount of time.  It is set up for control by a single DMX channel driving the speed control.  Short video below.

Note:  I couldn't figure out how to make it work with Chromatik's new view system -- it uses panel/edge information from TEWholeModel that doesn't seem to be available in Chromatik views.  There is view switching code in RandomStrobeEffect.java, but it's commented out for the moment, so we can fix it later, or if there's something simple I missed, we can put it in now!

https://www.youtube.com/watch?v=xDfLmDJFJpw


andrewlook:
Awesome!

Just edited the target branch of this PR to point to "dev" branch, since that's where the other PRs have been merged in.

IMHO doing random strobe at TEModel level makes more sense than using views, just aesthetically I think that'll look cooler. Also given that our concept of views is likely to be in flux a bit, can't hurt to not have another thing entangled there :)

andrewlook (approved):
Tested it out and it's working for me! Merging so we can integrate it into the show file

PR #408, opened 2023-10-15
Allow DmxColorModulator to force Fixed color mode

jkbelcher:
Adds a new option in DmxColorModulator, defaulting to On, that sets the target color to Fixed mode.  This keeps the palette color from cycling unexpectedly if it was set to Oscillate or Cycle modes.

image


zranger1 (approved):
Code changes look good.  Tested and confirmed that even if I deliberately set palette to cycle, DMX is able to set the correct color.

PR #410, opened 2023-10-15
DmxEngine bug fix: avoid blending channels that have never been run

jkbelcher:
Bug fix: If a file containing disabled channels was opened, and in Performance mode one of the disabled channels was aux-cued prior to being cued or enabled, the DmxEngine was crashing when attempting to look up unknown buffers.  This fix avoids the lookup for these channels; there is no need to blend if the channel has not run.

Both @andrewlook and @jvyduna have seen this error recently.

Also:

- Commented out beacon 1 in the config file by default, so we're not sending packets to a non-existent IP.
- Removed profiler nanoseconds tracking, that should be reserved for the normal mixer.


zranger1 (approved):
Code changes are simple and look fine.   Tested:
- confirmed that I can reproduce the bug on the current dev branch with the setup described
- confirmed with the same "broken" .lxp file, that the error did not occur on this branch

Approved!

PR #411, opened 2023-10-15
Update midi surfaces and Dev Switch

jkbelcher:
The commit messages describe all the changes.

Dev Switch now looks like:
image

Minor fixes to the GigglePixel UI:
image

MidiFighterTwister UI should now show the Aux & Bank settings.


zranger1 (approved):
Code changes are simple and logical.  Tested and confirmed that UI changes are as advertised.   Approved.  

(Would suggest that someone at San Carlos run this w/all the relevant midi hardware before merging though.)

jkbelcher:
I tested w/ bomebox, apc40, and 2 midiFighterTwisters at home.  I think the rack may be packed up so going ahead an merging.

PR #413, opened 2023-10-17
DmxRangeModulator

jkbelcher:
Although not immediate required, this is a common type of DMX parameter and therefore could be useful for field edits.

It maps a subset of DMX values within a channel to its range output.


zranger1 (approved):
Tested & working.  Another nice DMX feature!

PR #415, opened 2023-10-18
Add TE views to global

jkbelcher:
Load TE view definitions into the new global Model>Views so they can be used across project files.

### DIFF
--- a/src/main/java/titanicsend/model/justin/ViewCentral.java
+++ b/src/main/java/titanicsend/model/justin/ViewCentral.java
@@ -34,6 +34,7 @@ import heronarts.lx.model.LXModel;
 import heronarts.lx.model.LXView;
 import heronarts.lx.parameter.LXParameterListener;
 import heronarts.lx.parameter.ObjectParameter;
+import heronarts.lx.structure.view.LXViewDefinition;
 import heronarts.lx.studio.LXStudio;
 import heronarts.lx.studio.TEApp;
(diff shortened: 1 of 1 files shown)


andrewlook (approved):
Awesome, thanks for the quick turnaround! I was trying to figure out how to do this but didn't know about `lx.structure` (I was looking for `lx.views` or `lx.model` so you saved me an awful lot of time navigating the LX API docs - thanks!)

zranger1 (approved):
This is good - simple, functional and preserves our texture mapping "correction" for the car ends!

andrewlook:
@jkbelcher I went ahead and tested this. The `Edge` view does exactly what we need it to, and when I set it on a channel it persists across save/load, which solves our immediate problem so I'll merge this right away.

A couple notes that don't need to be fixed immediately, but just curious to hear what you think about them:
1. The list of views has duplicates of each name with a semicolon.. ex. `Edges` and `Edges;` (I think in the TE views, "Panels;" behaves differently from "Panels" and looks more like "Fixtures", but I'm not sure what the correct behavior there should be).
image

2. `Panel Sections` seems to behave the same as `Panels` (one big coordinate space)
3. `Fixtures` seems to behave the same as `Panels`

ex. here's what it looks like using LX model views of `Panel Sections`: 

image

vs. when I use the TE views for panel sections:

image

or TE view of fixtures:

image

PR #416, opened 2023-10-18
Kaliedosonic Pattern - Music Reactive Kaleidoscope

zranger1:
This actually turned out well, music reactivity and all!  Take it for this weekend if you've got time, or leave for next gig otherwise. 
Now with full palette compliance.  Works best in the "Panel Sections" view.

This is one of the weirder things I've done.  It converts the audio FFT into an animated 2D density field with "blobs" corresponding to energy in each band, then renders them in smooth radial slices.  You can turn the number of slices down to 1 to get a better picture of what's actually going on.

- Speed: Yes
- X/Y Offset: Yes
- Spin/Angle: Yes
- Brightness: Yes
- Size: Overall scale
- Quantity: Number of kaleidoscope "slices", from 1 to 13
- Wow1: "Blast radius" (Size of blobs in density field!)
- Wow2:  Audio reactivity fine tuning ("Volume", more-or-less)
- WowTrigger: No

### DIFF
--- a/src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
+++ b/src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
@@ -14,7 +14,7 @@ public class Kaleidosonic extends TEPerformancePattern {
     NativeShader shader;
 
     public Kaleidosonic(LX lx) {
-        super(lx, TEShaderView.DOUBLE_LARGE);
+        super(lx, TEShaderView.ALL_POINTS);
 
         controls.setRange(TEControlTag.WOW1, 0.04, 0, 0.08);  // bass response
         controls.setRange(TEControlTag.WOW2, 0.5, 0.2, 3);    // overall audio level adjustment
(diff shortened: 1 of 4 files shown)


src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
```diff
@@ -0,0 +1,51 @@
...
+import titanicsend.pattern.yoffa.framework.TEShaderView;
+import titanicsend.pattern.yoffa.shader_engine.NativeShader;
+
+@LXCategory("Native Shaders Panels")
+public class Kaleidosonic extends TEPerformancePattern {
+    NativeShaderPatternEffect effect;
+    NativeShader shader;
+
+    public Kaleidosonic(LX lx) {
+        super(lx, TEShaderView.ALL_POINTS);
```
andrewlook, line 17: FYI I changed this to default to full view

andrewlook (approved):
@zranger1 this pattern is SO cool! thanks for knocking this one out!  🙌

PR #417, opened 2023-10-18
Jeff/startup mvn

titanicsend:
Check for maven build in startup script, and build if not present.


andrewlook:
modified PR to point at dev as base branch

andrewlook (approved):
discussed with @jvyduna - we'll take some manual steps the morning before the show to ensure that:
- the right jars are built
- we have backup copies of the jar accessible
- we have backup copies of the project file(s) accessible

PR #418, opened 2023-10-19
Move AutoVJ to last channel

jkbelcher:
Site team wants autoVJ to be on the last channel, not channel 8.  Commented out existing code in case someone wants to undo the change later.


andrewlook (approved):
Verified that before this fix, clicking "enable autoVJ" splices all the autoVJ channels in after channel 7, displacing channel 8 to the end of the list.

Also verified that after this change, autoVJ channels are appended to the very end of the list 

👍

PR #414, opened 2023-10-18
Update config for fore/aft X modules

fakeplant:
Updated edge and panel routing config based on hardware changes made before Framework event.

Example new routing below:
Screenshot 2023-10-17 at 7 23 37 PM

PR #419, opened 2023-10-19
(main->dev) Update config for fore/aft X modules (#414)

andrewlook:
updating dev with #414


jkbelcher:
@andrewlook This looks like simple config changes.  Do you want to revolve the conflicts so we can merge it?

andrewlook:
@jkbelcher this was an attempt to merge `main` into `dev` before Oct 21 show. the changes here made it into dev before the framework show where this in #424 (the commit from #414 was included in #424)

I think what we might want to do instead is go the other direction and merge `dev` back into `main`, so I'll close this PR for now

PR #422, opened 2023-10-20
Showfile updates for Oct21: "High" and "Low" energy configurations for all patterns in unfiltered view

andrewlook:
## Goal

how do we give VJ's something quick to go to if next pattern not ready?
- We want high-energy patterns ready for big drops, lower-energy patterns for bridges/breaks.
- Ex. Take spin, set to zero. Unfortunately speed control is ceded via DMX this show.

Proposed Solution:
- Use Unfiltered channels 1,2,5 for high-driving EDM vibes, use channel 6 for the same set of patterns, all configured for low-energy

Process:
- [x] duplicate all patterns.
- [x] Put "+/-" Prefix for high-energy versions.
- [x] Iterate through, configure high-energy params on the "bang" patterns.

Later,
- [ ] create a new project file (so that 1 big list of patterns as high-energy and low-energy is preserved somewhere for reference)
- [ ] duplicate c1 to channel 6, remove high-energy
- [ ] in c1, remove all low-energy pats, then duplicate to chans 2,5

## Smaller Changes in this PR:

Patterns:
- [x] Add RainNoise
- [x] Add Kaleidosonic

Modulators/DMX: Whenever possible, link enable/disable so that turning off modulator activates LX right away
- [x] Speed modulator enable - link to Global Speed toggle 

### show file pattern selection

- high / low energy versions of edge patterns?

- experimenting with "clips" as a way to save pattern presets, where more variations are possible besides what we have for high/low. Candidate patterns where I think we could get more variability / mileage by saving additional clips:
  - Phasers
  - Kaleidosonic
  - RadialSimplex (also looks great on edge mask, when modulators applied)
  - Rainbands
  - SigmoidDance (varying wow1, size, quantity)
  - TriangleCross (note: hooking "quantity" param up to modulator looks sick)
  - TriangleInfinity (varying quantity, size, etc)

### show file DMX integration

DMX Grid Modulator:
- [ ] Look at GrandMA.lxp
- [ ] Add channel w/ DMX Grid Pattern
- [ ] Add DMX Grid Modulator 
- [ ] Connect them


zranger1 (approved):
This is a great idea!   Our patterns do a whole range of wildly different things depending on parameterization.  This gives VJs instant access to some of that power in a simple, logical way.  (Just don't forget to explain to them that they still have to adjust things themselves for best effect, even with the cool presets!)

On clips:  if you do them, do them as late in your dev process as possible -- like the very last change you make to the config file.  In my experience, they are *extremely* fragile and can break if their setup is perturbed in the smallest way.

PR #426, opened 2023-11-08
New Shader Preprocessor

zranger1:
Adds new shader preprocessor, plus a little code cleanup in advance of the larger refactor that will arrive with the memory management update.   The new preprocessor allows you to create and edit patterns entirely in GLSL, without involving Java at all.

### Supported Features:
	
Basic #include files (handles nested includes, up to 10 levels)
```
	#include "resources/shaders/library/file.fs"
	  ...or...
	#include <library/file.fs>  // prefixes with default resource path
```
	
New shader configuration #pragmas  (#pragma is used b/c GLSL implementations are required to ignore pragmas they don't understand)
	
	// use the automatic class creator for ths shader. Recommended, but only required if you
	// use no other configuration pragmas. 
	#pragma auto 
		
	// set the name of the pattern's java class (and the pattern name in the UI) If not specified
	// the name of the shader file (not including .fs) will be used.  If the specified class exists
	// a new class will not be created for the shader.)
	#pragma Name("ReallyCoolPattern");        // must be unique, and a valid Java class name

	// Set the shader's pattern browser category
	#pragma LXCategory("Best Shaders Ever!)
	
	// configure common controls at setup time.  Control tags are the same as the
    // TEControlTag values in Java. 
	#pragma TEControl.SPEED.Value(1.0)       // setValue()
	#pragma TEControl.WOW1.Range(1,0,5)     // setRange() 
	#pragma TEControl.SCALE.Label("Timmy")   // setLabel()
	#pragma TeControl.SPEED.Exponent(2.25)   // setExponent()
	#pragma TeControl.QUANTITY.NormalizationCurve(REVERSE,NORMAL,BIAS_CENTER,BIAS_OUTER)  // setNormalizationCurve()
	#pragma TEControl.SPEED.Disable	         // markUnused() - hides control in UI
	
	// specify texture channels and files directly in GLSL
	#pragma iChannel1 "shaders/resources/textures/test.png"
	... or ...
	#pragma iChannel1 <textures/test.png>	
	
	// choose how the pattern interacts with x/y translation controls
	#pragma TEControl.TranslateMode(DRIFT,NORMAL)

### Notes:
- Line number for compile errors in shader code is now (almost always) correct.  Yay! :fireworks:
- Framework changes will cause all shaders to need a recompile the first time you load it, so initial startup may
be slow.   Subsequent startups will return to normal speed.
- If you edit and reload a pattern, pattern name and LXCategory won't be updated until the next time you start Chromatik. 
[...]

PR #428, opened 2023-11-14
Refactor: break inner classes out of TEPerformancePattern

andrewlook:
Since there's 2 PR's affecting TEPerformancePattern out, I stacked this PR on top of #427, but can wait until #426 merges first and redo this work as necessary - I mainly wanted to see what the constraints were in terms of how the relationships between classes had to change.

These changes have been lightly tested, but everything seems to compile and run well so far.

### TECommonControls

needs access to:
- `TEPerformancePattern.addParam / removeParam` (LX `add/removeParameter` are protected methods, so I added package-protected passthrough methods on `TEPerformancePattern`)
- `TEPerformancePattern.setCustomRemoteControls`
- `TEPerformancePattern.view`

the best way to accomplish this seemed to be to just store a backreference to `TEPerformancePattern` at the class level, rather than pass it in to a few of the member methods.

Making package-protected `add/removeParam` that just passthrough to `add/removeParameter` felt like a hack, but since it's package protected it shouldn't be usable outside of these base pattern classes and related helper classes.

### TECommonAngleParameter

needs access to:
- `TEPerformancePattern.spinRotor`
- `TEPerformancePattern.getSpin()`

it might have been overkill to pass in a reference to `TEPerformancePattern` - I was hoping I could get away with just passing in a reference to `spinRotor`, but the `reset()` method needed access to `getSpin()`.

I could imagine a solution where the spin parameter and rotor are both instantiated within this class and then passed to TEPerformancePattern / TECommonControls as needed. Or, given that there's a bunch of state and logic related to spin/angle in TEPerformancePattern, maybe one class could instantiate and manage the various pieces of state - I'm very open to better ways of doing this. For now though, this seemed to me like the least invasive path to getting things working with the class split out.

### TEColorParameter

needs access to:
- `TEPattern.primaryGradient/secondaryGradient/foregroundGradient/paletteGradient`

I was less sure how we want to proceed here, given that calcColor/etc inherit from the global palette (good) and can be overriden by the `gradient` params (which I think we agreed we don't need).

I wasn't quite sure how best to de-tangle these concepts. It seems like having a `color` parameter is handy (and the offset picker could be nice to have for pattern development if we move it out of the top 16 params, the way @jkbelcher did for `view` in #427 - so for now I passed a reference to the parent pattern into `TEColorParameter` so that we could access `primaryGradient`/etc.


andrewlook:
closing in favor of #429 + #430

PR #429, opened 2023-11-16
Refactor(TEPerformancePattern): part 1 - preserve git history within split-out classes

andrewlook:
We realized that the per-line git history got lost when I first did the TEPerformancePattern (TEPP) split.

This will be a 2-part PR as a rehash of #428 :
1. (this PR) very rough chop of the lines of code into the right files, with no fixes to imports/etc but preserving git history
2. (subsequent PR to be stacked on this one) all of the individual fixes to actually update the code, so we can review them in a cleaner context.

---

The result is that the split-out files preserve the git history via `git annotate`/`git blame`:

image

Note that the above screenshot is when the text hasn't had its indentation updated. When I reindent, the latest committer will show up as my name, but importantly the git history for each of these files will preserve the past commits affecting this code so we can still navigate to when and where those changes happened.

---

Based on Sina's suggestion, copying the file within git (and then reverting the deletion of the original file, then adding a merge commit) is enough to trick git into preserving the history. I found a [stackoverflow answer](https://stackoverflow.com/questions/3887736/keep-git-history-when-splitting-a-file) that provided a little bash script that I used, `git-split.sh`:

```sh
#!/bin/sh

if [[ $# -ne 2 ]] ; then
  echo "Usage: git-split.sh original copy"
  exit 0
fi

git mv $1 $2
git commit -n -m "Split history $1 to $2"
REV=`git rev-parse HEAD`
git reset --hard HEAD^
git mv $1 temp
git commit -n -m "Split history $1 to $2"
git merge $REV
git commit -a -n -m "Split history $1 to $2"
git mv temp $1
git commit -n -m "Split history $1 to $2"
```


src/main/java/titanicsend/pattern/Rotor.java
```diff
@@ -1,91 +1,66 @@
...
-        }
+    /**
+     * Sets maximum spin rate for all patterns using this rotor.  Note that a Rotor
+     * object is associated with a timer, which can be a VariableSpeedTimer.  So
+     * "seconds" may be variable in duration, and can be positive or negative.
+     *
+     * @param radiansPerSecond
+     */
+    void setMaxSpinRate(double radiansPerSecond) {
+                                               maxSpinRate = radiansPerSecond;
```
jkbelcher, line 63: Why all these added spaces?  Lots of these in this commit.

jkbelcher (approved):

jkbelcher:
As discussed on last nights call, we are planning a big formatting sweep for the repo.  The discussion for whitespace formatting is pushed back to a later date, and this PR was merged to keep things moving.

PR #430, opened 2023-11-16
Refactor(TEPerformancePattern): part 2 - making inner classes work after extraction

andrewlook:
see part 1 ( #429 ) for how git history was migrated.

summary of actual code changes:

### TECommonControls

needs access to:
- `TEPerformancePattern.addParam / removeParam` (LX `add/removeParameter` are protected methods, so I added package-protected passthrough methods on `TEPerformancePattern`)
- `TEPerformancePattern.setCustomRemoteControls`
- `TEPerformancePattern.view`

the best way to accomplish this seemed to be to just store a backreference to `TEPerformancePattern` at the class level, rather than pass it in to a few of the member methods.

Making package-protected `add/removeParam` that just passthrough to `add/removeParameter` felt like a hack, but since it's package protected it shouldn't be usable outside of these base pattern classes and related helper classes.

### TECommonAngleParameter

needs access to:
- `TEPerformancePattern.spinRotor`
- `TEPerformancePattern.getSpin()`

it might have been overkill to pass in a reference to `TEPerformancePattern` - I was hoping I could get away with just passing in a reference to `spinRotor`, but the `reset()` method needed access to `getSpin()`.

I could imagine a solution where the spin parameter and rotor are both instantiated within this class and then passed to TEPerformancePattern / TECommonControls as needed. Or, given that there's a bunch of state and logic related to spin/angle in TEPerformancePattern, maybe one class could instantiate and manage the various pieces of state - I'm very open to better ways of doing this. For now though, this seemed to me like the least invasive path to getting things working with the class split out.

### TEColorParameter

needs access to:
- `TEPattern.primaryGradient/secondaryGradient/foregroundGradient/paletteGradient`

I was less sure how we want to proceed here, given that calcColor/etc inherit from the global palette (good) and can be overriden by the `gradient` params (which I think we agreed we don't need).

I wasn't quite sure how best to de-tangle these concepts. It seems like having a `color` parameter is handy (and the offset picker could be nice to have for pattern development if we move it out of the top 16 params, the way @jkbelcher did for `view` in #427 - so for now I passed a reference to the parent pattern into `TEColorParameter` so that we could access `primaryGradient`/etc.


src/main/java/titanicsend/pattern/TEColorParameter.java
```diff
@@ -21,39 +19,39 @@ public enum SolidColorSource {
...
+                    return this;
+                }
             }
-        }
-            .setDescription("For a solid color: Whether to use global TE palette (preferred), or a static color unique to this pattern");
+                    .setDescription("For a solid color: Whether to use global TE palette (preferred), or a static color unique to this pattern");
 
     public final CompoundParameter color2offset =
-        new CompoundParameter("C2Offset", 0.5);
+            new CompoundParameter("C2Offset", 0.5);
```
jkbelcher, line 46: Why these added spaces?

src/main/java/titanicsend/pattern/TEColorParameter.java
```diff
@@ -21,39 +19,39 @@ public enum SolidColorSource {
...
 
     // GRADIENT
 
     @SuppressWarnings("unchecked")
-    public final EnumParameter<TEGradient> gradient = (EnumParameter<TEGradient>)
-        new EnumParameter<TEGradient>("Gradient", TEGradient.FULL_PALETTE)
-            .setDescription("Which TEGradient to use. Full_Palette=entire, Foreground=Primary-Secondary, Primary=Primary-BackgroundPrimary, Secondary=Secondary-BackgroundSecondary")
-            .setWrappable(false);
+    public final EnumParameter<TEPattern.TEGradient> gradient = (EnumParameter<TEPattern.TEGradient>)
+            new EnumParameter<TEPattern.TEGradient>("Gradient", TEPattern.TEGradient.FULL_PALETTE)
```
jkbelcher, line 52: ?

src/main/java/titanicsend/pattern/TEColorParameter.java
```diff
@@ -66,16 +64,16 @@ public enum BlendMode {
...
 
     public final EnumParameter<BlendMode> blendMode =
-        new EnumParameter<BlendMode>("BlendMode", BLEND_MODE_DEFAULT) {
-            @Override
-            public LXParameter reset() {
-                // JKB: Don't worry about this, just avoiding a minor bug
-                // in EnumParameter. It'll be fixed soon.
-                setValue(BLEND_MODE_DEFAULT);
-                return this;
+            new EnumParameter<BlendMode>("BlendMode", BLEND_MODE_DEFAULT) {
```
jkbelcher, line 67: ?

src/main/java/titanicsend/pattern/TEColorParameter.java
```diff
@@ -170,13 +171,13 @@ public int calcColor() {
...
                 // TODO: scale brightness here
-                return _getGradientColor(getOffsetf(), TEGradient.FOREGROUND);
+                return _getGradientColor(getOffsetf(), TEPattern.TEGradient.FOREGROUND);
             default:
             case STATIC:
                 return LXColor.hsb(
-                    this.hue.getValue(),
-                    this.saturation.getValue(),
-                    this.brightness.getValue()
+                        this.hue.getValue(),
```
jkbelcher, line 178: ?

src/main/java/titanicsend/pattern/TEColorParameter.java
andrewlook: This one surprised me too - apparently inside `TEPerformancePattern` it was okay to use an inner class of a parent class without qualifying it.

Personally, I like having qualified imports for inner classes used outside of their parent class, so it's more obvious where to find them - I generally only do static imports for well-known classes where it's convenient to refer to them in shorthand (e.g. `Map.Entry<K, V>` is pretty clunky relative to how often it needs to be used).

I don't see any reason that `TEGradient` needs to be inside `TEPattern` though - seems like a distinct concept. Should I extract that one into its own class?

src/main/java/titanicsend/pattern/TECommonAngleParameter.java
```diff
@@ -3,15 +3,20 @@
...
 class TECommonAngleParameter extends CompoundParameter {
 
-    public TECommonAngleParameter(String label, double value, double v0, double v1) {
+    private final TEPerformancePattern parentPattern;
+    private final Rotor spinRotor;
+
+    public TECommonAngleParameter(TEPerformancePattern parentPattern, Rotor rotor, String label, double value, double v0, double v1) {
         super(label, value, v0, v1);
+        this.parentPattern = parentPattern;
+        this.spinRotor = rotor;
```
andrewlook, line 19: we could do something like `this.spinRotor = parentPattern.spinRotor` and avoid passing an extra constructor arg. It seems like that hides the real dependency between these classes though, I wish we could just pass in `spinRotor` and have access to the current value of the `spin` parameter (maybe some class encompassing rotor + parameter would be a nice abstraction).

That being said, I don't love the way this code looks now, and would be open to simplifying if anyone has a preference.

jkbelcher:
Made a bunch of changes.  They should be clear to follow if you look at each commit individually.

src/main/java/titanicsend/color/TEGradientSource.java
```diff
@@ -0,0 +1,38 @@
...
+    LXGradientUtils.ColorStops colorStops = new LXGradientUtils.ColorStops();
+    colorStops.setNumStops(numStops);
+    return colorStops;
+  }
+
+  // Whole palette gradient across all 5 stops. Usually starts and ends with black.
+  public LXGradientUtils.ColorStops paletteGradient = initColorStops(5);     // [X] [X] [X] [X] [X]  All five color entries
+  public LXGradientUtils.ColorStops primaryGradient = initColorStops(3);     // [X] [ ] [X] [ ] [ ]  Background primary -> Primary
+  public LXGradientUtils.ColorStops secondaryGradient = initColorStops(3);   // [ ] [ ] [ ] [X] [X]  Background secondary -> Secondary
+  public LXGradientUtils.ColorStops foregroundGradient = initColorStops(3);  // [ ] [ ] [X] [X] [ ]  Primary -> Secondary
```
andrewlook, line 18: thanks for the explanator comment here! 🙏  I'd never fully asked about or understood this layout

andrewlook:
@jkbelcher I reviewed all of your changes and they're looking great! Thanks for extracting the `color` package and the stuff in `utils`. Also i appreciate the sanitized method order in `TECommonControls`. I'm happy with this PR as it stands and am ready to merge it. Your `TODO(JKB)` comments make sense to me as little items to follow up on, but I agree that we can deal with those after this PR's merged 👍

PR #432, opened 2023-11-26
Mark unused controls in java patterns

sina-cb:
This is a short PR to mark unused parameters in the Java patterns.

I went through the code, and double checked the patterns in the UI to make sure the controls are not actually being used by the patterns.

To find which patterns to update, I found all places the "TEPerformancePattern.addCommonControls" and went through all the patterns that are not using shaders.

Also ignored these patterns:
 - Pixelblaze patterns and their sub-classes: the parent class is calling the addCommonControls and all children are inheriting all the controls automatically.
 - Constructed patterns and their sub-classes: Similar to the pixelblaze, the parent class call addCommonControls and all sub-classes are inheriting. But when looking at the code, these classes are using a deprecated fragmented shader class, and they actually have different parameters that are not showing up in the UI.

### DIFF
--- a/src/main/java/titanicsend/pattern/jon/ArcEdges.java
+++ b/src/main/java/titanicsend/pattern/jon/ArcEdges.java
@@ -24,6 +24,7 @@ public class ArcEdges extends TEPerformancePattern {
     public ArcEdges(LX lx) {
         super(lx, TEShaderView.ALL_POINTS);
 
+        // set up the controls used for this pattern.
         controls.setRange(TEControlTag.SIZE, 1, 5, 0.1);              // scale
         controls.setRange(TEControlTag.QUANTITY, 0.6, 0.72, 0.35);  // noise field position
         controls.setRange(TEControlTag.WOW1, 0.025, 0.001, 0.08);     // noise magnitude
--- a/src/main/java/titanicsend/pattern/jon/EdgeKITT.java
+++ b/src/main/java/titanicsend/pattern/jon/EdgeKITT.java
@@ -19,6 +19,15 @@ public class EdgeKITT extends TEPerformancePattern {
 
         controls.setRange(TEControlTag.SIZE, 0.5, 0.01, 1.5);
 
+        controls.markUnused(controls.getLXControl(TEControlTag.XPOS));
+        controls.markUnused(controls.getLXControl(TEControlTag.YPOS));
+        controls.markUnused(controls.getLXControl(TEControlTag.QUANTITY));
+        controls.markUnused(controls.getLXControl(TEControlTag.EXPLODE));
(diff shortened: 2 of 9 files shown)


src/main/java/titanicsend/pattern/jon/EdgeKITT.java
```diff
@@ -19,6 +19,15 @@ public EdgeKITT(LX lx) {
 
         controls.setRange(TEControlTag.SIZE, 0.5, 0.01, 1.5);
 
+        controls.markUnused(controls.getLXControl(TEControlTag.XPOS));
+        controls.markUnused(controls.getLXControl(TEControlTag.YPOS));
+        controls.markUnused(controls.getLXControl(TEControlTag.QUANTITY));
+        controls.markUnused(controls.getLXControl(TEControlTag.EXPLODE));
+        controls.markUnused(controls.getLXControl(TEControlTag.WOW1));
```
andrewlook, line 26: I was going to say that `wow1` is used in the `run()` method of this pattern, but it looks like that variable (`gradientMix`) is actually unused - maybe we should delete those 2 lines of code so it's consistent with `wow1` being unused?

```java
        // Wow1 controls the foreground vs. gradient color mix.  More Wow1 == more gradient
        double gradientMix = getWow1();
```

src/main/java/titanicsend/pattern/jon/ArcEdges.java
```diff
@@ -24,6 +24,7 @@ public class ArcEdges extends TEPerformancePattern {
     public ArcEdges(LX lx) {
         super(lx, TEShaderView.ALL_POINTS);
 
+        // set up the controls used for this pattern.
```
andrewlook, line 27: we can probably remove this comment so we don't touch this file in the git history, since we're not making any other changes here

andrewlook (commented):
@sina-cb this looks great!

I went through the patterns myself and as far as I can tell, all of the params you marked unused were correctly flagged. Thanks for that!

Just some context on 2 of these patterns:
- `Checkers` is mainly used for testing the lighting hardware and making sure that pixels are correctly mapped (since offset squares look funny if any pixels are misaligned).
- `ExampleDmxTEPerformancePattern` is an example pattern for controlling hardware outputs outside of our normal LED panels/edges (in this case, Justin wrote this so that we could control big spotlight beacons).

I think we should keep the controls marked unused on the above 2 patterns, but I wanted to mention them since they're not in our day-to-day lineup of patterns that get used during a show.

All of the other ones you touched, though, are ones we actually used at our LA show and I remember being frustrated by 1 or 2 of them not yet having the controls marked unused! So this is a great first PR - thanks again!

andrewlook (approved):

PR #439, opened 2023-12-05
reformat: google (using spotless)

andrewlook:
I made the commit come from @titanicsend user so that my name won't be everywhere in the git history.

Checked the git history and it appears that intellij actually skips the whitespace changes and has accurate author history, which is nice:

image


jkbelcher:
This is looking good to me.

PR #441, opened 2023-12-12
fix: pixelblaze reference to TEColorType

andrewlook:
Noticed some error messages in the logs where pixelblaze was looking for a now-nonexistent class (ColorType, renamed to TEColorType in #430 )


jkbelcher (approved):
Didn't test but it looks right.

PR #431, opened 2023-11-21
Update README.md

jvyduna:

PR #442, opened 2023-12-12
Prep for merge: Merge 'main' into 'dev' (before bulk reformatting)

andrewlook:
Before we merge `dev` into `main` and delete `dev` branch, there were some merge conflicts I wanted to resolve:

## java changes

The following changes were made to `main` but not yet ported to dev:
- #431 
- #432

I'm merging them into main, so we can apply reformatting after integrating these changes.

## edge/panel configs

Also, the following change was made on `dev`:
- #424 
Which conflicted with these changes to `main`:
- #414

I discussed with Eli (@fakeplant ) and he agreed that `dev` was the most recent set of config changes to edge/panels configs, since that was what we used for the October framework show. When resolving the merge conflict, I took the changes from #424 to be the true ones (though I'm not sure how to verify this apart from asking Eli).

Side note (outside scope of this PR): I did notice that one of the skirt panels seems to be dark in LX with this config, but i think it's been that way on dev for a while (and was hidden by speakers for the LA show so didn't cause a problem).

Screen Shot 2023-12-12 at 3 55 46 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,25 @@
 Titanic's End LXStudio
 ==
 
-Titanic's End is a mutant vehicle designed for Burning Man 2022. (and beyond!)
+Titanic's End is a mutant vehicle that debuted at Burning Man 2022 and has since participated in EDC and Framework events.
 
-This repo covers the code for the design and instrumentation of the many LEDS wrapped around this mutant vehicle.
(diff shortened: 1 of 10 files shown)


jkbelcher:
FYI this is showing zero files changed

jkbelcher:
Approved

PR #444, opened 2023-12-13
Merge dev into main

titanicsend:
After this, we'll delete dev branch and work off main (with tag releases as necessary)


jkbelcher (approved):
Tested, works with no errors.  Merging.

PR #447, opened 2024-02-12
Shader System Refactor + NDI Video Support

zranger1:
### Rewrite/refactor of OpenGL shader rendering engine for:

- Improved performance
- More efficient memory management
- Higher quality of life for pattern authors
- Access to backbuffer for postprocessing & feedback effects
- Multipass rendering (multi-shader patterns and effects)
- Memory efficiency and stability.
- Streaming video (NDI) input/output 

The idea is that you should have freedom to load whatever you need when building a project file for a show. So...

On a reasonably powered (say, 16GB M1 or higher) system, we should now be able to load several hundred shaders in a project, rapidly click through all those shaders to activate them, unload and reload the project multiple times. The system shouldn't run out of memory or otherwise complain or crash.

The shaderstress.lxp project is set up with 9 channels of 78 patterns for just this sort of testing. Each channel is set up to switch between patterns every 200ms. If you enable Auto-Cycle on all the channels, it will go through quite a lot of patterns, very quickly.

### Refactoring
The shader system has been largely rewritten for clarity and simplicity.

Every shader pattern in the system has been modified to use the GLShaderPattern class as a base (instead of NativeShaderPatternEffect). This should make it easier to write and maintain shader patterns, as well as adding support for new features that increase power and flexibility. It should not affect current functionality in any way.

**_But... I touched almost everything._**

So please check out your patterns and make sure I didn't break anything!

### iBackbuffer and Multipass Rendering
Patterns can now run multiple shaders, with the output of each shader available to the next one as the Sampler2D uniform, iBackbuffer.   (In the single shader case, iBackbuffer contains the output from the previous frame.)

Check out the [Fireflies](https://github.com/titanicsend/LXStudio-TE/blob/jon/NativeNDI/src/main/java/titanicsend/pattern/jon/Fireflies.java) and [MultipassDemo](https://github.com/titanicsend/LXStudio-TE/blob/jon/NativeNDI/src/main/java/titanicsend/pattern/jon/MultipassDemo.java) patterns to see how this works.  It also powers NDI video and shader-based effects, described below.

### NDI Video support
Added support for receiving up to 10 channels of NDI video.  Incoming video is automatically scaled and color converted for TE. But keep your input resolutions as low as possible. 640x480 is ideal.  Streaming 4k video to the car will be... non-fast.

See the NDIReceiverTest  pattern (pattern/jon/NDIRecieverTest.java) for details.  There are also new tools ([NDIRawOutEffect ](https://github.com/titanicsend/LXStudio-TE/blob/jon/NativeNDI/src/main/java/titanicsend/effect/NDIOutRawEffect.java)and the [ModelFileWriter pattern](https://github.com/titanicsend/LXStudio-TE/blob/jon/NativeNDI/src/main/java/titanicsend/pattern/jon/ModelFileWriter.java)) to enable NDI streaming of pixel data for projection, AI model training and other uses.

### GPU-Powered Effects
Added support for using shaders in Chromatik effects. The ["Explode"](https://github.com/titanicsend/LXStudio-TE/blob/jon/NativeNDI/src/main/java/titanicsend/effect/ExplodeEffect.java) effect is the first implementation.  It is now a normal effect. It works with all patterns, and can be applied per pattern, per channel, or to the master channel.

Accordingly, "Explode" has been removed from the common controls.

### Documentation
The [shader framework documentation](https://github.com/titanicsend/LXStudio-TE/blob/jon/NativeNDI/resources/docs/ShaderFramework.md) has been updated with details on the new features, plus several smaller quality-of-life features for pattern writers.

### New Feature Demo Videos

Video Input
https://www.youtube.com/watch?v=LJIvGhG4tio

Video Output (to Projector)
https://www.youtube.com/watch?v=z6oDxfDJz5s

GPU Effects
https://www.youtube.com/watch?v=zXAzV0qR5-4


src/main/java/titanicsend/effect/ExplodeEffect.java
```diff
@@ -0,0 +1,181 @@
...
+  private double getBasis() {
+    double r;
+    double gate = 1.0;
+
+    if (tempoSync.isOn()) {
+      r = this.lx.engine.tempo.getBasis(this.tempoDivision.getEnum());
+      r = (r + this.tempoPhaseOffset.getValue()) % 1.;
+    } else {
+      r = basis.getBasis();
+    }
```
andrewlook, line 104: awesome! thanks for adding this configuration to the explode effect's tempo behavior, that was a big wishlist item from @jvyduna and I after the last show! 🙏

src/main/java/titanicsend/pattern/TECommonControls.java
```diff
@@ -325,20 +320,13 @@ public void addCommonControls(TEPerformancePattern pat) {
...
       }
 
-      // disable explode control if the pattern isn't a shader
-      // (takes advantage of the fact that all shaders have a missingControls file entry,
-      // even if they're not missing any controls.)
-      // TODO - reminder to implement "explode" as an effect so all patterns can use it.
-      else if (param.getLabel().equals(TEControlTag.EXPLODE.getLabel())) {
-        markUnused(param);
-      }
-
```
andrewlook, line 341: niiice! so glad we reclaimed this parameter :)

src/main/java/titanicsend/pattern/glengine/ConstructedShaderPattern.java
```diff
@@ -0,0 +1,65 @@
...
+
+     Note that you can call addShader() multiple times to add multiple shaders
+     to the pattern.  Shaders will run in the order they are added, with
+     the output of each shader available to the next shader in line via the
+     iBackbuffer 2D sampler uniform.
+
+     Each shader can have its own onFrame() setup function, or you can
+     use the default function (which does nothing) by calling
+     addShader(shader) instead of addShader(shader, setupFunction).
+   */
```
andrewlook, line 62: excellent documentation, thanks for this!

Also I'm already relieved at seeing how much more readable the code is using this abstraction vs. `createEffects()`

src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
```diff
@@ -0,0 +1,139 @@
...
+    protected GLShader shader;
+    protected GLShaderFrameSetup setup;
+
+    public ShaderInfo(GLShader shader, GLShaderFrameSetup setup) {
+      this.shader = shader;
+      this.setup = setup;
+    }
+  }
+
+  protected GLEffectControl controlData;
```
andrewlook, line 30: just curious: do you imagine effect authors will need to extend GLEffectControl if they want to add parameters/controls? I wasn't immediately obvious to me from reading GLEffectControl whether it was meant to be subclassed itself, or called by subclasses of GLShaderEffect by any effects wishing to customize parameters

andrewlook: After re-reading ExplodeEffect as an example, it seems like it's calling `addParameter` on its controls, which gets passed up through `GLShaderEffect` to `TEEffect`, and it seems like effects need to pass their parameters directly to the shader as uniforms if they're needed there.

So are the implementors of GLControlData (GLPatternControl / GLEffectControl) meant just for any intrinsic uniforms that should be available by default, and unrelated to UI controls / "parameters"? And if pattern/effect authors need additional uniforms, they can define their own parameters and pass their own uniforms but they won't go through these GLControlData objects, so there's no need to extend them?

Sorry if this is a roundabout question, just want to make sure I understand how the pieces fit together as I'm reading through this code :)

src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
```diff
@@ -0,0 +1,139 @@
...
+  // Add a shader by fragment shader filename, using the default OnFrame() function.
+  // The simple option for shaders that use only the default TEPerformancePattern
+  // uniforms and don't require any additional computation in Java.
+  public void addShader(String shaderName,      String... textureFilenames) {
+    addShader(new GLShader(lx, shaderName, getControlData(), textureFilenames));
+  }
+
+  // Add a shader by fragment shader filename, with an OnFrame() function.
+  public void addShader(String shaderName, GLShaderFrameSetup setup) {
+    addShader(new GLShader(lx, shaderName, controlData), setup);
```
andrewlook, line 78: minor nit: any reason to use `controlData` here rather than `getControlData()` as on line 73?

src/main/java/titanicsend/pattern/glengine/TEAutoShaderPattern.java
```diff
@@ -37,24 +33,4 @@ protected String getShaderFile() {
...
-  public void onActive() {
-    super.onActive();
-    effect.onActive();
-    shader = effect.getNativeShader();
-  }
-
-  @Override
-  public void dispose() {
-    // do nothing for now.
-  }
```
andrewlook, line 59: nice!! this code cleanup is awesome, thanks so much for doing this 🙌

src/main/java/titanicsend/pattern/jon/ArcEdges.java
```diff
@@ -46,6 +42,21 @@ public ArcEdges(LX lx) {
...
+
+    // add the OpenGL shader and its frame-time setup function
+    addShader("arcedges.fs",
+      new GLShaderFrameSetup() {
+        @Override
+        public void OnFrame(GLShader s) {
+          // Here, we update line segment geometry
+          // Shader uniforms associated with a context stay resident
+          // on the GPU,so we only need to set them when something changes.
+          if (updateGeometry) {
```
andrewlook, line 54: if I read this correctly, we're only sending these uniform updates when the shader boots up, correct? Just wanted to sanity-check whether there's any other situations where something changes that would make us update the geometry.

src/main/java/titanicsend/pattern/jon/EdgeFall.java
```diff
@@ -25,6 +21,42 @@ public class EdgeFall extends TEPerformancePattern {
...
+
+      glowLevel = (float) getSize();
+
+      // tiny state machine for falling vs. resting states
+      if (isFalling) {
+        // simulate short explosive burst (by greatly increasing line
+        // width/glow) when event is first triggered
+        if (elapsedTime < burstDuration) {
+          glowLevel *= (float) (elapsedTime / burstDuration);
+        }
```
andrewlook, line 41: looks great! unrelated to this PR, but one feature request re: this pattern that we had from the LA show was for the lines to fall back into place at some point after hitting wowTrigger (since the behavior right now is that the lines stay gone and the pattern goes black until wowtrigger is hit again. Just remembered to mention this after coming across this pattern, no need to hold up this particular PR for the change. Or if you have any ideas how it might work I could take a stab at it too :)

src/main/java/titanicsend/pattern/jon/Fireflies.java
```diff
@@ -0,0 +1,38 @@
...
+    addCommonControls();
+
+    addShader(
+        "fireflies.fs",
+        new GLShaderFrameSetup() {
+          @Override
+          public void OnFrame(GLShader s) {
+            s.setUniform("iQuantity", (float) getQuantity());
+            s.setUniform("iSize", (float) getSize());
+            s.setUniform("iWow2", (float) getWow2());
```
andrewlook, line 34: Do these uniforms get set via `GLPatternControl`? Per my earlier question about GLControlData, just hoping to test my understanding of the relationship between params and uniforms

src/main/java/titanicsend/pattern/yoffa/shader_engine/PatternControlData.java
```diff
@@ -2,14 +2,15 @@
...
 
 public class PatternControlData {
 
-  private Map<Uniforms.Audio, Float> uniformMap;
-
   TEPerformancePattern parent;
+
+  // TODO - Legacy audio processing code preserved for compatibility
+  // TODO - with old shaders.  This should be removed at some point.
+  // TODO - The new (GLShaderPattern, etc.) system doesn't use it.
```
andrewlook, line 13: nice, thanks for calling this dependency out.

How can I tell which old shaders would be depending on this? If I understand correctly, `getFrequencyData` and `getWaveformData` below are still used when setting uniforms in the newer GLShaderPattern etc.

andrewlook (commented):
Awesome work @zranger1 !! Thank you so much for doing this.

I took a read through the code and added one or two suggestions. Apart from that I mostly asked questions or celebrated some of the improvements you made.

I'll do some testing on my machine but overall this PR is looking great to me - seems like it's worth prioritizing getting this merged since it's a relatively big change and we have some other PR's in flight depending on this (ex. @sina-cb 's work on TD).

Once I do my local testing I'll be ready to approve! 👍

resources/shaders/ndidefault.fs
```diff
@@ -0,0 +1,43 @@
...
+
+    // out-of-range pixels are transparent black
+    if (uv.x < 0.0 || uv.x > 1.0 || uv.y < 0.0 || uv.y > 1.0) {
+        fragColor = vec4(0.0);
+        return;
+    }
+
+    // otherwise, get the color, and calculate its luminance value
+    vec4 vid =  gain * texture(ndivideo, uv);
+    float gray = dot(vid.rgb, vec3(0.299, 0.587, 0.114));
```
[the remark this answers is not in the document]
zranger1: Yes, they are.   This converts to grayscale based on perceived brightness.   As with other such things, there's are a lot of conditions attached, and lots of "animated discussion", but this is widely used, and gives a good result.

src/main/java/titanicsend/effect/BasicShaderEffect.java
```diff
@@ -0,0 +1,36 @@
+package titanicsend.effect;
+
+import heronarts.lx.LX;
+import heronarts.lx.LXCategory;
+import heronarts.lx.parameter.*;
+import titanicsend.pattern.glengine.GLShader;
+import titanicsend.pattern.glengine.GLShaderEffect;
+
+@LXCategory("Titanics End")
```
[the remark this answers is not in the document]
zranger1: Oh wow, yes.  And we should have a big category cleanup before the next show.  The list is large and kind of ...not quite helpful... right now

src/main/java/titanicsend/pattern/jon/ArcEdges.java
zranger1: I've got to think about this.  Presumably, even once we have the ability to adjust fixtures at run time, the edges won't move around too much.   We might think about letting patterns register a "car geometry changed" listener when we get to that point.

src/main/java/titanicsend/pattern/jon/EdgeFall.java
zranger1: We can do that!  We should talk about how it should be parameterized.  I think having the lines "unexplode" after a little (adjustable) while would be quite cool.

src/main/java/titanicsend/pattern/jon/Fireflies.java
zranger1: Ordinarily, these would be set by GLPatternControl, but by design, the settings in user pattern code override the normal ones.   (There's no performance penalty for this -- GLPatternControl maintains these things in an internal map, and nothing actually gets written to the GPU 'till we're ready to let OpenGL draw.   I really ought to do some nice architecture diagrams with boxes and arrows.  I just keep putting it off.  )

src/main/java/titanicsend/pattern/yoffa/shader_engine/PatternControlData.java
zranger1: This code should actually never be hit by the TE App as it stands now.   All the shaders have been converted to the new system.   I left the old system in and intact for now, just in case.

src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
zranger1: You've got it right -- GLPatternControl handles intrinsic uniforms for patterns (which have TECommonControls),  GLEffectControl handles them for effects (which have more individualized controls.)

If we add a third class of things that uses shaders in yet another way, we'll likely need to derive a GLControlData variant for it too.

src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
zranger1: Good catch!   Just pushed a commit to fix it.

src/main/java/titanicsend/pattern/jon/Fireflies.java
zranger1: (and you're right -- it just resets the defaults.   Not setting those uniforms in OnFrame() would have exactly the same result in the shader.  There was some Really Good Reason I overrode the defaults there - probably just to illustrate how OnFrame() handles uniforms.    At least that's what I'll claim! )

resources/shaders/ndidefault.fs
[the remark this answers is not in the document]
andrewlook: got it, thank you!

src/main/java/titanicsend/effect/BasicShaderEffect.java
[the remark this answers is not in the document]
andrewlook: haha i think it'll be manageable :) I'm happy to help wrangle/organize the rest. I saw one of your other patterns uses the category "TE Examples", maybe that's a good category? (side note: we've discussed package structure within the "pattern" package before and migrating to something other than people's names... as a dev, seeing a package under `titanicsend.pattern.examples` might make things easier. Outside the scope of this PR definitely, just wanted to float that idea)

src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
andrewlook: excellent, thanks for confirming!

src/main/java/titanicsend/pattern/jon/ArcEdges.java
andrewlook: makes sense 👍

src/main/java/titanicsend/pattern/jon/EdgeFall.java
andrewlook: nice, i like the idea of an "adjustable" while. could be time (1sec) or tempo based (one measure / four beats / etc), the way you've been parametrizing things like the random strobe effect (even though that's a strobe interval, i like that style of configurability for time-related things)

src/main/java/titanicsend/pattern/jon/Fireflies.java
andrewlook: ah nice, that makes sense! It was certainly helpful for me to see where these get set :)

src/main/java/titanicsend/pattern/yoffa/shader_engine/PatternControlData.java
andrewlook: cool, thanks for clarifying 👍

src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
andrewlook: nice, thanks!

andrewlook (approved):
finished stress-testing locally and it's looking great to me 🙌 🚀

PR #449, opened 2024-03-27
Add APC40 setting to disable device control knobs

jkbelcher:
Allows knobs in the DEVICE CONTROL section of the APC40mkII to be used for generic mapping instead of controlling the focused device.

Note the knob indicator lights will not update for generic mappings.

### DIFF
--- a/src/main/java/titanicsend/lx/APC40Mk2.java
+++ b/src/main/java/titanicsend/lx/APC40Mk2.java
@@ -862,7 +862,7 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
     this.deviceListener = new DeviceListener(lx);
     addSetting("masterFaderEnabled", this.masterFaderEnabled);
     addSetting("crossfaderEnabled", this.crossfaderEnabled);
-    addSetting("deviceControlEnabled", this.deviceControl);
+    addSetting("deviceControl", this.deviceControl);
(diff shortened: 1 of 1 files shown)


src/main/java/titanicsend/lx/APC40Mk2.java
```diff
@@ -847,6 +849,10 @@ private void sendChannelKnob(int value) {
       new BooleanParameter("Crossfader", true)
           .setDescription("Whether the A/B crossfader is enabled");
 
+  public final BooleanParameter deviceControl =
+    new BooleanParameter("Device Control", false)
+      .setDescription("Use the device knobs on the right side of the APC40mkII to control the focused device");
```
andrewlook, line 854: just to make sure I understand, is this a setting within LX to enable a "mode" where the device knobs are unbound? (as opposed to being some input on the device that controls the mode?)

jkbelcher: Yes, and by default ("off") the knobs are not mapped to the focused device.  You'll also want to turn on midi control events for that surface under the MIDI INPUT section:
image

andrewlook (approved):
Just tested this out on APC40 and it works great!

Also thanks Justin for the pointers about what to pay attention to in the MIDI/OSC tab of Chromatik, super helpful.

PR #450, opened 2024-03-27
remove RhythmicFlashingStatic (from base Oct_21 showfile and available patterns list)

andrewlook:
The pattern killed FPS during LA show so I'm commenting the pattern code out, so it doesn't show up in the pattern list in Chromatik. (Note that I didn't need to remove it from TEApp, since TEApp bulk-includes patterns from `OrganicPatternConfig.java`

Keeping the underlying logic around for future reference - this could look good as a native shader, particularly with audio-reactivity hooked up to the amount of static.

### DIFF
--- a/Projects/Oct_21.lxp
+++ b/Projects/Oct_21.lxp
@@ -1,6 +1,6 @@
 {
   "version": "0.4.2.TE.7-SNAPSHOT",
-  "timestamp": 1697916459321,
+  "timestamp": 1711571514462,
   "model": {
     "id": 2,
     "class": "heronarts.lx.structure.LXStructure",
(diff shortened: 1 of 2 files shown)


zranger1 (approved):
Approved!   @andrewlook, do we need a flashing static pattern?   If so, I imagine I can come up with a fast one pretty quickly.

andrewlook:
> Approved! @andrewlook, do we need a flashing static pattern? If so, I imagine I can come up with a fast one pretty quickly.

Thanks @zranger1 - If we had an audio-reactive flashing static pattern that would be really cool, if you think it wouldn't be too much trouble!

PR #451, opened 2024-03-27
Pattern Improvements: Electric.fs audio-reactivity

andrewlook:
A few notes so far:
- I found that using bass/trebleRatio over bass/trebleLevel generates more movement, since (especially with a moderate “release” setting), the raw bass/treble levels tend to stay pretty high once the song gets going, but ratio will have more dynamic movement.
- The main thing I want to fix in the example below is to clamp the effect of the ratio when it’s over a certain value. When the audio input goes from silent to active song, there’s a huge color blast when the ratio is very high at first, before a moving average builds up.
- In the example below I hooked up iWow1 (circular distortion) to treble, and a small effect from bass on the thickness/light diffusion params. I left iWow2 as a free param for VJ (was experimenting with hooking it up to bass, may try that in a forked version).


resources/shaders/electric.fs
```diff
@@ -80,19 +80,39 @@ void mainImage(out vec4 fragColor, in vec2 fragCoord)
...
+
+    // TODO: this might be better named to something like "zoom"
+    float thickness = 3.0 + iScale;
+    thickness += 0.1*bassEffect;
+
+    float lineWidth = iQuantity;
+    lineWidth += 0.1*bassEffect;
+
     // distort UVs a bit
     uv = cart2polar(uv);
```
zranger1, line 105: Just tested this w/a nice, loud recording of a live show.  It looks fantastic!  Adding reactivity is really going to make this gig amazing!  (I do see what you mean about brief oversaturation of the ratios when the music starts though.)

One suggestion:   To add a little high frequency response to this, try messing with the x coordinate while the coordinates are transformed to radial.  Maybe after line 105, add something like this (but modulated by the appropriate control)

```
    float wave =  texelFetch(iChannel0, ivec2(fragCoord.x, 1), 0).x;
    uv.x += wave;
```
I'll post video to Slack in a minute so you can see if you like the idea.   (and this is def not a blocker)

andrewlook:
That’s a fantastic idea, thank you!

resources/shaders/electric.fs
```diff
@@ -80,39 +80,35 @@ void mainImage(out vec4 fragColor, in vec2 fragCoord)
...
+    uv = cart2polar(uv);
     float twistDistortion = iWow1;
     twistDistortion += trebleEffect;
-    //twistDistortion += bassEffect;
-
-    float repetitionFactor = iWow2;
-    repetitionFactor += 0.2*bassEffect;
+    uv.y += 2.5 * twistDistortion * (.5 + .5 * sin(cos(uv.x) * len));
+    if (frequencyReact > 0.){
+        float wave = texelFetch(iChannel0, ivec2(fragCoord.x, 1), 0).x;
```
zranger1, line 92: Looks great!   Just for fun, give this a try and see what you think.  It makes the coordinate distortion circular, and at high values kinda looks like a speaker cone moving. (at line 92)
       ` float wave = 8.0 * texture(iChannel0, vec2(uv.x / 4.0,1)).x;`

andrewlook: wow! that's a fantastic idea, i like this version a lot! thanks, i'll make this change.

zranger1 (approved):
Looks good!

PR #452, opened 2024-03-27
Add Audio Reactivity controls

zranger1:
The new controls,  Level Reactivity and Frequency Reactivity, will be the first two controls on the midi surfaces.  The old per-pattern color controls have been moved to the second page, and are no longer on the midi controller.

### DIFF
--- a/resources/shaders/circuitry.fs
+++ b/resources/shaders/circuitry.fs
@@ -7,7 +7,7 @@
 #pragma name "Circuitry"
 #pragma TEControl.SIZE.Range(2.0,5.0,0.1)
 #pragma TEControl.QUANTITY.Range(4.0,3.0,6.0)
-#pragma TEControl.WOW2.Value(0.6)
+#pragma TEControl.WOW2.Disable
 #pragma TEControl.WOW1.Disable
 #pragma TEControl.WOWTRIGGER.Disable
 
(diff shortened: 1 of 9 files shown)


resources/shaders/circuitry.fs
```diff
@@ -37,9 +37,9 @@ float fractal(vec2 p) {
 
         // animated manhattan distance gives size changing rectangular
         // features
-        float m = abs(p.x+sin(iTime * 2.));
+        float m = abs(p.x+fract(bassRatio) + 0.3 * sin(beat));
```
andrewlook, line 40: cool! nice job combining the tempo beat with the bass here 👍

andrewlook: do you think the `levelReactivity` should modulate how much `bassRatio` and/or `beat` have a visible effect?

I noticed that when I scaled `levelReactivity` all the way down, the `beat` seems to still be driving some movement in the pattern.

andrewlook: (this is non-blocking feedback btw! i'll merge this as-is since it's a big improvement and helps us get the controls code updated, just thinking out loud about whether beat effects should fall under levelsReactivity)

andrewlook (approved):
Thanks for making these changes @zranger1 !

resources/shaders/circuitry.fs
zranger1: You're right!   I left that in, thinking the minor beat-linked pulsing was fairly harmless.   Wonder if we shouldn't eventually have three reactivity knobs though -- maybe, Tempo, Volume, Frequency?

andrewlook: hmm that could make sense - i'll think about it some more too. maybe tempo and levels/volume would tend to affect the same variables and operator could switch between them.

PR #453, opened 2024-03-31
Kaleidosonic pattern update: audio reactivity

zranger1:
Greatly enhances response to audio.
- Wow1 controls the overall feature size
- Wow2 controls contrast
- level and frequency reactivity controls are implemented andhave reasonable default settings, but can also be pushed to do very weird things.

### DIFF
--- a/src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
+++ b/src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
@@ -12,8 +12,8 @@ public class Kaleidosonic extends GLShaderPattern {
   public Kaleidosonic(LX lx) {
     super(lx, TEShaderView.ALL_POINTS);
 
-    controls.setRange(TEControlTag.WOW1, 0.04, 0, 0.08); // bass response
-    controls.setRange(TEControlTag.WOW2, 0.5, 0.2, 3); // overall audio level adjustment
+    controls.setRange(TEControlTag.WOW1, 1., 0, 2.5); // complexity
+    controls.setRange(TEControlTag.WOW2, 0.5, 0.2, 3); // contrast
     controls.setRange(TEControlTag.SIZE, 1., 5, 0.2); // scale
     controls.setRange(TEControlTag.QUANTITY, 7, 1, 13); // number of kaleidoscope slices
     controls.setValue(TEControlTag.SPIN, 0.125);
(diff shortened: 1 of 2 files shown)


src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
```diff
@@ -12,26 +12,15 @@ public class Kaleidosonic extends GLShaderPattern {
   public Kaleidosonic(LX lx) {
     super(lx, TEShaderView.ALL_POINTS);
 
-    controls.setRange(TEControlTag.WOW1, 0.04, 0, 0.08); // bass response
-    controls.setRange(TEControlTag.WOW2, 0.5, 0.2, 3); // overall audio level adjustment
+    controls.setRange(TEControlTag.WOW1, 1., 0, 2.5); // complexity
+    controls.setRange(TEControlTag.WOW2, 0.5, 0.2, 3); // contrast
     controls.setRange(TEControlTag.SIZE, 1., 5, 0.2); // scale
```
andrewlook, line 17: The default size works very well - I'd consider setting the range to go between 0.5-3.0.

at the very low end, 0.2 seems to be zoomed in so far that it produces very quick flashes. at the very high end, it seems like the shape is zoomed too small on the car. I think about 2.5 is the threshold where it covers the car geometry, so I like having a little margin where it could take up *slightly* less space if we want it to (but I very much like that the default size covers the car 👍 ).

I also noticed the size parameter's numerical range is inversed from the direction of the knob. Not a huge deal, but maybe it'd be worth inverting how scale gets applied inside the shader so that the parameter values make more sense to an operator?

andrewlook: related to default parameter values: is it possible to set the default LvlReact and FreqReact to be around 0.05 - 0.1 or something in that range? That's the only other feedback that comes to mind

andrewlook (approved):
This pattern is looking excellent!! I love how versatile it is, and I feel like I can find it useful as a low energy pattern with lower reactivity settings, or a super high-energy pattern. 

Lots of variability with the complexity / contrast / quantity options too. I think we'll get a lot of mileage out of this one. Thanks so much for this!

src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
zranger1: - Size range changes: done, will be in next commit (also changed default size to 2 b/c it looks better on startup.
- Reactivity defaults: done, will be in next commit.
- Size direction:  Not sure about the best thing to do here.  (Possibly just hide the numerical value from the operator :-P)    Or better still, have LX display a different range than the control is actually producing.  We could probably generically have the control report its normalized range, 0-100%,  regardless of what the pattern math is doing.  Since we're short on time though, I'm inclined to leave this for now and do it post-Coachella.

andrewlook (approved):
Awesome, thanks for making these control changes!

src/main/java/titanicsend/pattern/jon/Kaleidosonic.java
andrewlook: totally agree, the size numbers are not at all a big deal - thanks for making the more relevant changes!

PR #454, opened 2024-04-02
Add audio reactivity to light_beam shader

sina-cb:
Making light_beams.fs sound reactive.

This PR adds sound reactivity to the light_beams shader by using the bassLevel to control the spin of the shader, and the trebleLevel to control the speed.

This PR also adds three utility classes:
 1. SignalLogger: which periodically writes signal values to a CSV file. This class is accompanied by a TouchDesigner sketch that will read the CSV data periodically and visualizes the signals. This can be used to visually inspect the signals used to make the patterns sound reactive.
 2. GaussianFilter: adds a gaussian filter on the input signal. This class has a filter width parameter and higher values will result in smoother outputs.
 3. SignalLag: This is a simple lag effect on the input signal. This smoothens sudden changes in the signal and allows smoother transitions when needed.

Tested the new shader with different live music sets and the shader can be used in high beat and low beat music scenarios.

Here's a sample recording of the new shader:
https://github.com/titanicsend/LXStudio-TE/assets/1153540/52348541-155d-4550-add0-9aff260ca21e


src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -28,16 +34,107 @@ protected List<PatternEffect> createEffects() {
...
+
+    // Kept internally to keep track of the time difference applied to the speed of the shader
+    // based on the trebleLevel.
+    private float timeDiff = 0;
+
+    // Gaussian filter used to smoothen out the bassLevel
+    GaussianFilter degreesFilter = new GaussianFilter(10);
+
+    // Gaussian filter used to smoothen out the trebleLevel
+    GaussianFilter timeDiffFilter = new GaussianFilter(20);
```
andrewlook, line 48: Just for the sake of standardizing, I'd make these `private` member variables unless there's a good reason not to.

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -28,16 +34,107 @@ protected List<PatternEffect> createEffects() {
...
+    float levelReactivityControl = 0;
+    float freqReactivityControl = 0;
```
andrewlook, line 50: I'd lean towards omitting these member variables - patterns usually access the controls directly through the controls API.

I'm a big fan of when the data flow is clear from the function signatures - so personally, I'd either:
1. declare these as local variables on lines 69 and 71, and pass them into the other functions as `getRotationDiff(levelReactivityControl)` and `getTimeDiff(frequencyReactivityControl)` (which would make it clear to someone reading the code which values are going to be modulated by audio), or
2. do the control lookups directly inside getRotationDiff/getTimeDiff (i.e. moving line 69 -> 97, 71->133)

src/main/java/titanicsend/util/GaussianFilter.java
```diff
@@ -0,0 +1,61 @@
...
+  private int filterWidth;
+  private double filteredValue;
+
+  public GaussianFilter(int filterWidth) {
+    this.filterWidth = filterWidth;
+    filterBuffer = new double[filterWidth];
+    Arrays.fill(filterBuffer, 0.0f);
+  }
+
+  public double applyGaussianFilter(double currentBassLevel) {
```
andrewlook, line 29: maybe `currentBaseLevel` could be renamed to `rawValue` or `latestValue`?

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -3,9 +3,15 @@
...
 import heronarts.lx.LXCategory;
 import heronarts.lx.parameter.LXParameter;
+import java.util.Arrays;
+import java.util.List;
 import titanicsend.pattern.glengine.ConstructedShaderPattern;
+import titanicsend.pattern.glengine.GLShader;
 import titanicsend.pattern.jon.TEControlTag;
 import titanicsend.pattern.yoffa.framework.TEShaderView;
+import titanicsend.util.GaussianFilter;
+import titanicsend.util.SignalLogger;
```
andrewlook, line 13: looks like you might be able to remove the `SignalLogger` import (I'm assuming some code was there for logging before but has since been removed)

andrewlook (approved):
This is so cool! I love this pattern and how it uses bass in more creative ways. Jeff and I were complaining at the LA show how most of our ideas for bass effects are modulating size, so it's cool to see a more creative way to use that signal!

Also thanks for making the changes discussed in slack around default parameter values 👍  I think that'll make life easier when we're operating it live.

Also thanks for making utils for `GaussianFilter`/etc - those will definitely come in handy :)

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -28,16 +34,107 @@ protected List<PatternEffect> createEffects() {
   // multiple
   @LXCategory("Native Shaders Panels")
   public static class LightBeamsPattern extends ConstructedShaderPattern {
```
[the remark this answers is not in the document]
sina-cb: Moved into a new pattern class file.

I know we don't wanna keep using author names, but for the lack of a better structure, I put the pattern under sinas/ for now.

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
sina-cb: I actually thought about this, but I was lazy to make the change at the time. Changed them to local variables and passing them to the function that uses them.

src/main/java/titanicsend/util/GaussianFilter.java
sina-cb: I blame and credit Gemini for this. Changed! :)

src/main/java/titanicsend/util/SignalLogger.java
```diff
@@ -0,0 +1,115 @@
...
+ * to start the logging. > List<String> signalNames = Arrays.asList("signal_1", "signal_2");
+ *
+ * <p>> logger = new SignalLogger(signalNames, "Logs/signal_data.csv"); > logger.startLogging(1); //
+ * Write to CSV every 1s.
+ *
+ * <p>In your code, where you want to keep track of the signals, you can log the signals as it
+ * follows: > List<Float> values = Arrays.asList(signal_1_value, signal_2_value); >
+ * logger.logSignalValues(values);
+ */
+public class SignalLogger {
```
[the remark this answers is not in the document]
sina-cb: Exactly.

I first added a "debug" flag in the class that if set to false, all functions short circuited and returned quickly to easily disable the class.

but until we have a better approach for taking care of the debug value (I think a runtime argument would be great) I removed that and I think we should just remove the logger when we're done with it.

For visualizing, just install the latest TD, and open the sketch and it should start showing the signals automatically.

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
sina-cb: Done

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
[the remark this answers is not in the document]
sina-cb: Update: I removed the old light_beam pattern and it started causing failures with the project file. To keep backward compatibility, I added a new Pattern called LightBeamsAudioReactive.java and kept the original one in place too.

src/main/java/titanicsend/pattern/sinas/LightBeamsAudioReactivePattern.java
```diff
@@ -0,0 +1,117 @@
...
+  private float timeDiff = 0;
+
+  // Gaussian filter used to smoothen out the bassLevel
+  private GaussianFilter degreesFilter = new GaussianFilter(10);
+
+  // Gaussian filter used to smoothen out the trebleLevel
+  private GaussianFilter timeDiffFilter = new GaussianFilter(20);
+
+  public LightBeamsAudioReactivePattern(LX lx) {
+    super(lx, TEShaderView.ALL_PANELS_INDIVIDUAL);
```
andrewlook, line 27: I think we usually use `TEShaderView.ALL_POINTS` for these, since `ALL_PANELS_INDIVIDUAL` means that the coordinate space gets repeated on each panel, like this:

image

So if we want to indicate we prefer for this pattern to default to a single coordinate space covering all the edges and panels, `ALL_POINTS` would be a better bet. (For context, some of the patterns that have been around longer had `ALL_PANELS_INDIVIDUAL` as their default but would have benefited from `ALL_POINTS` as the default.

That being said, I think given that we migrated LX/Chromatic to channel-based views being the default rather than pattern-based view selection right before the October show, this might not be a huge deal - that default may not be getting used at the moment. But given that this is our way of signalling to the superclass what the preferred view for this pattern is, I think it'd be smart to use `ALL_POINTS`. (So this is not a huge deal, but it seemed like a good time to give you this context - happy to explain more offline too, since the codebase is in a bit of a transition due to this "views" concept).

src/main/java/titanicsend/util/SignalLogger.java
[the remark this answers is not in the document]
andrewlook: excellent, thank you so much for adding docs here too - i think this is going to be really useful! (and it's a good motivation for me to finally install TD :) )

andrewlook (approved):
this looks great! I left one tiny comment about the default view but otherwise I tested this and it looks great so I'm ready to merge!

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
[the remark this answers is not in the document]
andrewlook: got it - FWIW it would have been OK to break the project file since we're starting over, but I'm perfectly happy to have a forked version of this pattern as a standalone.

Also very much agreed that we'll want to better organize our patterns later, but for now putting it in a per-user folder is the simplest move!

src/main/java/titanicsend/util/GaussianFilter.java
andrewlook: hahahaha awesome :)

PR #455, opened 2024-04-03
eletric.fs: circular effect

andrewlook:
whoops! Before #451 got merged, I forgot to push a one-line change based on Jon's suggestion for a circular waveform effect


zranger1 (approved):
Approved!  Glad you remembered this!

PR #458, opened 2024-04-04
tweak(edgefall): use trebleLevel over volumeRatio for line dance. add bassLevel to glow

andrewlook:
I like the line-dance effect, though I noticed that volumeRatio tends to stay pretty stable while a song is active, so the line dancing didn't seem to me to obviously correspond to the beat/etc.

So when a song transition takes us from totally silent to sound starting, there'd be a big movement. But during the song when a beat dropped, there wasn't a big visible change.

I changed a couple lines of code here that I think behave really nicely throughout the song lifecycle.


zranger1 (approved):
This is excellent, thank you!   It makes the movement both more varied and more subtle -- I like it!

PR #459, opened 2024-04-04
tweak(circuitry): modulate fract(bassRatio) by levelReact

andrewlook:
tiny change here - i was playing with circuitry, and noticed that when I dialed levelReact all the way down, there was still some audio-driven movement. I just went ahead and multiplied in levelReact there, but this pattern is looking fantastic with the audio reactivity already!


zranger1 (commented):
Thanks!  Good to go now!

zranger1 (approved):

PR #460, opened 2024-04-04
RhythmicStatic - replacement for RhythmicFlashingStatic

zranger1:
Pretty much duplicates old pattern's functionality at this point.   What else do you think we should make it do?
- Quantity controls the static field's density
- Size controls the static "pixel" size.  (in power-of-two-squares!)
- Wow1 controls depth of flash-with-beat.


andrewlook (approved):
This looks great! Going to merge it now, and I have some experimental ideas for audio-reactivity that I'll share as a PR to get your feedback!

PR #461, opened 2024-04-04
Waterfall: Various Enhancements

zranger1:
- **Audio (level) reactivity**:  Messes with vortex rotation in a way that preserves the "waterfallness" while still being obviously reactive.  Can be cranked to a silly (but still cool) level!
- **Tint**:  Now uses palette colors by default.  Improved definition of tinted colors, especially in background and mist.
- **WowTrigger**:  Try it.  This was @andrewlook's idea!  It is pretty dramatic and fun!

Note for users:  This pattern really only works on the full car (default view) because it uses some car geometry to determine where to break the flow around the DJ booth.

### DIFF
--- a/resources/shaders/waterfall.fs
+++ b/resources/shaders/waterfall.fs
@@ -4,6 +4,7 @@
 #pragma name "Waterfall"
 #pragma iChannel1 "resources/shaders/textures/icecliff.png"
 #pragma TEControl.WOW1.Range(0.4,0.0,1.0)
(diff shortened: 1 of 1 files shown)


andrewlook (approved):
This is excellent, thanks for making these changes!! Super solid pattern, and I feel like the defaults will be really safe if anyone has to hit the panic button in the moment. Very cool idea for the audio reactivity!

Also the wowtrigger works really well, either to pulse it with the beat or to hold it!

PR #462, opened 2024-04-04
SpiralDiamonds:  Audio reactivity

zranger1:
This pattern is now capable of... extreme grooviness.

- Level Reactivity controls bass response
- Frequency Reactivity controls treble response


andrewlook (approved):
OMG Wow!

This one is awesome right out the gates. Merging immediately!

It's super flexible too - some aesthetic notes to self on how we might apply this pattern:
- experimenting with really low values of quantity and/or size makes completely new patterns that are also cool (more abstract, fill a lot of the car geometry while still being audio reactive)
- Looks great even with freqReact=0 or levelReact=0.
- High freqreact + high slope (on global audio panel) creates really interesting organic shapes

PR #463, opened 2024-04-05
RhythmicStatic: adding optional audio reactivity

andrewlook:
Some tweaks I made based on #460 -

- bass/lvlReact modulates size
- treble/freqReact modulates quantity (also setting a lower default quantity so the effect is more pronounced).
- wowTrigger temporarily speeds up time proportionately to treble

Note: the bass effect is quantized (converted to integer before being used as a power) and clamped (so the bassRatio doesnt go overboard when there's a big transition but is still visible).

Also, removed the old version of the shader so it's in our git history but doesn't add any work when we reorganize patterns in the future.


zranger1 (approved):
Framework will love this one -- it's essence of pure Techno!

PR #464, opened 2024-04-05
Add sound reactivity and noise to neon_cell shader

sina-cb:
This PR add sound reactivity to the neon_cell shader.

In this PR I am also adding a simplex3D noise in the neon_cell shader which is controlled by the wow1 control.

The levelReacitivityControl is used to control the speed on the shader, and frequencyReactivityControl is used to control the glow of the neon cells.


resources/shaders/neon_cells.fs
```diff
@@ -16,15 +68,33 @@ vec2 rotate(vec2 point, float angle) {
...
+void mainImage(out vec4 fragColor, in vec2 fragCoord) {
+    // Normalized UV coordinates
+    vec2 uv = (fragCoord.xy - 0.5 * iResolution.xy) / iResolution.y;
+
+    // Glow logic
+    float glow_audio = bassLevel * trebleLevel * frequencyReact;
+    float glow = 0.1 + min(0.6, glow_audio); // glow levels that look good: min=0.1, avg=0.2, max=0.5
+
+    // Associate noise to the p and q parameters of "pat" function with the iWow1.
+    float noise_uv_multiplier = 10; //3 * (1 + bassLevel); // TODO: Make this a very smooth bass
```
andrewlook, line 80: I tried changing this line of code to:
```
float noise_uv_multiplier = 3. * (1. + bassRatio);
```

and it looked awesome! BassRatio is a little smoother than bassLevel, and this also lets us use `iWow1` to switch on/off another kind of audio reactivity - the 3D simplex noise looks super cool in how it modulates the shape!

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -256,17 +262,105 @@ protected void createShader() {
...
     protected void createShader() {
-      controls.setRange(TEControlTag.SIZE, 1, 2, 0.25); // overall scale
-
-      addShader("neon_cells.fs");
+      controls.setRange(TEControlTag.SIZE, 1, 4, 0.25);
+      controls.setRange(TEControlTag.SPIN, 0.0, -.4f, .4f);
+      controls.setRange(TEControlTag.WOW1, 0.1, 0, 1);
+
+      controls.setRange(TEControlTag.LEVELREACTIVITY, 0.2, 0, 3);
+      controls.setRange(TEControlTag.FREQREACTIVITY, 0.8, 0, 2);
```
andrewlook, line 295: really good defaults! Thanks for these.

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -256,17 +262,105 @@ protected void createShader() {
...
+      float filteredRms = (float) rmsFilter.applyGaussianFilter(rms);
+      float controlledRms = rms * controlLevel;
+
+      sumTimeDiff += controlledRms;
+      logger.logSignalValues(Arrays.asList(controlLevel, rms, filteredRms, controlledRms, sumTimeDiff));
+      return sumTimeDiff;
+    }
+
+    /**
+     * Calculates the Root Mean Square (RMS) of a given audio sample array, with the option to
```
zranger1, line 324: Weird, accidental discovery of the day:  LX already provides RMS level as well as level in decibels, through strangely named, nearly undocumented APIs.   We have access through TEAudioPattern, so to get the current RMS level from java, call:

`eq.getRaw();   // get raw (without envelope processing) RMS level`

and to get the level in db, call

`eq.getDecibels();`

In both cases the variants that take an integer argument let you get those values for a single fft band.   Don't know if these will work for you, but maybe worth a try!

(These, and some other interesting audio tools we're not currently using and probably should be, live in the file:
`repo\com\heronarts\lx\0.4.2.TE.7-SNAPSHOT\lx-0.4.2.TE.7-SNAPSHOT.jar!\heronarts\lx\audio\DecibelMeter.class`)

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -256,17 +262,105 @@ protected void createShader() {
...
+     * perform downsampling for optimization.
```
andrewlook, line 325: cool, thanks for adding this and documenting it! 

so we're summing squares of all the eq frequency buckets available, averaging them, and taking the square root - is the idea that we want to detect big spikes in the frequency distribution (is this what you meant by "kick detection"?)

Also noob question: what function does downsampling serve here, is it for efficiency  / looking at less samples?

resources/shaders/neon_cells.fs
andrewlook: (FWIW I actually like not having `levelReactivity` multiplied in here - it lets us control the "bass pulse" separately from that nice movement effect controlled by levelReactivity).

andrewlook (approved):
This is looking great!

A couple of small notes:
- I really liked the bass effect on the noise multiplier when I played with it - I left a comment on that line, using `bassRatio` instead of `bassLevel` worked very well for me.
- I'd add `TD/Utils/Backup/` to `.gitignore` and remove those extra TD sketches from the PR.

Apart from that, this is looking good and ready to merge! Thanks for this!

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
zranger1: Def not a showstopper at this point though.   This is lovely and fluid work!

andrewlook (approved):

PR #466, opened 2024-04-06
Fix for NDI send on project file load/unload

zranger1:
One line fix:  When the sender is destroyed, explicitly close the channel and release the channel name.  Don't wait for Java to do it.   Doh!

What was happening:  Project file unload was destroying the Java objects, which didn't result in immediately closing and removing the active NDI channel. 

NDI receivers interpret that as the channel still existing, but just not sending new frames. Subsequent attempts to re-open the channel fail silently because NDI believes a channel of that name is already active until the native object associated with the now-dead Java channel object finally gets released.  The Java GC will take care of this eventually, but since it involves native memory, "eventually" might be a very long time.

Since we're using one channel name for all TE NDI output, explicitly closing the channel on dispose frees that name for immediate re-use by the next loaded pattern.   (FWIW, we've seen this, "must explicitly release native resources if you want to use them in a new project or pattern", with OpenGL too. )

### DIFF
--- a/src/main/java/titanicsend/effect/NDIOutRawEffect.java
+++ b/src/main/java/titanicsend/effect/NDIOutRawEffect.java
@@ -62,6 +62,9 @@ public class NDIOutRawEffect extends TEEffect {
 
   @Override
   public void dispose() {
+    if (isInitialized) {
+      ndiSender.close();
+    }
     super.dispose();
   }
 }


sina-cb (approved):

andrewlook (approved):
Nice, thanks for the fix! Those native resource leak bugs can be very subtle. merging this now.

Just out of curiosity, what were the steps to reproduce this bug? Run LX with an NDI pattern open, and then close or open a project and try to run another NDI pattern?

sina-cb:
This is affecting the NDI out which is an effect added to a channel. When the NDI out is added, if you switch project files, thee NDI out effect stops working.

zranger1:
@andrewlook , to confirm the fix/repro the bug:
- add the NDI output effect to the master channel of a project and save it.
- verify using a program that receives our NDI (either TD, or my Processing TEModelViewer) that correct data is being sent
- load any other project file (If the new project doesn't use NDI out, the receiver should stop showing new frames at this point.)
- load your previously saved NDI output pattern, or any other that has the NDI effect on Master.
- now that the bug is fixed, NDI output should start showing the current LX output on the receiver within a few seconds. (The bug was that output was not able to restart due to channel naming conflict.)

andrewlook:
Thanks for explaining!

PR #467, opened 2024-04-06
April 12th showfile (draft)

andrewlook:
About the showfile:

## Channel Layout

I'm using channel 1 to build up our list of full-view patterns, and duplicating it (rather than adding new patterns to each channel individually). Channel 4 for edge patterns, channel 8 for FX (patterns with a wow trigger).

Screen Shot 2024-04-06 at 2 41 01 PM

Within that channel, I put our new audio-reactive patterns up top, prefixed with `@`. High energy patterns will have a `+` prefix, and low-energy patterns have a `-` prefix in the name (Selecting the pattern and using `cmd-R` lets you rename a pattern instance once its been added to a channel).

Later, I'll make our "Low-energy" channels by erasing everything except the `-` patterns, and "High-energy" channels by erasing everything except the `+` patterns. (I'll probably try to make `+` and `-` versions of all our audio-reactive patterns).

But I duplicated channel 1 into roughly the final APC40 layout so we have something to practice with in the meantime.

Before showtime, we should enable autovj once and disable it, so all the autogenerated channels get added ahead of time.

## Effects

- Added explode effect on master channel, in manual mode.
  - [ ] TODO: MIDI mapping for the "manual" trigger mapped to a bank button on the APC40

## Modulators

- Removed all tempo modulators
- DMX opacity dimmer changed to brightness-only
  - Connected the brightness modulator to a global Hue/Saturation Effect on the Master channel

Screen Shot 2024-04-06 at 2 07 33 PM

- Removed "Group Main" on Channel 1
- Removed "Grid Solid"
- Kept Channel "DMX Grid 1" - it's last in the channel list, and set to "multiply" so it acts as a mask on the output of all previous channels. (you can test this out by fading the channel all the way up, and selecting the "Chevron" pattern in that channel.
  - [ ] TODO: test DMX grid - does this channel need to be fully enabled before the show starts?
- [ ] TODO: test DMX controls using Luminair

## NDI

- [ ] TODO: do we add a separate channel with the NDI input pattern?
- [ ] testing


zranger1:
@andrewlook, on NDI:  yes, I think we add a separate channel.  It just blends with everything else.  Bet they'll want a DMX control though, so they can determine when the projection is the only thing (or at least the dominant thing) shown on the car.

andrewlook:
@zranger1 sounds good - I can add that channel in a separate PR after this one (I'd like to get a base showfile into `main` and then make incremental PR's as needed 👍

zranger1 (approved):
This looks good so far, and as you said, once it's in we can make small changes as needed.

sina-cb:
The new layout of the channels look good. I tried to play with the showfile a little and pretty much things looked good.

A couple of things to note that I think might be due to the branch not having the latest code maybe.
1. When loaded the showfile, IntelliJ was complaining about Waterfall and Circuitry patterns being missing
2. The neon_cells pattern didn't have the audio reactivity change. again might be due to base version.

Overall looks great tho.

andrewlook:
Sina good callouts, thanks - neonCells was because that branch had not
been merged in yet.

Waterfall and circuitry used the new “auto class generator” for shader
patterns without a corresponding Java class. Cc @zranger1 - This has been
handy for development, though maybe for show time its worth creating a java
class so those patterns can persist in an LXP file?

zranger1:
Waterfall and circuitry load and run normally from April12.lxp file here, on both Windows 10 and Mac. (Java Temurin 17.0x on both machines - it might matter.)

@sina-cb,  can you pull and rebuild main and see if the shaders still won't load?     

Out of an abundance of caution, I'll make a small PR that moves those guys to ShaderPanelsConfig.

PR #468, opened 2024-04-06
Audio Reactivity for Turbulent Noise Lines

zranger1:
- Level Reactivity uses bass energy to adjust the noise field generator and coordinate warping (it makes the lines slide and dance around.)
- Frequency Reactivity uses treble to control how much spare black "ink" is flying around in the wind (xpos/ypos), and also bounces extra light around in the spaces between the lines.


andrewlook (approved):
Love this pattern! merging now.

PR #470, opened 2024-04-07
cleanup: remove TD/Utils/Backup/ files (and add to .gitignore)

andrewlook:
cleaned out the `TD/Utils/Backup` dir and added it to gitignore, so we can use that dir locally without committing the files

I'm not sure if this change gets in the way of how we're using TD - happy to adjust this if there's a better way to go about this!


sina-cb (approved):
Thanks

sina-cb (commented):
Addressed these changes in this new PR and also added a few more changes to the Visualizer sketch
https://github.com/titanicsend/LXStudio-TE/pull/472

PR #471, opened 2024-04-07
Minor changes to NeonCells

zranger1:
- Commented out calls to logger.
- Using system RMS value instead of pattern local one, after checking that they're identical to at least 3 decimal places, and making sure the pattern still looks great.

(The tiny difference is likely because system value is calculated from the incoming stereo signal.  Since LX does this for us on every frame, we might as well save CPU time and use it.)

### DIFF
--- a/src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
+++ b/src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
@@ -300,7 +300,7 @@ public class ShaderPanelsPatternConfig {
 
       List<String> signalNames = Arrays.asList("control", "RMS", "filteredRms", "ControlledRMS", "sumTimeDiff");
       logger = new SignalLogger(signalNames, "Logs/signal_data.csv");
-      logger.startLogging(10);
+     // logger.startLogging(10);
     }
 
     @Override
@@ -329,12 +329,12 @@ public class ShaderPanelsPatternConfig {
     }
 
     private float getTimeDiff(float controlLevel) {
-      float rms = getRMS(eq.getSamples(), 0.5f);
+      float rms = eq.getRaw(); // getRMS(eq.getSamples(), 0.5f);
       float filteredRms = (float) rmsFilter.applyGaussianFilter(rms);
       float controlledRms = rms * controlLevel;
 
       sumTimeDiff += controlledRms;
-      logger.logSignalValues(Arrays.asList(controlLevel, rms, filteredRms, controlledRms, sumTimeDiff));
+      //logger.logSignalValues(Arrays.asList(controlLevel, rms, filteredRms, controlledRms, sumTimeDiff));
       return sumTimeDiff;
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -378,6 +378,8 @@ private float getRMS(float[] samples, float downsamplePercentage) {
       float meanOfSquares = sumOfSquares / downsampledLength;
 
       // Calculate the Root Mean Square (RMS)
+      float r = (float) Math.sqrt(meanOfSquares);
+      System.out.println("RMS: " + r + "Sys RMS: " + eq.getRaw());
```
andrewlook, line 382: whoops, we may want to comment this line out as well (thanks for being thorough and checking the RMS calculation!)

andrewlook (approved):

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
sina-cb: oop, sorry! (:

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -300,7 +300,7 @@ public NeonCells(LX lx) {
 
       List<String> signalNames = Arrays.asList("control", "RMS", "filteredRms", "ControlledRMS", "sumTimeDiff");
       logger = new SignalLogger(signalNames, "Logs/signal_data.csv");
-      logger.startLogging(10);
+     // logger.startLogging(10);
```
sina-cb, line 303: Thanks Jon.

I think we can completely remove the logger calls.

sina-cb (commented):
Addressed these changes in this new PR
https://github.com/titanicsend/LXStudio-TE/pull/472

But the new PR might have too much changes in one go, if that's the case, I can remove the changes from my new PR and just merge them with this PR

I just multiplied the RMS value by 100 to make the effect more visible with audio.

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
zranger1: Got rid of stray println() that I introduced -- wasn't getting called, but still... ick!

src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
zranger1: Logger calls removed.

zranger1:
@sina-cb, I left the x100 RMS multiplier out of this one.  Just tested w/a loud live techno set which will probably be comparable to Framework in signal level -- x1 had plenty of reactivity and looked great..  x100 was hard to control.  Are you seeing something different?

sina-cb (approved):

PR #473, opened 2024-04-07
Adding TdNdiPattern to allow sending mapped data from TouchDesigner to TE

sina-cb:
This PR adds a new pattern which is to allow us to receive mapped data from TouchDesigner and send them to TE.

This PR also changed the NDIEngine a little bit to match the channel name using "contains" check instead of equality. This is needed since TD prefixes the channel name with the computer name.

### DIFF
--- a/src/main/java/titanicsend/ndi/NDIEngine.java
+++ b/src/main/java/titanicsend/ndi/NDIEngine.java
@@ -49,7 +49,7 @@ public class NDIEngine extends LXComponent implements LXLoopTask {
    */
   public boolean connectByName(String sourceName,DevolayReceiver receiver) {
     for (DevolaySource source : sources) {
-      if (source.getSourceName().equals(sourceName)) {
+      if (source.getSourceName().contains(sourceName)) {
         receiver.connect(source);
         return true;
       }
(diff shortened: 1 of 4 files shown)


src/main/java/titanicsend/pattern/jon/NDIReceiverTest.java
```diff
@@ -9,7 +9,7 @@
  * Test class for NDIReceiver: displays video frames, and adds a second shader for
  * a simple edge detection effect (controlled by WOW2)
  */
-@LXCategory("AAA")
+@LXCategory("NDI")
```
zranger1, line 12: Ha!  Thanks for changing this!  I'd forgotten about it.

src/main/java/titanicsend/ndi/NDIEngine.java
```diff
@@ -49,7 +49,7 @@ public void printSourceNames() {
    */
   public boolean connectByName(String sourceName,DevolayReceiver receiver) {
     for (DevolaySource source : sources) {
-      if (source.getSourceName().equals(sourceName)) {
+      if (source.getSourceName().contains(sourceName)) {
```
zranger1, line 52: You've used NDI more than I have -- is there a reason I should know about to compare source name strings with ".contains" rather than ".equals"?

When testing this originally,the NDI SDK tools were giving me numbered channels, like  "Camera1","Camera2", etc., and I didn't want "Camera12" to be opened by a request for "Camera1", so ".equals" seemed the best choice.

zranger1 (approved):
Looks good!  Approved.

src/main/java/titanicsend/ndi/NDIEngine.java
zranger1: Yeah... the computer name.  That makes sense.  This is fine, since we're not likely to be running a bunch of nearly identically named channels, distinguished only by suffix, in the immediate future.

PR #474, opened 2024-04-08
Fixes for NdiEngine and TdNdiPattern

sina-cb:
This PR is addressing two bugs:
  1. NdiEngine is not checking if the source is null or not, and it can cause a crash when the NDI pattern is the first pattern loaded on LX.
  2. Showing the last frame on the TdNdiPattern if there's no new frames received on NDI. This helps remove a flicker issue when NDI is being published on a bad connection over network.


zranger1 (approved):
Looks good!  Thanks!

PR #478, opened 2024-04-11
Metallic waves reactivity

worldveil:


andrewlook (approved):
Tested locally, this version is tempo reactive (controlled by "freq") and incorporates max of bass/trebleRatio (controlled by "level")

PR #480, opened 2024-04-11
Reactivity for RadialSimplex

worldveil:


andrewlook (approved):
Tested locally and it works for me - pattern is also unchanged when levelReactivity=0

PR #481, opened 2024-04-11
tweak: add 'view' as a per-pattern control back into top 16

andrewlook:
per @worldveil 's request, having a per-pattern view option on the midifighter will likely be useful


worldveil (approved):

src/main/java/titanicsend/pattern/TECommonControls.java
```diff
@@ -377,9 +377,6 @@ protected void setRemoteControls() {
           getControl(TEControlTag.WOWTRIGGER).control,
           null,  // former shader-specific explode control
```
jkbelcher, line 378: This comma should be deleted, addressed in #485

PR #482, opened 2024-04-11
MetallicBitWave: fork of MetallicWave (bit explosion + treble effect)

andrewlook:


src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
```diff
@@ -403,6 +403,29 @@ protected void createShader() {
     }
   }
 
+  @LXCategory("DREVO Shaders")
+  public static class MetallicBitWaves extends ConstructedShaderPattern {
+    public MetallicWaves(LX lx) {
```
jkbelcher, line 408: Constructor name needs to be updated to match the class name.

jkbelcher: Fixed.

jkbelcher (approved):
Tested, now runs without error.

PR #484, opened 2024-04-11
Coachella Config

fakeplant:
- Edge switches
- Fix some panel striping instructions
- "Hide" all backside components (adding x in front of IP addresses, ex: x10.7.0.0)


jkbelcher (approved):
Tested, runs without error.

PR #486, opened 2024-04-11
Add MaxMSP patch for use with BLT

jkbelcher:
For use with BLT, as an alternative to the other max scripts.


jkbelcher:
Closing,  outdated.

PR #491, opened 2024-04-14
Avoid NullPointerException resulted from initial null variable

jkbelcher:
Fixes this error:

```
[LX 2024/04/13 16:28:47] Unexpected error in device loop titanicsend.pattern.sinas.TdNdiPattern: Cannot read the array length because "<local3>" is null
java.lang.NullPointerException: Cannot read the array length because "<local3>" is null
	at titanicsend.ndi.NDIEngine.connectByName(NDIEngine.java:51)
	at titanicsend.pattern.sinas.TdNdiPattern.onActive(TdNdiPattern.java:128)
	at heronarts.lx.pattern.LXPattern.onLoop(LXPattern.java:532)
```


zranger1 (approved):
Will work this in asap.

PR #492, opened 2024-05-01
Allow user to modify default values for pattern parameters

jkbelcher:
Adds a button on TEPattern to set all parameter defaults to current values.  Saves default values to the project file.

The button location is arbitrary. 
image

After clicking the SetDefaults button, subsequent presses of the Panic button will restore parameters to the new defaults instead of the original defaults.  Defaults are unique to each pattern instance.


jkbelcher:
Good idea on the visual feedback for whether something has changed.  Will tinker with options.

andrewlook (approved):
Tested locally and it's working for me! merging now.

PR #495, opened 2024-06-10
Minor audio fixes + shader cache disable.

zranger1:
Small PR that:
- prevents division by zero in audio processing code if no signal is present
- disables the shader precompiler on startup because macOS Sonoma doesn't support the required binary shader format.  (This will actually make startup a little faster on macOS. )
- fixes a couple of minor bugs resulting from last minute coding at the Coachella event.

Wanted to get this one in before @jkbelcher's big dynamic model PR changes everything.

### DIFF
--- a/src/main/java/titanicsend/pattern/TEAudioPattern.java
+++ b/src/main/java/titanicsend/pattern/TEAudioPattern.java
@@ -19,6 +19,10 @@ public abstract class TEAudioPattern extends TEPattern {
   // The GraphicMeter holds the analyzed frequency content for the audio input
   protected final GraphicMeter eq = lx.engine.audio.meter;
 
+  // Small offset for audio signals to avoid divide by zero in ratio calculations,
+  // which can cause (possibly dramatic) visual glitches if the audio drops out.
+  protected static final double levelOffset = 0.01;
+
   // Fractions in 0..1 for the instantaneous frequency level this frame.
   // If we find this useful and track many more bands, a collection of ratio
   // tracker objects would make sense.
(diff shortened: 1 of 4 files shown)


src/main/java/titanicsend/pattern/TEAudioPattern.java
```diff
@@ -80,16 +84,16 @@ protected void run(double deltaMs) {
    */
   protected void computeAudio(double deltaMs) {
     // Instantaneous normalized (0..1) volume level
-    volumeLevel = eq.getNormalizedf();
+    volumeLevel = eq.getNormalizedf() + levelOffset;
```
jkbelcher, line 87: This was a normalized value but now could be 1.01.  Is that bad?

zranger1: Well, i didn't see any during testing on some Coachella bro techno, even with input signal cranked to +4dbu (and amp turned ***way*** down)  Saw a lot of red lights on the meters, but patterns behaved well.

Still, best to leave normalized values normalized so nobody gets anything they're not expecting.  Fixed in latest commit.

jkbelcher (approved):
In general I feel like consumers of a parameter should take responsibility to avoid dividing by zero.  But I haven't been using this as actively so I'll let @zranger1 make the call here.

Minor notes:
-  For style consistency levelOffset could be all caps
- It could now be called a minimum instead of an offset

Otherwise looks fine to me.  Thanks!

zranger1:
Thanks for the review, @jkbelcher!!
Agree about the variable names, and will change them in next audio-related PR.

On where to check for divide-by-0:   In principle, I agree, but at this point we have patterns doing ad-hoc feature detection in many different ways.  Many of them involve dividing something by a volume metric, and it seems better to just keep them out of trouble. We can revisit this, and maybe replace it in patterns with a better "official" set of feature detectors soon.

PR #496, opened 2024-06-26
Adding the TE-Projection Touchdesigner App

sina-cb:
This PR adds some new changes to the ModelFileWriter to write out the mask PNG files along with UV coordinates.

This PR also adds some refactoring to the ModelFileWriter to make it easier to support boot up model updates.


zranger1 (changes requested):
This works very well!  Just for the sake of neatness, remove the unused imports, and I'll approve. 

And thanks for talking to projection people and figuring out what they needed!  In my corner of the universe, uv coordinates are always 0.0 to 1.0 floating point.   Never would've occurred to me that they'd want something different.

Projects/CoachellaMaster.lxp
jkbelcher, line 1: Let's not overwrite the CoachellaMaster project file.  It should be preserved for reference.  If you think the NDI stuff needs an example file just make a new one.  OTOH if a project file was just for testing let's avoid commiting them because it bloats the repo.  Those changes get pulled forever when someone runs `git clone`.

.gitignore
```diff
@@ -27,3 +27,4 @@ resources/shaders/cache/*.bin
 
 .devSwitch
 TD/Utils/Backup/
+TD/TE-Projection/Backup/
```
jkbelcher, line 30: Should this be a local gitignore or does the TD app generate this folder?

src/main/java/titanicsend/pattern/jon/ModelFileWriter.java
```diff
@@ -2,45 +2,165 @@
...
 
   public ModelFileWriter(LX lx) {
     super(lx, TEShaderView.ALL_POINTS);
     addCommonControls();
+  }
+
+  @Override
+  public void runTEAudioPattern(double deltaMs) {
+
+    if (doneWriting) {
```
jkbelcher, line 47: Thanks for moving this out of the `for` loop.

src/main/java/titanicsend/pattern/jon/ModelFileWriter.java
```diff
@@ -2,45 +2,165 @@
...
-      Files.writeString(path, "x,y,z\n", CREATE, APPEND);
+
+      // Write all points
+      // This is the main model and is required to be written out.
+      LXPoint[] all_points_array = lx.getModel().points;
+      writeModel(all_points_array, "all");
+
+      writeOptionalModels();
+
+      doneWriting = true;
```
jkbelcher, line 60: Maybe move this to the start of the `try` block?  If the file operation fails it's going to try again every frame which could be a problem.

src/main/java/titanicsend/pattern/sinas/TdNdiPattern.java
```diff
@@ -29,123 +30,125 @@
...
-    protected DevolayReceiver receiver = null;
-    private ByteBuffer frameData;
-
-    protected boolean lastConnectState = false;
-    protected long connectTimer = 0;
-    protected String channel_name = "TE_TD_Mapped";
-
-    public TdNdiPattern(LX lx) {
-        this(lx, TEShaderView.ALL_POINTS);
+  protected NDIEngine ndiEngine;
```
jkbelcher, line 33: Why did all this stuff get moved?  Was the formatter not run before?

jkbelcher (commented):
Given the size of these binaries I'm wondering if we should start a new repo for the TD app.

Also - the png files will change every time a panel is adjusted, right?  Which will happen every show.  Maybe we should exclude them from the repo and generate as-needed, given they're large files and almost guaranteed to be out of date.

jkbelcher:
Would like to get @zranger1 's eyes on the changes to ModelFileWriter and TdNdiPattern.

zranger1:
Both ModelFileWriter and TeNdiPattern are good to go!

TeNdiPattern code is super simple -- it's just 1:1 mapping incoming data to pixels, where the incoming "video" stream is really a vector of ordered pixels.

ModelFileWriter, maybe there are a couple more minor optimizations we could add, but it is fast enough, and works well.  I just tested with TouchDesigner, and it all looks great.

TDTest1

.gitignore
sina-cb: The TD app generates this. And I just added a new entry here following the gitignore that was already added a while ago. I think a local gitognore would be nice. Moved it.

src/main/java/titanicsend/pattern/jon/ModelFileWriter.java
sina-cb: Good point. moved it.

src/main/java/titanicsend/pattern/sinas/TdNdiPattern.java
sina-cb: yeah, I think the formatter wasn't run on this file before. it was okay, but to be consistent I did the formatting on all of the file.

Projects/CoachellaMaster.lxp
sina-cb: I must have forgotten to revert it back. I usually test on the default project file and revert before committing. I should ask you all how you set up a project to be default without changing the .lxpreference (or maybe it was another file) file on git

jkbelcher (commented):
My gut here is to gitignore the resources/model folder to reduce noise as these PNGs and CSVs files turn over.  They will be outdated every time a panel or edge moves which is all the time. @zranger1 what do you think?

jkbelcher:
BTW thanks for relocating the app and trimming this PR down to a much more manageable size.

src/main/java/titanicsend/pattern/jon/ModelFileWriter.java
```diff
@@ -2,45 +2,166 @@
...
-        } catch (IOException e) {
-          throw new RuntimeException(e);
-        }
+  private void writeOptionalModels() throws IOException {
+    // Write "panel" points only
+    List<LXPoint> panel_points_list = new ArrayList<>();
+    for (TEPanelModel panel : TEApp.wholeModel.getPanels()) {
+      panel_points_list.addAll(Arrays.asList(panel.points));
+    }
+    writeModel(panel_points_list, "panels");
```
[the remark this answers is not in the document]
jkbelcher: Edit: those collections are working fine.  I had the Mothership model loaded instead of TE.

sina-cb:
The follow on PR is here:
https://github.com/titanicsend/LXStudio-TE/compare/sinas/te_separate/panels?expand=1

jkbelcher (approved):

jkbelcher:
Approving for merge but there are unresolved conversations from that aren't visible because there was a force-push after they happened. Recommend not using force-push once things are under collaboration. I'll let you guys sort that out... @sina-cb @zranger1 can you mark those conversations as resolved?

sina-cb:
> This works very well! Just for the sake of neatness, remove the unused imports, and I'll approve.
> 
> And thanks for talking to projection people and figuring out what they needed! In my corner of the universe, uv coordinates are always 0.0 to 1.0 floating point. Never would've occurred to me that they'd want something different.

This should be done. I will reach out to Jon to resolve this.

zranger1 (approved):

jkbelcher:
Feel free to merge when ready.

sina-cb:
Yay! Thank you guys! :)

PR #497, opened 2024-07-03
OSC handler for audio stems

jkbelcher:
Adds a top-level component to monitor OSC messages for audio stem float values.  Stores the values in parameters where they can be used for mappings or by calling the static accessor e.g.:

`AudioStems.current.drums.getValue()`

image


jkbelcher:
Demo:
[](https://www.youtube.com/watch?v=wslrXm5dyXY)

worldveil:
Looks good to me. We can eventually do moving averages and historical normalization here too if we want!

Will shaders be able to access these values eventually?

jkbelcher:
> Looks good to me. We can eventually do moving averages and historical normalization here too if we want!

Exactly.  That's why I left a handler method for each stem.

> Will shaders be able to access these values eventually?

Java patterns can access them now through the static `AudioStems.current`.  For the final version we might want to route this stuff through an abstraction layer to keep flexibility for the end user.

worldveil (dismissed):

jkbelcher:
Replaced by #500

PR #498, opened 2024-07-09
Dynamic Model axis swap & texture mapping update

zranger1:
The Z and X axes were swapped in the change to dynamic model.   Patterns and shaders now know which model they're running with and adjust appropriately.   This was done with the following priorities:
- minimize performance impact
- make the test/swap code as simple and as easily removable as possible.

Even so, note that there is a small performance cost to supporting both models, so we should remove the test/swap code once we fully commit to the new model.

2d and 3d texture mapping (the ModelBender thing) will also now work correctly with both static and dynamic models.

### DIFF
--- a/src/main/java/titanicsend/model/TEWholeModelDynamic.java
+++ b/src/main/java/titanicsend/model/TEWholeModelDynamic.java
@@ -15,6 +15,7 @@ import heronarts.lx.model.LXModel;
 import heronarts.lx.model.LXPoint;
 import heronarts.lx.transform.LXVector;
 import titanicsend.dmx.model.DmxModel;
+import titanicsend.pattern.jon.ModelBender;
 
 public class TEWholeModelDynamic implements TEWholeModel, LX.Listener {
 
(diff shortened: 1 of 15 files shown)


src/main/java/titanicsend/model/TEWholeModelDynamic.java
```diff
@@ -258,20 +259,25 @@ public void modelGenerationChanged(LX lx, LXModel model) {
     this.mutableChildren.addAll(this.panelModels.keySet());
     this.children = this.mutableChildren.toArray(new LXModel[0]);
 
+    // adjust model geometry for easy texture mapping in views
+    ModelBender mb = new ModelBender();
+    mb.adjustEndGeometry(this, this.lx.getModel());
+    mb.restoreModel(this, this.lx.getModel());
```
andrewlook, line 265: Just curious - my understanding from how this worked in the static model was that we'd adjust the end geometry, then define the views (using the adjusted geometry), then restore the model - so there was a bunch of code in between the two calls to `ModelBender`. For instance, here's the old snippet of code I was looking at:

```java
    // adjust model geometry for easy texture mapping in views
    ModelBender mb = new ModelBender();
    mb.adjustEndGeometry(this);

    // Create LXViews once
    for (ViewDefinition v : views) {
      if (v.viewEnabled && (v.viewSelector != null) && !v.viewSelector.isEmpty()) {
        addGlobalView(lx, v);
      } else {
        LX.warning("View Definition is not a filter.");
      }
    }

    // restore original model geometry
    mb.restoreModel(this)
```

Did I understand that wrong? In other words, after these two calls happen in sequence, are the points in the model "adjusted" or are they restored back to the original coordinates?

src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -225,4 +225,13 @@ public List<LXPoint> getGigglePixelPoints() {
     }
     return rv;
   }
+
+  /**
+   * utility method for use during the static-to-dynamic model transition.
+   * IMPORTANT:  There is a performance cost to this, so it should be
+   * removed when we no longer need to support the static model.
```
andrewlook, line 232: thanks for calling out that this is the source of the performance cost 🙏

andrewlook (approved):
This is looking great!

Not blocking, but just to solidify my understanding: I added a question inline about how the new model uses `adjustEndGeometry`/`restoreModel` (where are the views getting defined / how are they getting affected by this?).

I tested:
- before this PR (most GLSL shaders looked funky)
- after this PR, with static model
- after this PR, with dynamic model (most GLSL shaders looked great)

So overall I'm ready to merge this so we can keep trucking on the dynamic model migration 💪 

One side note (outside the scope of this PR but just as FYI): ArcEdges and EdgeFall were totally black before this PR, now they show up as very small in the middle of the top two panels. I wasn't expecting them to be fixed by this PR, but just wanted to note that in case it makes sense to you at all (are they normalizing coords differently when they pass in that list of edges?)

src/main/java/titanicsend/model/TEWholeModelDynamic.java
zranger1: When these two calls happen in sequence, we're left with a TEWholeModelDynamic that has 'adjusted' normalized coordinates and 'original' real-space coordinates.   At this point, it probably only applies to the "default" view.

View creation with the dynamic model seems to be internal to Chromatik now.  And things get even weirder should we decide to include the FoH car and TE in the same project file.   I'm not yet sure exactly how it works or what needs to change, but it's likely to be one of those "interesting" solutions.

The ModelBender math will stay the same though, so the final views fix can be its own PR.   IMO, this one's worth taking now just for the axis swap, and because it unblocks new pattern and project file development using the dynamic model.

andrewlook:
Excellent, thanks for explaining! It makes sense now to me that it’s only
modifying the normalized coords. Agreed that it’s worth merging for the
axis swap alone and dealing with the views later (IMHO default view is top
priority, edges are second, all other views are considerably lower on my
list)

PR #499, opened 2024-07-09
tweak: CoachellaMaster.lxp show file + run config using dynamic model

andrewlook:
Just sharing this so we have an easy way to do our testing using a dynamic model


zranger1 (approved):
Yes!  This will be super handy for testing.  (and so far seems to mostly behave itself!)

PR #501, opened 2024-07-09
Fix edge.connectedPanels

jkbelcher:
Fixes string comparison issue for TEEdgeModel.connectedPanels.  Tested as working with ArcEdges, EdgeFall, and EdgeRunner.


zranger1 (approved):
Awesome!   Thanks!!

PR #503, opened 2024-07-10
Mothership fixtures in progress

jkbelcher:
Adds a Mothership fixture and project file so people can start testing patterns against the approximate car shape.

Note this is in-progress and there will be further changes to LED counts, positions, window shapes, tags, and outputs.


andrewlook (approved):
awesome, thanks for this!

non-blocking question: was commenting this line out related to Mothership? `lx.registry.removePattern(PlanesPattern.class)`

jkbelcher:
Yes, I was using it to check fixture orientation/alignment against the assembled windows.  We can take it at some point before showtime.

andrewlook:
Thanks for explaining! I just was curious about it, no need to change

PR #505, opened 2024-07-16
Color Controls: LXComponent, with UI embedded in right performance pane

andrewlook:
## TLDR

Similar to the idea of "cue" and "aux", I added two independent sets of color controls, so we can tweak and prepare two color palettes alongside each other.

image

## Details

The very top row has a view of Chromatik's global active palette. Below that are two sets of color controls.

Each set of color controls has:
- **Sliders for hue, saturation, and brightness**
- **Knob for "Color Palette Type" selection**, with the following options:
  - `MONOCHROME`, `ANALOGOUS` (+/- 30 on hue wheel), `COMPLEMENTARY` (+180), `TRIADIC` (+/- 120), `SPLIT_COMPLEMENTARY` (+150, +210), and `GOLDEN_RATIO_CONJUGATE` (`hue * 0.618033`, shout-out to @zranger1  for that idea!)
- **Miniature color palette preview**, with a larger rectangle for the color selected via hue/saturation sliders, and two smaller squares for the colors selected by the palette type knob (I wanted to emphasize that the hue/sat sliders are manipulating one color, and the other colors are dependent - open to suggestions on how this could work!)
- **SWAP** button, to push the cue'd up colors live into Chromatik's global active palette.

### Color Management UX

The approach shown here is most flexible, in that it allows selection of any hue, but enforces how colors 2 and 3 are selected.

1. (semi-flexible) have a fixed allowlist of "main colors", which can be scrolled around in hue order (instead of a free-selection hue slider). Then still allow color theory / palette type combinations to choose colors 2 and 3.
2. (inflexible) have a fixed allowlist of full color combinations palettes. Allowing navigation of them from performance view, (and ordering them according to the hue wheel) would still be a small improvement on Chromatik's global palette, if we want to further restrict the color selection.

### Palette Types

- [ ] consider removing `COMPLEMENTARY` colors, since the contrast between two opposite colors might be too aggressive / unsightly

### UI Elements

image

This is very POC/prototype stage - these are meant to be midi-mapped. I don't love the hue/sat sliders, but it seems workable. brightness might not be needed, since we have so many other places to control that. 

Most ideal UI would be something more like this:

Screen Cast 2024-07-15 at 10 16 25 PM

### UI Placement

**Design View:** placed in `leftPane.global`, above the system's default palette manager.

image

**Performance View**: in the right-hand pane (TBD on final placement, current state is "good enough" to test but can be refined further, as described below).

image

`LXStudio.UI.rightPane` is private, so I had to use reflection to grab it and inject the palette manager UI. 

Note: The ordering is a little funky. The new UI (red box) is added at boot time, before the aux controls get added to the rightPane (yellow box). I'd prefer to place it down in the bottom right of the screen, replacing `rightPerformancePalette` (green box).

> _To do this, I think I need to make the class implement `UIPerformanceTool`, and then use reflection to forcibly override `LXStudio.UI.mainContext.rightPerformancePalette`, since there's some absolute positioning in the UI and I couldn't figure out how to add another top-level module between `rightPane` and `rightPerformancePalette`. I think this is doable, but wanted to iterate on the core UX before fiddling too much with chromatik monkey-patching._

### Internals

The two sets of color controls take advantage of system palette slots to store their colors, naming the top 2 palettes "CUE" and "AUX" respectively (creating them if they don't exist). NOTE: the logic around creating them is a bit funky, and I'm not sure how defensive to be - if the palettes get deleted mid-show, should they get re-created instantly? etc. I'd love feedback / ideas / QA help on this 🙏 

image

`ColorPaletteManager` is implemented pretty naively, and there's probably some things that would work better if they followed LX/Chromatik idioms better. 

- [ ] `ColorPaletteManager.onParameterChanged()` is called for any parameter change at all. Could be simpler if it used listeners (`listener(hue / saturation / brightness) -> updates color1`, `listener(palette type / color1) -> updates color2 + color3`)
- [ ] I'm unsure if it's redundant to use `color1,color2,color3` as member fields of the class, when we're also using the "CUE/AUX" palettes in the system palette. Should I drop the idea of manipulating system palettes for the, if it's redundant?
- [ ] I tried to display the "CUE/AUX" system palettes alongside the global active palette at the top of the UI, but the LXSwatch displays would be frozen as red even as the real underlying swatches were being updated (this worked in a previous branch `look/color-controls-mockup`, where this UI was implement as `UIDeviceControls` of an `LXEffect`). Set `UIColorPaletteManager.DISPLAY_MANAGED_SWATCH_ROWS = true` to observe this behavior.

Chromatik UI housekeeping I'm unfamiliar with: 

- [x] Is there anything I need to do in terms of saving custom UI state to LXP? I believe the `addParameter` calls in `ColorPaletteManager` should save the parameter states, but I'm not sure if I'm missing anything.
- [ ] Is there anything I need to do with `dispose` methods to clean up listeners? The only listeners I'm currently using are the toggle listeners for the "SWAP" buttons used to push the selected palettes live.


jkbelcher:
Hey lots of discussion to be had from this. Overall the direction is great.  Big things I like:

- Forces good color combinations through fixed offsets
- Adds visual preview of the main color & offsets while still being wide open to any main color

jkbelcher:
An upgrade from displaying the knob, maybe not top priority, would be to display all the Color Palette Types.  Right now you're choosing between a set of things but can only see one of the things at a time.

jkbelcher:
>  if the palettes get deleted mid-show, should they get re-created instantly?

 Yes.  You'll have to make sure it works around file close/open too.  Don't create them if they're disposing for a file close.  Do create them if a file opened that didn't have them.

jkbelcher:
The names `CUE` and `AUX` already have a distinct role as cue = main preview (and left side preview in Performance mode), and Aux = right side preview in Performance mode.  May be good to have distinct names to keep people from thinking they're locked together. `ColorA` and `ColorB`?

src/main/java/titanicsend/ui/effect/UIColorPaletteManager.java
```diff
@@ -0,0 +1,307 @@
...
+      horizontalBreak(ui, this.width).addToContainer(this);
+
+      buildPaletteSelectionRow(auxMgr).addToContainer(this);
+
+      horizontalBreak(ui, this.width).addToContainer(this);
+
+      buildColorSlidersRow(auxMgr).addToContainer(this);
+
+      this.addListener(
+          auxMgr.toggleCue,
```
jkbelcher, line 87: This can be self-contained in the manager class.  Such as from the constructor:
`this.toggleClue.addListener(...)`

andrewlook:
> The names `CUE` and `AUX` already have a distinct role as cue = main preview (and left side preview in Performance mode), and Aux = right side preview in Performance mode. May be good to have distinct names to keep people from thinking they're locked together. `ColorA` and `ColorB`?

very good point, i realized later that cue/aux were already used but I wanted a similar concept of "something that's on deck but the audience can't see". `A` and `B` sound better 👍

andrewlook:
> > if the palettes get deleted mid-show, should they get re-created instantly?
> 
> Yes. You'll have to make sure it works around file close/open too. Don't create them if they're disposing for a file close. Do create them if a file opened that didn't have them.

This sounds sensible, I'll come up with a set of test cases i can run through manually.

One clarification: "if they're disposing for a file close" - how would I be able to distinguish whether this is happening?

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    cueSwatch.label.setValue(this.name);
+    if (cueSwatch.colors.size() < LXSwatch.MAX_COLORS) {
+      for (int i = cueSwatch.colors.size(); i < LXSwatch.MAX_COLORS; ++i) {
+        cueSwatch.addColor();
+      }
+    }
+    return cueSwatch;
+  }
+
+  public void updateSwatches() {
```
jkbelcher, line 148: The CUE swatch already has the current colors.  Something like
`this.managedSwatch.recall.trigger()`

...Should move to that swatch while honoring the Color Palette transition controls.

jkbelcher (changes requested):
Only had a few minutes but here are a couple things to dig on!

andrewlook:
> An upgrade from displaying the knob, maybe not top priority, would be to display all the Color Palette Types. Right now you're choosing between a set of things but can only see one of the things at a time.

Ooh yeah great idea! I was thinking it would be cool to somehow do it on the APC mini potentially, so the button grid could display the palettes 🤔

jkbelcher:
`Is there anything I need to do in terms of saving custom UI state to LXP? I believe the addParameter calls in ColorPaletteManager should save the parameter states, but I'm not sure if I'm missing anything.`
Technically not saving the UI.  You're saving the component underlying the UI.

By extending `LXComponent` and calling `lx.engine.registerComponent(...)`, all the parameters you called `addParameter()` for will be saved/loaded.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  protected void setColorAtPosition(LXSwatch swatch, DmxColorModulator.ColorPosition colorPosition, int color) {
+    if (colorPosition != DmxColorModulator.ColorPosition.NONE) {
+      swatch.getColor(colorPosition.index).primary.setColor(color);
+      swatch.getColor(colorPosition.index).mode.setValue(LXDynamicColor.Mode.FIXED);
+    }
+  }
+
+  public void onParameterChanged(LXParameter parameter) {
```
jkbelcher, line 165: `@Override`

jkbelcher, line 165: in the onParameterChanged() method, standard practice is to check the parameter to make sure it's what you're trying to monitor.

```
// Did anything change that would affect output color?
if (parameter == this.hue || parameter == this.saturation || parameter == this.brightness || parameter == this.currPaletteType ) {
  // yep! Update the cue swatch.
}
```

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    // ensure there are at least enough swatches in the global palette list to fetch
+    // the correct index for this "managed swatch". TODO: listen in case it's deleted?
+    if (lxPalette().swatches.size() <= (this.index + 1)) {
+      lxPalette().saveSwatch();
+    }
+    // run `getManagedSwatch` to ensure the swatch is created, has the right name/num colors
+    this.managedSwatch = managedSwatch();
+  }
+
+  protected LXPalette lxPalette() {
```
jkbelcher, line 133: I think you did this to shorten the characters in all the palette calls.  Opinion here but I think this reduces readability... when I see lxPalette() in code I don't know if it's simple or complicated behind it.  And pulling it into a `protected` method implies a child class might override it.  And that context is beyond what I can infer from code.

jkbelcher (changes requested):
Hey looking great so far.  Tried to clarify some details so you can keep momentum on it.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+        cueSwatch.addColor();
+      }
+    }
+    return cueSwatch;
+  }
+
+  public void updateSwatches() {
+    updateSwatches(lxPalette().swatch);
+  }
+  protected void updateSwatches(LXSwatch swatch) {
```
jkbelcher, line 151: Nitpick: missing a newline

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+      new CompoundParameter("B", 100, 0, 100)
+          .setUnits(CompoundParameter.Units.PERCENT)
+          .setPolarity(CompoundParameter.Polarity.BIPOLAR)
+          .setDescription("Sets the amount to increase or decrease brightness");
+
+  public final ColorParameter color1 = new ColorParameter("Color1", LXColor.BLACK);
+  public final ColorParameter color2 = new ColorParameter("Color2", LXColor.BLACK);
+  public final ColorParameter color3 = new ColorParameter("Color3", LXColor.BLACK);
+  // update this so we know whether to re-render the palette
+  public PaletteType currPaletteType = PaletteType.TRIADIC;
```
jkbelcher, line 61: No need to monitor the last known value here.  See notes below for using onParameterChanged() to monitor for any changes to the seed parameters, after which you can update the other colors.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    SPLIT_COMPLEMENTARY,
+    COMPLEMENTARY,
+    TRIADIC,
+    // TODO(look): Jon Idea of Golden Ratio Conjugate (mult hue by )
+  }
+
+  public final EnumParameter<PaletteType> paletteType =
+      new EnumParameter<PaletteType>("Color Position", PaletteType.TRIADIC)
+          .setDescription(
+              "Destination color position (1-based) in the global palette current swatch");
```
jkbelcher, line 95: Copy/paste description doesn't match

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    ANALOGOUS,
+    GOLDEN_RATIO_CONJUGATE,
+    SPLIT_COMPLEMENTARY,
+    COMPLEMENTARY,
+    TRIADIC,
+    // TODO(look): Jon Idea of Golden Ratio Conjugate (mult hue by )
+  }
+
+  public final EnumParameter<PaletteType> paletteType =
+      new EnumParameter<PaletteType>("Color Position", PaletteType.TRIADIC)
```
jkbelcher, line 93: "Palette Type" instead of "Color Position"?

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+          .setDescription(
+              "Destination color position (1-based) in the global palette current swatch");
+
+  public static final String DEFAULT_NAME = "CUE";
+  public static final int DEFAULT_INDEX = 0;
+
+  private final String name;
+  private final int index;
+
+  public LXSwatch managedSwatch;
```
jkbelcher, line 103: This should be `private`, don't want another class changing it.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    // run `getManagedSwatch` to ensure the swatch is created, has the right name/num colors
+    this.managedSwatch = managedSwatch();
+  }
+
+  protected LXPalette lxPalette() {
+    return this.lx.engine.palette;
+  }
+
+  public LXSwatch managedSwatch() {
+    LXSwatch cueSwatch = lxPalette().swatches.get(this.index);
```
jkbelcher, line 138: I think this is fine for now and we could leave deletion-detection for a later PR.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+
+  private void updateColors2And3(int color1) {
+    this.currPaletteType = this.paletteType.getEnum();
+
+    float hue = this.color1.hue.getValuef();
+    float saturation = this.color1.saturation.getValuef();
+    float brightness = this.color1.brightness.getValuef();
+    int color2 = color1;
+    int color3 = color1;
+    if (this.paletteType.getEnum() == PaletteType.MONO) {
```
jkbelcher, line 186: `switch (this.paletteType.getEnum()) {`

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    if (color1 != this.color1.getColor() || this.currPaletteType != this.paletteType.getEnum()) {
+      this.color1.setColor(color1);
+      this.currPaletteType = this.paletteType.getEnum();
+      updateColors2And3(color1);
+      updateSwatches(managedSwatch());
+    }
+  }
+
+  private void updateColors2And3(int color1) {
+    this.currPaletteType = this.paletteType.getEnum();
```
jkbelcher, line 179: Not needed

src/main/java/titanicsend/ui/effect/UIColorPaletteManager.java
jkbelcher, line 1: This is is package `titanicsend.ui.effect` but seems like `titanicsend.ui.color` would be more appropriate?

jkbelcher (changes requested):
More comments

src/main/java/titanicsend/ui/effect/UIColorPaletteManager.java
andrewlook: ah thanks, that was left over from a refactor - updated now.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+import heronarts.lx.parameter.LXParameter;
+import titanicsend.modulator.dmx.DmxColorModulator;
+
+@LXCategory(LXCategory.COLOR)
+@LXComponentName("Look Color Palette")
+public class ColorPaletteManager extends LXComponent {
+
+  public final CompoundParameter hue =
+      new CompoundParameter("H", 0, 0, 360)
+          .setPolarity(CompoundParameter.Polarity.BIPOLAR)
```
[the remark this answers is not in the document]
andrewlook: done

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+  public final CompoundParameter saturation =
+      new CompoundParameter("S", 100, 0, 100)
+          .setUnits(CompoundParameter.Units.PERCENT)
+          .setPolarity(CompoundParameter.Polarity.BIPOLAR)
+          .setDescription("Sets the amount to increase or decrease saturation");
+
+  public final CompoundParameter brightness =
+      new CompoundParameter("B", 100, 0, 100)
+          .setUnits(CompoundParameter.Units.PERCENT)
+          .setPolarity(CompoundParameter.Polarity.BIPOLAR)
```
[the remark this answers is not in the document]
andrewlook: done

src/main/java/titanicsend/ui/effect/UIColorPaletteManager.java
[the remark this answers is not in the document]
andrewlook: sg, done

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return cueSwatch;
+  }
+
+  public void updateSwatches() {
+    updateSwatches(lxPalette().swatch);
+  }
+  protected void updateSwatches(LXSwatch swatch) {
+    // Send to target color in global palette
+    setColorAtPosition(swatch, this.color1Pos.getEnum(), this.color1.getColor());
```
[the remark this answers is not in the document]
andrewlook: good call, done

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,209 @@
...
+    setColorAtPosition(swatch, this.color2Pos.getEnum(), this.color2.getColor());
+    setColorAtPosition(swatch, this.color3Pos.getEnum(), this.color3.getColor());
+  }
+
+  protected void setColorAtPosition(LXSwatch swatch, DmxColorModulator.ColorPosition colorPosition, int color) {
```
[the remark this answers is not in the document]
andrewlook: i just used TEColorType and got the swatch index inside this function 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: done

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: done

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: sounds good - i tested this, and when you delete the "cue" palette, the next time you update any of the color controls it takes the top palette and renames it to "cue" (or second palette for "aux").

I noticed though that if you delete all of the swatches, it would crash - so just to be safe I moved a couple lines of code down from the constructor to ensure enough swatches exist.

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: done

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: fixed

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: good call, done

src/main/java/titanicsend/ui/effect/UIColorPaletteManager.java
andrewlook: this is super helpful, thank you!

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: I replaced these with listeners on the individual params 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: made this change. Just FYI, it doesn't actually use the transition time even when transitions are enabled in the global palette pane.

src/main/java/titanicsend/ui/color/UIColorPaletteManager.java
```diff
@@ -0,0 +1,266 @@
...
+    LXStudio.UI.MainContext mainContext = getLXUIMainContext(ui);
+    UI2dContainer parentContainer = mainContext.rightPerformance;
+    float xOffsetWithinParent = 14;
+    new UIColorPaletteManager(ui, cueMgr, auxMgr, parentContainer.getContentWidth() - 28, xOffsetWithinParent)
+        .addToContainer(parentContainer, parentContainer.getChildren().size()-1);
+  }
+
+  private static LXStudio.UI.MainContext getLXUIMainContext(LXStudio.UI ui) {
+    try {
+      // use reflection to access the private field `mainContext` in `LXStudio.UI`
```
jkbelcher, line 192: We don't want to be doing this but I'll leave it for now until we can establish a proper method in upstream.

src/main/java/titanicsend/ui/color/UIColorPaletteManager.java
```diff
@@ -0,0 +1,266 @@
...
+    new UIColorPaletteManager(ui, cueMgr, auxMgr, parentContainer.getContentWidth(), xOffsetWithinParent)
+        .addToContainer(parentContainer, 3);
+  }
+
+  public static void addToRightPerformancePane(LXStudio.UI ui, ColorPaletteManager cueMgr, ColorPaletteManager auxMgr) {
+    LXStudio.UI.MainContext mainContext = getLXUIMainContext(ui);
+    UI2dContainer parentContainer = mainContext.rightPerformance;
+    float xOffsetWithinParent = 14;
+    new UIColorPaletteManager(ui, cueMgr, auxMgr, parentContainer.getContentWidth() - 28, xOffsetWithinParent)
+        .addToContainer(parentContainer, parentContainer.getChildren().size()-1);
```
jkbelcher, line 187: Ideally [later] we could get this above the existing Color Palette in the right performance pane.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+    addParameter("pushSwatch", this.pushSwatch);
+    addParameter("pinSwatch", this.pinSwatch);
+
+    this.pushSwatch.addListener(pushListener);
+    this.hue.addListener(colorListener);
+    this.saturation.addListener(colorListener);
+    this.brightness.addListener(colorListener);
+    this.color1.addListener(paletteListener);
+    this.paletteStrategy.addListener(paletteListener);
+  }
```
jkbelcher, line 149: Need to initialize color1 in the constructor to the current values of hue/sat/bright.  Otherwise your color1 is black until a component is modified, and if pinned it pushes black to palette.

Projects/TestColorPaletteManager.lxp
jkbelcher, line 1: This looks like a test file not needed in the repo. ?

src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -159,14 +162,17 @@ public Plugin(LX lx) {
...
 
-      //      lx.ui.preview.addComponent(visual);
-      //      new TEUIControls(ui, visual,
-      // ui.leftPane.global.getContentWidth()).addToContainer(ui.leftPane.global);
       this.dmxEngine = new DmxEngine(lx);
       this.ndiEngine = new NDIEngine(lx);
       this.glEngine = new GLEngine(lx,staticModel);
 
       lx.engine.registerComponent("audioStems", this.audioStems = new AudioStems(lx));
+      lx.engine.registerComponent("paletteManagerA", this.paletteManagerA = new ColorPaletteManager(lx, "SWATCH A", 0));
```
jkbelcher, line 170: Simplify to `new ColorPaletteManager(lx)` ?

jkbelcher (changes requested):
Looking much better!  Mainly nitpicky details and style comments now.

Projects/CoachellaMaster.lxp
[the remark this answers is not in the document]
andrewlook: no, thanks for catching - reverted them using:

```sh
git checkout main

cp Projects/CoachellaMaster.lxp ~/Desktop

git checkout look/color-controls-device

cp ~/Desktop/CoachellaMaster.lxp Projects
```

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
+/**
+ * Copyright 2013- Mark C. Slee, Heron Arts LLC
```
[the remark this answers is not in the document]
andrewlook: removed 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+    this.pushSwatch.addListener(pushListener);
+    this.hue.addListener(colorListener);
+    this.saturation.addListener(colorListener);
+    this.brightness.addListener(colorListener);
+    this.color1.addListener(paletteListener);
+    this.paletteStrategy.addListener(paletteListener);
+  }
+
+  @Override
+  public void dispose() {
```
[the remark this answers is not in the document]
andrewlook: ok sounds good. personally i like having the teardown live near the setup, but if it's a style thing that works for me - just changed it

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+
+  public final CompoundParameter saturation =
+      new CompoundParameter("S", 100, 0, 100)
+          .setUnits(CompoundParameter.Units.PERCENT)
+          .setDescription("Sets the amount to increase or decrease saturation");
+
+  public final CompoundParameter brightness =
+      new CompoundParameter("B", 100, 0, 100)
+          .setUnits(CompoundParameter.Units.PERCENT)
+          .setPolarity(CompoundParameter.Polarity.BIPOLAR)
```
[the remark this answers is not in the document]
andrewlook: removed 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+   */
+  private final LXParameterListener paletteListener = (p) -> {
+    updateColors2And3();
+    updateManagedSwatch();
+    if (this.pinSwatch.isOn()) {
+      pushToActiveSwatch();
+    }
+  };
+
+  public final BooleanParameter pushSwatch =
```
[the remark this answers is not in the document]
andrewlook: whoa cool! thanks for the trick 🙌

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+      new BooleanParameter("Push Swatch", false)
+          .setDescription("Push the managed swatch to the global active swatch");
+
+  public final BooleanParameter pinSwatch =
+      new BooleanParameter("Pin Swatch", false)
+          .setDescription("Pin the managed swatch to the global active swatch, " +
+              "so any changes are reflected in the global palette immediately.");
+
+  private final LXParameterListener pushListener = (p) -> this.pushToActiveSwatch();
```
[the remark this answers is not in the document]
andrewlook: gotcha, thanks for explaining!

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  public ColorPaletteManager(LX lx, String swatchName, int swatchIndex) {
+    super(lx);
+
+    this.swatchName = swatchName;
+    this.swatchIndex = swatchIndex;
+
+    // ensure the swatch is created, has the right name/num colors
+    this.managedSwatch = managedSwatch();
```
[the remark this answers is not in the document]
andrewlook: ah yes, good catch. fixed.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+  }
+
+  private void updateManagedSwatch() {
+    this.managedSwatch = managedSwatch();
+    setColorAtPosition(this.managedSwatch, TEColorType.PRIMARY, this.color1.getColor());
+    setColorAtPosition(this.managedSwatch, TEColorType.SECONDARY, this.color2.getColor());
+    setColorAtPosition(this.managedSwatch, TEColorType.SECONDARY_BACKGROUND, this.color3.getColor());
+  }
+
+  private void setColorAtPosition(LXSwatch swatch, TEColorType teColorType, int color) {
```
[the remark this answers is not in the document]
andrewlook: done 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+    swatch.getColor(teColorType.swatchIndex()).primary.setColor(color);
```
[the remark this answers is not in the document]
andrewlook: done 👍

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+        color2 = LXColor.hsb(hue + 30, saturation, brightness);
+        color3 = LXColor.hsb(hue - 30, saturation, brightness);
+        break;
+      default:
+        throw new RuntimeException("Unknown PaletteStrategy: " + this.paletteStrategy.getEnum());
+    }
+    this.color2.setColor(color2);
+    this.color3.setColor(color3);
+  }
+}
```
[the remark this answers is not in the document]
andrewlook: this is good feedback, thanks!

Projects/TestColorPaletteManager.lxp
andrewlook: thanks, removed. I also added a line to `.gitignore` for `Projects/Test*.lxp`, since this is probably a relatively common case

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: i made this change - though when i have an LXP file open, save it, and re-open it, my color1 seems to have been reloaded from disk properly. The only case where I think it'd be unset is when a new LXP file is created, is that right? Just wanted to solidify my understanding of this.

src/main/java/heronarts/lx/studio/TEApp.java
andrewlook: done

src/main/java/titanicsend/ui/color/UIColorPaletteManager.java
andrewlook: yeah, this is the only way to do what we want to do here. Would really love if this wasn't private upstream, lmk if there's anything I can do to help there!

src/main/java/titanicsend/ui/color/UIColorPaletteManager.java
andrewlook: Ideally yes. I had some trouble navigating how LX UI works, particularly when absolute positioning is involved (which it seems to be for the top-level panes). One idea I had was to outright replace the existing performance color palette (since we can fairly easily show a superset of the same information).

As far as I understand, we'd either have to:

(a) have `UIColorPaletteManager` extend `UIPerformancePalette`, so that we could hot-swap it into this list:

```java
    public class MainContext extends UI2dContext {
      public final UILeftPane leftPane;
      public final UIRightPane rightPane;
      public final UIPerformancePane leftPerformance;
      public final UIPerformancePane rightPerformance;
      public final UIPerformanceAudio performanceAudio;
      public final UIPerformancePalette performancePalette; // <-- 
```

or (b) somehow figure out how to fit a new top-left UI pane so that the right panes resize themselves appropriately.

I think option (a) would be fairly easy to try, but yeah I'd rather try it in a separate PR once we have this one merged and can test the basic plumbing.

src/main/java/titanicsend/color/ColorPaletteManager.java
```diff
@@ -0,0 +1,234 @@
...
+
+  private void updateColors2And3() {
+    float hue = this.color1.hue.getValuef();
+    float saturation = this.color1.saturation.getValuef();
+    float brightness = this.color1.brightness.getValuef();
+    int color2;
+    int color3;
+    switch(this.paletteStrategy.getEnum()) {
+      case MONO:
+        color2 = LXColor.BLACK;
```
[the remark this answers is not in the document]
andrewlook: My assumption here was to mirror how our monochromatic palettes have worked in the past, with one color entry and the remaining ones black:

image

Now that I think about it, if we're doing the per-pattern gradient offset knob, it'd be easier not to check for black and disable gradient offset - so I'll go ahead and change this.

Projects/CoachellaMaster.lxp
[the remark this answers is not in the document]
jkbelcher: I looked up how to do this in one branch:

`git log`
(find commit hash before file was modified)
`git checkout <commit-hash> -- <file-path>`
(now you have the old version of the file)
```
git add <file-path>
git commit -m "Revert <file-path>"
```

src/main/java/titanicsend/color/ColorPaletteManager.java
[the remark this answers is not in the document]
jkbelcher: Yeah I always found those singles annoying in their different behavior.

src/main/java/titanicsend/color/ColorPaletteManager.java
jkbelcher: Yes, or at application start the LXP opened was before this component existed.

jkbelcher:
For a future PR, the PALETTE MANAGER could be shortened.  Also the ui boxes init to black:

image

jkbelcher (approved):
Solid round of work right there!

src/main/java/titanicsend/color/ColorPaletteManager.java
andrewlook: makes sense, thank you

PR #506, opened 2024-07-16
Mothership fixtures: window shapes and positions

jkbelcher:
Mothership window fixture refinements:
- Rounded corners on all windows
- Correct pixel counts on all windows
- Pixel locations should be "about" perfect
- Outputs to controller are testing and working

image

You may notice the start/end corner is not rounded.  These will likely stay as-in until the car is built and we see what installation anomalies we need to handle in the fixture files.  For now these straight ends make it easy to "push" the strip around the window by removing pixels from the front and adding them to the end.

I'm very happy with the new mechanism of strip placement.  It takes advantage of the perfect strip position relative to the CAD model by combining:
- Length of edges between faces, measured from CAD.
- Inset of the LED strip from the edges.
- Radius of the corners

This means the window placement should be perfect without any x/y adjustment.  To demonstrate I added a strip tracing each edge to the config files. You can see how sharp it looks:

image

As a bonus, the inset can be changed in one (well, two) places which will move the points closer to the edge of the window without affecting the # of pixels on each edge and corner.  This is a nice safety net if the "perceived" location of the pixel is outside of the "actual" location due to the polycarb diffuser.  In which case adjusting the `inset` parameter will look like this:

image

### DIFF
--- a/Fixtures/Mothership/window/RoundedQuad.lxf
+++ b/Fixtures/Mothership/window/RoundedQuad.lxf
@@ -1,26 +1,28 @@
 {
-  /* Triangle with Rounded Corners
-     Common geometry for triangular windows on TE Mothership */
+  /* Quadrilateral with Rounded Corners
+     Common geometry for 4-sided windows on TE Mothership */
 
-  label: "RoundedTriangle",
-  tags: [ "roundedtriangle" ],
+  label: "RoundedQuad",
+  tags: [ "roundedquad" ],
 
   parameters:  {
     # TE defaults. Can be overridden but likely do not need to be.
     "spacing": { type: "float", default: ".2", description: "LED spacing. Used for first and last edge." },
     "radius": { type: "float", default: "1.125", description: "Corner radius" },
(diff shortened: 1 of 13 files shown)


fakeplant:
@jkbelcher 
Screenshot 2024-07-16 at 5 36 31 PM

I'm noticing a few comment lines seem to be formatted "wrong", as in they don't conform to JSON5 specs.

ex: `# Debug tools...` should be `/* Debug tools... */` or `// Debug tools..`

Just a nice to have for syntax highlighting.

fakeplant:
Another JSON spec issue (in TE).

`type: float` should be `type: "float"`

Screenshot 2024-07-16 at 5 43 21 PM

jkbelcher:
> I'm noticing a few comment lines seem to be formatted "wrong", as in they don't conform to JSON5 specs.
> 
> ex: `# Debug tools...` should be `/* Debug tools... */` or `// Debug tools..`
> 
> Just a nice to have for syntax highlighting.

This was lazy on my part, it was an accident and when it didn't break the parser I stuck with it.  But you're right if it doesn't meet spec I should change it.

andrewlook (approved):
Looks great! 👍 ready to merge this

PR #507, opened 2024-07-20
APC mini Mk2 Driver: Color grid controls

andrewlook:


src/main/java/titanicsend/app/dev/DevSwitch.java
```diff
@@ -493,8 +503,10 @@ private void setMidiSurfacesEnabled(boolean enabled) {
   /** Returns TRUE for normal surfaces that should be enabled for TE production */
   private boolean isTESurface(LXMidiSurface surface) {
     return surface instanceof titanicsend.lx.APC40Mk2
+        || surface instanceof TEAPCminiMk2
         || surface instanceof titanicsend.lx.MidiFighterTwister
         || surface instanceof heronarts.lx.midi.surface.APC40Mk2
+        || surface instanceof heronarts.lx.midi.surface.APCmini
```
andrewlook, line 509: TODO: this should be `APCminiMk2` not `APCmini`

src/main/java/titanicsend/lx/APCminiSurface.java
```diff
@@ -0,0 +1,633 @@
...
+import heronarts.lx.parameter.EnumParameter;
+import heronarts.lx.parameter.LXListenableNormalizedParameter;
+import heronarts.lx.parameter.LXParameter;
+import heronarts.lx.parameter.LXParameterListener;
+import heronarts.lx.pattern.LXPattern;
+import heronarts.lx.utils.LXUtils;
+import titanicsend.color.ColorPaletteManager;
+import titanicsend.util.TE;
+
+public abstract class APCminiSurface extends LXMidiSurface implements LXMidiSurface.Bidirectional {
```
andrewlook, line 58: TODO: just merge this with `TEAPCminiMk2`, no need for the layer of abstraction here.

src/main/java/titanicsend/lx/LXUtilsWithRange.java
```diff
@@ -0,0 +1,39 @@
+package titanicsend.lx;
+
+public class LXUtilsWithRange {
```
andrewlook, line 3: TODO: once we update the upstream LX, these functions have been added to `LXUtils`

src/main/java/titanicsend/lx/FocusedChannel.java
```diff
@@ -0,0 +1,79 @@
...
+
+package titanicsend.lx;
+
+import heronarts.lx.LX;
+import heronarts.lx.mixer.LXBus;
+import heronarts.lx.mixer.LXChannel;
+import heronarts.lx.parameter.LXParameter;
+import heronarts.lx.parameter.LXParameterListener;
+
+public class FocusedChannel {
```
andrewlook, line 27: TODO: this class can be deleted since it's only used by the upsteam APCmini drivers for the params/pattern controls

src/main/java/titanicsend/midi/MidiNames.java
```diff
@@ -4,8 +4,10 @@ public class MidiNames {
 
   public static final String BOMEBOX = "FoH: ";
   public static final String APC40MK2 = heronarts.lx.midi.surface.APC40Mk2.DEVICE_NAME;
+  public static final String APCMINI = heronarts.lx.midi.surface.APCminiMk2.DEVICE_NAME;
```
jkbelcher, line 7: Let's name this APCMINIMK2 to prevent confusion with the v1.

jkbelcher:
This looks like draft status.  Better to do the cleanup before merging than after.

Also don't do this:
image

src/main/java/titanicsend/midi/MidiNames.java
andrewlook: thanks, done 👍

src/main/java/titanicsend/lx/DirectorAPCminiMk2.java
```diff
@@ -0,0 +1,547 @@
...
+    LXMidiEngine.error("APCmini unmapped control change: " + cc);
+  }
+
+  private void sendGrid() {
+    for (int row = 0; row < 8; row++) {
+      for (int col = 0; col < 8; col++) {
+        int idx = (row * 8) + col;
+        sendIndividualSysEx(idx, 0x000000);
+        try {
+          // Need to sleep between each SysEx message (to avoid MIDI overflow?)
```
jkbelcher, line 322: Yeah the small buffer on that device fills up easily.  With normal midi notes it was really only a problem at startup if you set every button to Off and then immediately set every button to a color.  Unsure where the limit is with these sysex colors.

jkbelcher:
Made some edits.  FYI:

- Removed the references to specific fader assignments in the director module.  I'll think it'll be nice to modify those without also changing the midi surface class.
- Ignoring white button in Chromatik
- Added the new fancy fader modes including pickup & scale
- Pointed the midi master fader at Director master since it's different from Chromatik master for now.
- Removed the full grid clear on startup. Not needed since the color grid covers every button and saves some sleep cycles.

Thanks for working this up!

jkbelcher (approved):
Merging to make available for use this weekend.

PR #508, opened 2024-07-23
small typo fix in AudioStems.java

worldveil:


src/main/java/titanicsend/audio/AudioStems.java
```diff
@@ -14,7 +14,7 @@ public class AudioStems extends LXComponent implements LXOscListener {
   public static final String PATH_STEM = "/te/stem/";
   public static final String PATH_BASS = "bass";
   public static final String PATH_DRUMS = "drums";
-  public static final String PATH_VOCAL = "vocal";
+  public static final String PATH_VOCAL = "vocals";
```
jkbelcher, line 17: I wondered about that.  Do you also want to make it PATH_VOCALS?

worldveil: @jkbelcher good catch - did this everywhere in the class so should be consistent!

jkbelcher (approved):
Looks good.  Tested.

PR #509, opened 2024-07-23
Director component

jkbelcher:
Adds new top-level component for central control of all TE lighting systems.

To apply filtering the DirectorEffect needs to be added to the master channel.  In a future PR we'll set that to happen automatically.


src/main/java/titanicsend/app/director/Director.java
```diff
@@ -0,0 +1,142 @@
...
+      // TODO: find matching DMX fixtures
+    }
+  }
+
+  public static Director current;
+
+  private final List<Filter> mutableFilters = new ArrayList<Filter>();
+  public final List<Filter> filters = Collections.unmodifiableList(this.mutableFilters);
+
+  public final CompoundParameter main =
```
andrewlook, line 98: nit: `mainFader` or `mainBrightness` might be a bit more clear

src/main/java/titanicsend/app/director/Director.java
```diff
@@ -0,0 +1,142 @@
...
+    public void run(int[] colors) {
+      // TODO: scale the dimmer for DMX fixtures
+    }
+
+    public void modelChanged(LXModel model) {
+      // TODO: find matching DMX fixtures
+    }
+  }
+
+  public static Director current;
```
andrewlook, line 93: I'd consider making this private and having a public static `.get()` or `.getInstance()`, just so this reference doesnt accidentally get mutated by something else

src/main/java/titanicsend/app/director/Director.java
```diff
@@ -0,0 +1,142 @@
...
+    addFilter(new TagFilter("te", "TE", "te"));
+    addFilter(new TagFilter("panels", "Panels", "panel"));
+    addFilter(new TagFilter("edges", "Edges", "edge"));
+    addFilter(new TagFilter("foh", "FOH", "mothership"));
+    addFilter(new Filter("lasers", "Lasers"));
+    addFilter(new DmxFilter("beacons", "Beacons", "beacon"));
+
+    addParameter("main", this.main);
+
+    lx.addListener(this);
```
andrewlook, line 118: just curious, what Listener methods is this overriding? sorry I'm less familiar with `LXListener` so don't know what to expect there. Is it `modelGenerationChanged` (and if so does that need an `@Override`)?

src/main/java/titanicsend/app/director/Director.java
```diff
@@ -0,0 +1,142 @@
...
+import heronarts.lx.LXComponent;
+import heronarts.lx.color.LXColor;
+import heronarts.lx.model.LXModel;
+import heronarts.lx.model.LXPoint;
+import heronarts.lx.osc.LXOscComponent;
+import heronarts.lx.parameter.CompoundParameter;
+
+public class Director extends LXComponent implements LX.Listener, LXOscComponent {
+
+  public static class Filter {
```
andrewlook, line 17: style nit: this one's more of a personal preference, but I find it hard to get to the meat of what this class does when there are inner classes higher up in the source code - when I'm scanning a file, ideally I'd like to see the member fields, constructor, key methods first.

feel free to disregard this if you prefer it as-is, but given that `director` is a new package I wouldn't mind having these Filters live in their own files as siblings of this class

src/main/java/titanicsend/app/director/UIDirector.java
```diff
@@ -0,0 +1,65 @@
...
+
+import heronarts.glx.ui.UI2dContainer;
+import heronarts.glx.ui.component.UICollapsibleSection;
+import heronarts.glx.ui.component.UISlider;
+import heronarts.lx.parameter.LXListenableNormalizedParameter;
+import heronarts.lx.studio.LXStudio.UI;
+import heronarts.lx.studio.ui.device.UIControls;
+import titanicsend.app.director.Director.Filter;
+
+public class UIDirector extends UICollapsibleSection implements UIControls {
```
andrewlook, line 11: FWIW I like this code organization of keeping the UI and Effect colocated with the Director class (vertically organized by feature), compared to the convention of ui/effect/etc living in separate packages (horizontally organized by type of component). Just wanted to call that out, I might do the same with my color stuff if no one objects.

src/main/java/titanicsend/ui/color/UIColorPaletteManager.java
```diff
@@ -170,11 +170,11 @@ private UI2dContainer row(String label, int labelWidth, UI2dComponent component)
         16, 4, new UILabel(labelWidth, 12, label), component);
   }
 
-  public static void addToLeftGlobalPane(LXStudio.UI ui, ColorPaletteManager cueMgr, ColorPaletteManager auxMgr) {
+  public static void addToLeftGlobalPane(LXStudio.UI ui, ColorPaletteManager cueMgr, ColorPaletteManager auxMgr, int index) {
     UI2dContainer parentContainer = ui.leftPane.global;
     float xOffsetWithinParent = 0;
     new UIColorPaletteManager(ui, cueMgr, auxMgr, parentContainer.getContentWidth(), xOffsetWithinParent)
-        .addToContainer(parentContainer, 3);
+        .addToContainer(parentContainer, index);
```
andrewlook, line 177: thanks for externalizing that config, didn't feel great to have a magic number sitting in there.

src/main/java/titanicsend/app/director/Director.java
jkbelcher: Good point, "main" often refers to the main car which currently I have labeled "TE".  How about "master" for this?  Adding the word `fader` or `brightness` still doesn't distinguish it from the car.

src/main/java/titanicsend/app/director/Director.java
jkbelcher: Thanks,done.

src/main/java/titanicsend/app/director/Director.java
jkbelcher: Yes it is `modelGenerationChanged`.  Good point, although it's not required it is best practice to mark it with `@Override`.  Added!

src/main/java/titanicsend/ui/color/UIColorPaletteManager.java
jkbelcher: Yeah it got knocked out of position when I added the director component up top.

src/main/java/titanicsend/app/director/Director.java
jkbelcher: Now that you mention it, I'm happy when inner classes are at the bottom of the file too.  This was based on LX convention where they are usually at the top.  Since they're static I'll just bump them out to their own files.

jkbelcher:
Condensed the faders to one row with staggered labels:

image

src/main/java/titanicsend/app/director/UIDirector.java
andrewlook: cool! yeah IMHO for a framework/library it makes sense to horizontally organize, but for a lot of what we're doing (outside of patterns anyways) I like the vertical package organization too 🙌

src/main/java/titanicsend/app/director/Director.java
```diff
@@ -0,0 +1,69 @@
...
+import heronarts.lx.LXComponent;
+import heronarts.lx.model.LXModel;
+import heronarts.lx.osc.LXOscComponent;
+import heronarts.lx.parameter.CompoundParameter;
+
+public class Director extends LXComponent implements LX.Listener, LXOscComponent {
+
+  private static Director current;
+
+  public static Director get() {
```
andrewlook, line 17: super minor / non-blocking tip: you could actually make this package-protected, so only classes inside `titanicsend.app.director` could access it (which works b/c `UIDirector` is in here)

src/main/java/titanicsend/app/director/UIDirector.java
```diff
@@ -0,0 +1,74 @@
...
+    new UISlider(
+      UISlider.Direction.VERTICAL,
+      (column * this.columnWidth) + ((this.columnWidth - FADER_WIDTH) / 2),
+      0,
+      FADER_WIDTH,
+      FADER_HEIGHT,
+      parameter)
+    .setShowLabel(false)
+    .setWidth(20)
+    .addToContainer(uiDevice);
```
andrewlook, line 65: non-blocking style nit: indentation looks a little funky, might want to run `mvn spotless:apply` here in a subsequent PR

andrewlook (approved):
Changes are looking great! When I tested locally the faders display properly and nothing breaks. I didn't actually see the brightness change when I tried the master fader, not sure if that should be working - but I'm gonna go ahead and merge this so I can start mapping the MIDI controller to these faders, and I'll circle back with you offline to see if I was testing it properly or if there's anything we need to fix about the showfile or java code for the faders to have the desired effect.

src/main/java/titanicsend/app/director/Director.java
jkbelcher: I think the `Director` midi surface needs access to this package and also the `color` package.  And might should live in the `midi` package.

jkbelcher:
It just needs the DirectorEffect added to the master channel, then the faders should work.  I'm waiting to merge a "locked effect" feature Mark just added for us in upstream and then I'll set the effect to be added automatically.

PR #510, opened 2024-07-24
Add legacy Edge and Panel tags to fixtures

jkbelcher:
Added capitalized "Edge" tag to edge LXFs and "Panel" tag to panel LXFs.

This fixes the Edges and Panels views in older project files.  Going forward I'd vote we use all lowercase for tags.


andrewlook (approved):
Awesome, thanks for the backwards compatibility here. Agreed on lowercase tags moving forwards.

PR #512, opened 2024-07-25
Fix model axes for HandTracker pattern

zranger1:
HandTracker is needed for debugging during build, and got missed in the first axis-swap pass because it uses "real" model coords instead of the normalized set.

It should work normally now, on both static and dynamic models.

Also fixed a stray vocal/vocals stem-related variable name that was hiding in the shader framework.  (And changed it in the docs, too.)


jkbelcher (approved):
Tested.  Looks good!

PR #514, opened 2024-07-26
Restore Panel Backings and Labels

zranger1:
#### What is new
- Panel backings and panel labels now work on both static and dynamic models.
- All 3D UI elements (all labels and laser boxes) rebuild automatically on model changes.  They are dependent on metadata, so will not be displayed on models (like the Mothership) for which the metadata is not present.

#### What is unchanged
- Vertex labels are not available on the dynamic model because there is no vertex position information in the model.  
- Laser box display is not available on the dynamic model because there is no data for the boxes.

_(both of these items can be added to the dynamic model preview if it is high enough priority_)

### DIFF
--- a/src/main/java/heronarts/lx/studio/TEApp.java
+++ b/src/main/java/heronarts/lx/studio/TEApp.java
@@ -94,7 +94,8 @@ import titanicsend.pattern.yoffa.config.OrganicPatternConfig;
 import titanicsend.pattern.yoffa.config.ShaderEdgesPatternConfig;
 import titanicsend.pattern.yoffa.config.ShaderPanelsPatternConfig;
 import titanicsend.pattern.yoffa.effect.BeaconEffect;
-import titanicsend.ui.*;
+import titanicsend.ui.UI3DManager;
+import titanicsend.ui.UITEPerformancePattern;
 import titanicsend.ui.color.UIColorPaletteManager;
 import titanicsend.ui.effect.UIRandomStrobeEffect;
 import titanicsend.ui.modulator.UIDmx16bitModulator;
@@ -177,7 +178,6 @@ public class TEApp extends LXStudio {
       } else {
         this.paletteManagerB = null;
       }
-      lx.engine.registerComponent("ui3dManager", this.ui3dManager = new UI3DManager(lx));
 
       // create our loop task for outputting data to lasers
       this.laserTask = new TELaserTask(lx);
@@ -564,26 +564,11 @@ public class TEApp extends LXStudio {
       UIColorPaletteManager.addToLeftGlobalPane(ui, this.paletteManagerA, this.paletteManagerB, 4);
       UIColorPaletteManager.addToRightPerformancePane(ui, this.paletteManagerA, this.paletteManagerB);
 
+      // Add 3D Ui components
+      this.ui3dManager = new UI3DManager(lx, ui, this.virtualOverlays);
+
       // Set camera zoom and point size to match current model
       applyTECameraPosition();
-
-      // 3D UI components - save objects in ui3dManager so we can
-      // rebuild them on model changes.
-      ui3dManager.modelLabels = new UIModelLabels(lx, this.virtualOverlays);
-      ui.preview.addComponent(ui3dManager.modelLabels);
-
-      ui3dManager.backings = new UIBackings(lx, this.virtualOverlays);
-      ui.preview.addComponent(ui3dManager.backings);
-      ui3dManager.backingsAux = new UIBackings(lx, this.virtualOverlays);
-      ui.previewAux.addComponent(ui3dManager.backingsAux);
-
-      // TODO - laser preview not currently supported on dynamic model.  Do we
-      // TODO - need them?  Do we want to render at least beam color and
-      // TODO - angle if we have that info from Pangolin?
-      ui3dManager.lasers = new UILasers(lx, this.virtualOverlays);
-      ui.preview.addComponent(ui3dManager.lasers);
-      ui3dManager.lasersAux = new UILasers(lx, this.virtualOverlays);
-      ui.previewAux.addComponent(ui3dManager.lasersAux);
 
       // precompile binaries for any new or changed shaders
       ShaderPrecompiler.rebuildCache();
--- a/src/main/java/heronarts/lx/studio/TEApp.java
+++ b/src/main/java/heronarts/lx/studio/TEApp.java
@@ -567,16 +567,23 @@ public class TEApp extends LXStudio {
       // Set camera zoom and point size to match current model
       applyTECameraPosition();
 
-      // 3D components
-      ui.preview.addComponent(new UIBackings(lx, this.virtualOverlays));
-      ui.previewAux.addComponent(new UIBackings(lx, this.virtualOverlays));
-
-      ui.preview.addComponent(new UILasers(lx, this.virtualOverlays));
-      ui.previewAux.addComponent(new UILasers(lx, this.virtualOverlays));
-
-      // 3D model labels - save object so we can use it whenever model changes.
+      // 3D UI components - save objects in ui3dManager so we can
+      // rebuild them on model changes.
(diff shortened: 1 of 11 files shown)


src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -94,10 +94,7 @@
 import titanicsend.pattern.yoffa.config.ShaderEdgesPatternConfig;
 import titanicsend.pattern.yoffa.config.ShaderPanelsPatternConfig;
 import titanicsend.pattern.yoffa.effect.BeaconEffect;
-import titanicsend.ui.UIBackings;
-import titanicsend.ui.UILasers;
-import titanicsend.ui.UIModelLabels;
-import titanicsend.ui.UITEPerformancePattern;
+import titanicsend.ui.*;
```
jkbelcher, line 97: Did the formatter do this?  I thought we were staying away from .* imports

src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -177,6 +177,7 @@ public Plugin(LX lx) {
       } else {
         this.paletteManagerB = null;
       }
+      lx.engine.registerComponent("ui3dManager", this.ui3dManager = new UI3DManager(lx));
```
jkbelcher, line 180: I think this doesn't need to be registered as a component since it doesn't have parameters that need to be saved/loaded.  However I tried making it not extend LXComponent and it crashes, so... leave that part in at least.

src/main/java/titanicsend/ui/UIBackings.java
```diff
@@ -58,33 +54,54 @@ protected void bufferData(ByteBuffer buffer) {
...
 
+    // if using the static model, we can build our backing panels now
+    // TODO - remove when switch to dynamic model is complete
+    if (TEApp.wholeModel.isStatic()) {
+      rebuild();
+    }
+  }
+
+   /** Build backing panels for the current model. Called when the model is loaded or changed. */
+  public void rebuild() {
```
jkbelcher, line 82: This is called by the engine thread but `UIBackings.onDraw()` runs on the ui thread.  Danger?

src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -566,16 +567,23 @@ public void onUIReady(LXStudio lx, LXStudio.UI ui) {
...
-
-      ui.preview.addComponent(new UIModelLabels(lx, this.virtualOverlays));
-      // Do we need model labels in the secondary view?  Uncomment if so.
-      // ui.previewAux.addComponent(new UIModelLabels(lx, this.virtualOverlays));
+      // 3D UI components - save objects in ui3dManager so we can
+      // rebuild them on model changes.
+      ui3dManager.modelLabels = new UIModelLabels(lx, this.virtualOverlays);
+      ui.preview.addComponent(ui3dManager.modelLabels);
+
+      ui3dManager.backings = new UIBackings(lx, this.virtualOverlays);
```
jkbelcher, line 575: Can we pass virtualOverlays to the `UI3DManager` constructor and initialize `UIBackings` and `UILasers` there?  If so they could be `public final`.

src/main/java/titanicsend/ui/UI3DManager.java
```diff
@@ -0,0 +1,28 @@
...
+// Thin wrapper that allows global access to 3D UI components, so we can
+// update them easily when the model changes
+public class UI3DManager extends LXComponent {
+  public static UI3DManager current;
+
+  public UIModelLabels modelLabels = null;
+
+  // Chromatik requires a separate UI object per context (main and aux) if
+  // we want the element to show up in performance mode.
+  public UIBackings backings = null;
```
jkbelcher, line 18: Candidate for `public final`?

src/main/java/titanicsend/ui/UIModelLabels.java
```diff
@@ -33,40 +33,79 @@
...
+    if (this.isStatic) {
+      rebuild();
+    }
+  }
+
+  /** Build labels for the current model. Called when the model is loaded or changed. */
+  public void rebuild() {
+
+    // TODO - remove static checks when switch to dynamic model is complete
+    float fontScale = (isStatic) ? 15000f : 0.5f;
```
jkbelcher, line 65: Solid reverse compatibility.

src/main/java/titanicsend/ui/UIBackings.java
```diff
@@ -58,33 +54,54 @@ protected void bufferData(ByteBuffer buffer) {
...
 
     this.colorBuffer =
-        new DynamicVertexBuffer(glx, this.colorBufferMaxSize, VertexDeclaration.ATTRIB_COLOR0);
+      new DynamicVertexBuffer(glx, colorBufferMaxSize, VertexDeclaration.ATTRIB_COLOR0);
+  }
+
+  // Free resources allocated by previous model
+  public void clear() {
+    for (PanelBuffer b : this.panels) {
+      b.dispose();
```
jkbelcher, line 101: Wondering if these are safe to dispose before clearing the collection if `onDraw()` is a different thread.

jkbelcher (commented):
Thanks for building this!  I think this will make people a lot more comfortable with the new dynamic model.

Added a few minor comments and one important one regarding threading.

src/main/java/heronarts/lx/studio/TEApp.java
zranger1: Wow, the formatter must've done it.  I'll fix shortly.

zranger1: Fixed in latest commit.

src/main/java/heronarts/lx/studio/TEApp.java
zranger1: Fixed in latest commit.  I was expecting it to have to do more - as it turns out it doesn't need to be registered and it doesn't need to extend LXComponent.   It just has to initialize the UI components at a point when the LX object has valid information about Chromatik's 3D ui context.

src/main/java/titanicsend/ui/UIBackings.java
zranger1: Another good catch - on project load/reload which is what I initially tested, the draw thread isn't running, so no problem.   However, if you turn on Model Labels, and use the fixture interface to move a panel around, it is (was) usually possible to eventually cause an exception.   

Fixed by protecting rebuild and draw w/atomic boolean.

src/main/java/heronarts/lx/studio/TEApp.java
zranger1: Good call! Done in latest commit -- implementing this let me clean things up, and inflict less clutter on TEApp.java!

src/main/java/titanicsend/ui/UI3DManager.java
zranger1: Done in latest commit.

src/main/java/titanicsend/ui/UIModelLabels.java
zranger1: Took me a little while to realize that the coords had changed -- first time I turned it on, labels were 1000 times as big as the car!

src/main/java/titanicsend/ui/UIBackings.java
zranger1: Fixed in latest commit.  Drawing will now be locked out during model rebuilds.

src/main/java/titanicsend/ui/text3d/Label.java
```diff
@@ -42,6 +44,8 @@ public Label(String text, Vector3f pos, Vector3f rot) {
   }
 
   public void dispose() {
+    bgfx_destroy_vertex_buffer(vbh);
```
andrewlook, line 47: ah, is this the subtle memory leak you mentioned?

andrewlook (approved):
Changes are looking great! Thanks @jkbelcher for catching that thread-safety issue. Tested locally and it's working well, merging now

PR #515, opened 2024-07-27
CSV export of color grid

jkbelcher:
CSV export, adds white button, reduces hue shift to 31 steps instead of 32.


andrewlook (approved):
Looking great, thanks for doing this!

PR #516, opened 2024-07-27
tweak: add .lxpreferences to .gitignore

andrewlook:
my thinking here: we almost never actually *need* to commit changes to this file (and if we do, we can run `git add -f .lxpreferences`).

More often than not, it just leaves us with a dirty tree that we need to reset or stash before switching branches. Especially as some nontechnical camp-mates have expressed interest in being able to play around with the app locally to get familiar with the patterns, I think it'd make life considerably easier for them to be able to `git pull` without extra obstacles.

Open to skipping this change if anyone feels strongly, just wanted to put the idea out there.


jkbelcher:
Definitely agree. I looked into it once though and it seemed to be not guaranteed to be gitignored once it has been added.

andrewlook:
@jkbelcher the instructions here worked well for me: https://superuser.com/questions/1397199/how-to-ignore-a-tracked-file-in-git-without-deleting-it

i just ran `git rm --cached .lxpreferences`, which deleted the file from the git tree (but left my local copy), and i made a commit that deletes the file.

So the first time people pull the repo after this PR is merged, it'll delete their local `.lxpreferences`. I checked this for myself, deleting the file and running the app - Chromatik started up fine, and I just had to accept the EULA again. 

Note: If there are any particular settings we'd like people to have in there, we could commit a file `.lxpreferences.example` and leave instructions in the README for people to `cp .lxpreferences.example .lxpreferences`. I don't *think* we need this, but LMK if we do.

jkbelcher:
Sounds great.  Let's try it.

jkbelcher (approved):
Agreed this is better as a local user file.

PR #519, opened 2024-07-29
Modulator for audio stems

jkbelcher:
Simple modulator for easy mapping of audio stems.

Could be expanded later to include high/low pass, smoothing, etc.

image


worldveil:
works for me very well!

worldveil (approved):

PR #520, opened 2024-07-30
Add smoothing parameter to AudioStemModulator

jkbelcher:
Adds an adjustable smoothing time (in milliseconds) to the stem modulator output.

Math is from the LX Smoother modulator.

image


jkbelcher:
Changed to EMA.  The output looks much nicer.

image

zranger1 (approved):
Thanks for doing this!  I think we'll be way happier doing this in Chromatik where the VJ can control it. 

And using milliseconds instead of alpha in the UI and EMA setup - nice touch!  Much easier for people to understand the smoothness vs. responsiveness tradeoff that way.

PR #522, opened 2024-08-02
Replace SpiralDiamonds w/Shader

zranger1:
- SpiralDiamonds:  Much faster now.
- OldSpiralDiamonds:  Old java version, in case it is needed

#### Discussion:
This is a 1:1 plugin replacement for the old SpiralDiamonds except:
When I looked at the java pattern to see how hard the port would be, I saw to my astonishment that it WASN'T ACTUALLY MAKING A SPIRAL.  It is supposed to.  But it was just making rings of expanding boxes. 

So I wrote a shader that fixes the math (there was a mistake when calculating the variable 'tilt' rotation term that makes the box edges join into a spiral.)   Now it actually make a spiral of diamonds, as intended.

Level reactivity is also slightly improved over the old version. 

And oh, it will no longer cause frame rate drops!

### DIFF
--- a/resources/shaders/spiral_diamonds.fs
+++ b/resources/shaders/spiral_diamonds.fs
@@ -1,0 +1,52 @@
+// MUCH FASTER replacement for the Spiral Diamonds java pattern
+//
+uniform float speedAngle;
+
+mat2 rot(float a) {
+    float s=sin(a), c=cos(a);
+    return mat2(c,s,-s,c);
+}
+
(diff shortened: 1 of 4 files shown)


jkbelcher (approved):
Confirmed this is much faster! Nice! Previously 60% CPU on my laptop, now much lower and not clipping the channel.  The anti-aliasing is nice too.

Thanks for fixing the rotation direction.

Additional idea, the "amount of waviness" could be tied to a Wow knob.

PR #523, opened 2024-08-03
Quantize!

jkbelcher:
First implementation of a Quantize feature.

The Quantizer module delays method calls to align them with system events such as beat/bar.

Changed WowTrigger to be a QuantizeTriggerParameter for quantized input.


src/main/java/titanicsend/quantize/Quantizer.java
```diff
@@ -0,0 +1,177 @@
...
+  @Override
+  public void onBeat(Tempo tempo, int beat) {
+    if (ready() && this.event == QuantizeEvent.BEAT) {
+      fire();
+    }
+  }
+
+  @Override
+  public void onBar(Tempo tempo, int bar) {
+    if (ready() && this.event == QuantizeEvent.BEAT) {
```
andrewlook, line 167: should this be `QuantizeEvent.BAR`?

src/main/java/titanicsend/quantize/QuantizeTriggerParameter.java
```diff
@@ -0,0 +1,51 @@
...
+    // Delay trigger on, do not delay trigger off
+    setValue(value, !value);
+    return this;
+  }
+
+  public QuantizeTriggerParameter setValue(boolean value, boolean immediate) {
+    if (immediate) {
+      super.setValue(value);
+    } else {
+      this.quantizer.queue();
```
andrewlook, line 26: should we also pass `value` in to this? or is that not relevant because we're only dealing with `TriggerParameter`?

src/main/java/titanicsend/pattern/TECommonControls.java
```diff
@@ -117,9 +118,8 @@ public void buildDefaultControlList() {
     setControl(TEControlTag.WOW2, p);
 
     p =
-        new BooleanParameter(TEControlTag.WOWTRIGGER.getLabel(), false)
-            .setMode(BooleanParameter.Mode.MOMENTARY)
-            .setDescription("Trigger WoW effects");
+      new QuantizeTriggerParameter(this.pattern.getLX(), TEControlTag.WOWTRIGGER.getLabel())
+          .setDescription("Trigger WoW effects");
```
andrewlook, line 122: I noticed that some of our wow triggers that could be held down

jkbelcher:
We're pausing on this since some WowTrigger parameters are press-and-hold.  We need a QuantizeBooleanParameter that delays the Off action until after a queued On action has executed.  But does not quantize the off.

src/main/java/titanicsend/quantize/QuantizeTriggerParameter.java
jkbelcher: Correct

src/main/java/titanicsend/pattern/TECommonControls.java
jkbelcher: Confirmed.  We'll need a momentary boolean working before we can replace WowTrigger.

andrewlook (approved):

PR #524, opened 2024-08-03
Added stem runner conf and bash script

worldveil:
Adds a configuration for lighting-1 as well as a bash script we can automate to run each startup on lighting-1

### DIFF
--- a/resources/stems/lighting-1.conf
+++ b/resources/stems/lighting-1.conf
@@ -1,0 +1,55 @@
+[Global]
+; Maximum size in sample frames of system buffers
+RingBufferCapacity = 4096
+
+; Displayable stem names.  Must reflect the order of the model's output.
+Stem0 = Bass
+Stem1 = Drums
+Stem2 = Other
+Stem3 = Vocals
+
+[Audio]
+; Specify input device by name, or "default" for the default device. Name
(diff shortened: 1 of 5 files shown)


zranger1 (approved):
This looks great.  (One thing to remember:  We may need to change the OSC address in the config files to send stem traffic to Max so it can be routed to Sina's Stable Diffusion project as well as to Chromatik. )

resources/stems/lighting-1.conf
```diff
@@ -0,0 +1,55 @@
...
+; To get a list of compatible devices, start the executable with the
+; -d or --devices option.
+CaptureDevice = default
+
+; Don't change this unless the model has changed to require a
+; different sample rate.
+SampleRate = 44100
+
+[Model]
+ModelPath = /Users/te/src/code/LXStudio-TE/stemmie-convnet-v1.1.onnx
```
andrewlook, line 24: just to be safe, should we consider either committing this `.onnx` file, or maybe adding a `wget` command to the `start_stem_runner.sh` script? I may have missed another spot where this got committed, just double-checking. happy to make the change myself if helpful.

andrewlook: nvm, just remembered that the model file should be private - please disregard this comment

PR #526, opened 2024-08-04
Fix Fireflies pattern

zranger1:
- Fixes Fireflies lack of significant trails (controlled by Wow1).   This has been broken-ish since the big Coachella reactivity push, and I never had time to fix it.  Flies can now have very long, or very short trails,  as intended.
- Add stem-based reactivity,  controlled by levelReact (drums) and freqReact (vocals) controls.  Um... wow.  That just works.

Video demo of improved Fireflies w/new reactivity in software channel on Slack. 

- Also includes experimental RingOfFire shader for use on Mothership.   Give it a try and let me know what we can improve.  Until I've seen the thing assembled, it's a little hard to feel for how patterns will read IRL.


resources/shaders/fireflies.fs
```diff
@@ -1,27 +1,54 @@
...
-    // blob size pulses with the beat, controlled by
-    // WOW2
-    float pulse = 0.01 - (iScale * iWow2 / 100. * beat);
+
+    // blob size pulses with drums, amount controlled by levelReact
+    float blobScale = (stemDrums > 0.1) ? -1.0 + 2.0 * stemDrums : 0.0;
+    float radius = iScale + (levelReact * iScale * blobScale);
+
+    // inter-firefly spacing changes w/stem value, controlled by frequencyReact
+    // turned up, this gets really fun!
```
andrewlook, line 17: testing it out now - this is AWESOME! 🙌

resources/shaders/starfield1.fs
```diff
@@ -0,0 +1,94 @@
+#pragma name "StarField1"
+#pragma TEControl.QUANTITY.Range(1.5,2.5,0.15)
+#pragma TEControl.SIZE.Range(0.05,0.025,0.125)
+#pragma TEControl.WOW2.Value(0.7);
```
andrewlook, line 4: minor nit: i think we may want to add `WOW1.Disable` here

zranger1: Those patterns shouldn't be in this PR.  I must've messed up a checkin at some point.   They're actually supposed to be in https://github.com/titanicsend/LXStudio-TE/pull/533.   The versions there have all unused controls disabled. 

These guys basically work, so it'd be ok to check them in 'till they get updated by 533.   Or I can remove them if you'd rather.

zranger1: (Sneaky stowaways are the updated circuitry and both starfields, btw.)

src/main/java/titanicsend/pattern/jon/FxXWave.java
```diff
@@ -0,0 +1,119 @@
...
+
+      if (dist <= width) {
+        //color = setBrightness(color, (float) TEMath.clamp(dist, 0f, 1f));
+        colors[point.index] = color;
+      }
+    }
+  }
+
+  // Button activation logic.  Keep going while button is held,
+  // stop at end of cycle when released.
```
andrewlook, line 106: nice idea for the wowtrigger!! I appreciate the comment pointing this out 🙏

andrewlook (approved):
Tested all of these patterns out locally, and they are looking excellent! I'm super excited about the fireflies revamp, given how useful and versatile that one proved to be at coachella.

resources/shaders/starfield1.fs
andrewlook: no worries! I'll review the other PR's next, sorry if I went out-of-order!

PR #527, opened 2024-08-04
Set auto-focus preferences in case .lxpreferences does not exist

jkbelcher:
Force these two user preferences to TE defaults.  Now that we removed `.lxpreferences` from the repo they are reverting to default values.

image


andrewlook (approved):
verified this locally and it's working for me - i delete .lxprefs, start app, cue something, and see that the channel is focused. merging now.

PR #529, opened 2024-08-07
Jeff/bm24 ips and showkontrol

jvyduna:
1. ShowKontrol Automator App needed a rename to work with new OSX Privacy/Accessibility permissions
2. IP address update for Mothership Chromatik to receive tempo from ShowKontrol via Max/MSP patch


jkbelcher (approved):
Thanks for updating this.  Looks fine on the surface.  I'll take your word it's tested.

PR #530, opened 2024-08-09
New stem enabled patterns for use with StemmieNet

worldveil:


jkbelcher (approved):
These are just presets.  Merge away!

andrewlook (approved):
Awesome! Merging these now - I'll try to spin up a showfile with these loaded up and share it with the group 🙌

PR #531, opened 2024-08-09
Color offset within gradient, per pattern

jkbelcher:
Pulls our `TEColorParameter` off the bench and revamps it to be a color offset [within a gradient of the current global palette/swatch].  Also tuned up related color things while we were under the knife:

- Moved gradient calculations to happen ONCE at a global level instead of EVERY PATTERN.  This might be a measurable performance improvement.
- Removed swatch colors from indexes [0,1,4] that were black.  We can construct dark gradients without those now.
- Simplified our gradient options to "Normal" and "Dark".  The previous options were too advanced / not intuitive enough to get any use.
- `TEColorParameter`: removed some clutter including `Offset2`
- `TEColorParameter`: combined "Static Color" and "Gradient" into a single "Color Source"
- Moved the color parameter back onto the MidiFighterTwister.  I picked a random spot of bottom right.
- Added a badass new preview icon for the color parameter.  It shows the gradient and moves a triangle above and below the gradient to indicate the Offset value.  Now you can see if it's at the beginning or middle of the gradient:
image

### DIFF
--- a/Colors/te_colors_2024.lxc
+++ b/Colors/te_colors_2024.lxc
@@ -1,0 +1,900 @@
+{
+  "swatches": [
+    {
+      "class": "heronarts.lx.color.LXSwatch",
+      "internal": {
+        "modulationColor": 0,
+        "modulationControlsExpanded": true,
+        "modulationsExpanded": true
+      },
+      "parameters": {
+        "label": "PlumIce",
+        "recall": false,
+        "autoCycleEligible": true
+      },
+      "children": {},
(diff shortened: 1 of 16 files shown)


jkbelcher:
This is the new look of the overlay when you click the color parameter:
image

worldveil:
Added the '24 colors!

The palette manager doesn't default to the current swatch, but that might be by design. 

LGTM

worldveil (approved):
LGTM

PR #532, opened 2024-08-10
Update README.md for Termurin JDK via brew

titanicsend:


andrewlook (approved):
just double-checked on my system and it seems like `temurin17` still works but it's an alias for `temurin@17`. Merging now

PR #534, opened 2024-08-11
Set shader rendering resolution from TEApp command line

zranger1:
You can now use the command line option `--resolution WIDTHxHEIGHT` to specify a rendering resolution for the shader system.  

This helps tremendously in cases where we're dealing with a larger multi-car model, or where the aspect ratio of the model is significantly different from the original Titanic's End car (which is roughly 4:3).  For example, the Mothership has a 1:1 aspect ratio, and will look better with `--resolution 512x512` or some other square resolution.   

For now there are no limits on how low or how high you can go.  But be aware, higher resolutions can have a large impact on performance.

### DIFF
--- a/resources/shaders/ringofire.fs
+++ b/resources/shaders/ringofire.fs
@@ -1,0 +1,74 @@
+// Experimental fire for Mothership
+#pragma name "RingOfFire"
+#pragma TEControl.SIZE.Range(0.91,1.15,0.8)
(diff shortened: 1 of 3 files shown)


andrewlook (approved):
tested locally - works great! just for fun I tried `--resolution 64x64` on TE main, and it's like a funny retro pixel effect :) 

image

PR #535, opened 2024-08-12
Chromatik libs 1.0.1.TE.2

jkbelcher:
Update to latest Chromatik dev build.  

Includes:
- A beta version of allowing custom components in performance panes. 
- Lockable effects, which we'll use to prevent accidental deletion of the DirectorEffect.
- Other cool stuff.


jkbelcher:
Regarding MIDI surfaces:
- The APC40mk2 now extends the framework version, so it contains only our custom pieces.  This will be MUCH easier to maintain than mirroring every edit from the framework version to our copy.
- Removed our copy of the MidiFighterTwister class, as those changes had been absorbed into upstream.

andrewlook:
working great for me! Just tested locally - excited to see the new bottom-right perf pane! 
image

andrewlook (approved):

PR #536, opened 2024-08-12
Director APCminiMk2: improve name for local device

jkbelcher:
It will now work to rename the local device to "Director" in MIDI Studio.

In the end this presents to Chromatik differently than the Bomebox virtual midi port.  The virtual port comes across as "Director" while the local one appears as "Director Control".

### DIFF
--- a/src/main/java/titanicsend/lx/DirectorAPCminiMk2.java
+++ b/src/main/java/titanicsend/lx/DirectorAPCminiMk2.java
@@ -48,7 +48,9 @@ import titanicsend.util.TE;
  */
 public class DirectorAPCminiMk2 extends LXMidiSurface implements LXMidiSurface.Bidirectional {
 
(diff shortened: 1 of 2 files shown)


src/main/java/titanicsend/lx/DirectorAPCminiMk2.java
```diff
@@ -48,7 +48,9 @@
  */
 public class DirectorAPCminiMk2 extends LXMidiSurface implements LXMidiSurface.Bidirectional {
 
-  public static final String DEVICE_NAME = "APC mini mk2 (Director) Control";
+  // To use locally, rename system device to match LOCAL_DEVICE_NAME
```
andrewlook, line 51: nice! thanks for the comment

andrewlook (approved):

PR #538, opened 2024-08-12
WIP: Shaders use real coordinates

jkbelcher:
Rough draft of a new concept: use LXPoint normalized coordinates in shader math instead of normalized coordinates of an x,y grid.  Rather than output an image, output the point colors in sequential order matching the colors[] array.  This should result in precise calculation for every point and a faster loop in the java pattern.

Took a shot at adapting FollowThisStar as an example.  Something is still not quite right... needs someone familiar with the engine who can see what I was going for here and identify the missing piece.

Update: 12/2024:
This branch is now ready to merge with main, pending test, code review, fixing any issues, etc.

The 3d renderer concept is fully implemented, along with other performance improvements to the shader and audio analysis systems.

(For a brief introduction to the new renderer, see the first section of the [shader framework documentation](https://github.com/titanicsend/LXStudio-TE/blob/justin/3d/resources/docs/ShaderFramework.md). )


.run/Titanic's End Dynamic.run.xml
```diff
@@ -1,10 +1,11 @@
...
   <configuration default="false" name="Titanic's End Dynamic" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="temurin-17.0.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="openjdk-18" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="LXStudio-TE" />
-    <option name="PROGRAM_PARAMETERS" value="dynamic" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
+    <option name="PROGRAM_PARAMETERS" value="dynamic --resolution 320x320" />
+    <option name="VM_PARAMETERS" value="-ea -Djava.awt.headless=true " />
```
andrewlook, line 8: just a reminder not to commit this run config please

zranger1:
### Current status
- All single pass shaders working, including those that use their own textures
- Shaders that use iBackground (like Fireflies) working, with minor change to background pixel fetch
- Multipass shaders partially working (see issues section below)
- Shader effects (Explode and some NDI) not working (see issues)

#### Issues and Work Items
**The biggest thing**:  Texture mapping encodes adjacency - the spatial relationship between pixels. In the current shader engine, it is trivial to sample a pixel immediately up, down, left, right, etc..   This encoded relationship doesn't exist in LX model coordinates, and it makes doing convolution and similar things that rely on manipulating nearby pixels much harder. 

We don't use a lot of this kind of algorithm in TE shaders, so it's not a showstopper.  But it does impact a few important pieces, so I'm working on a workaround -- a "fake" texture mapping system that shaders that need to work this way can use.   In the meantime, the global Explode effect doesn't work, and some aspects of NDI may do weird things too.

Adding automatic calculation of model aspect ratio and minimum enclosing rectangle on model change.   This will let us add and remove models as we like, while still rendering the minimum possible number of pixels, and drawing circular circles and square squares.

This likely means we'll wind up with an physical render resolution (which will be a small rectangle or square) and an effective resolution (with applied aspect ratio correction) that will be presented to the shaders.  Working on figuring out how to make sure this is as simple and hands-free as possible.

.run/Titanic's End Dynamic.run.xml
zranger1: This one kinda needs the --resolution 320x320 flag atm.   I'll fix the file as soon as I get the automatic buffer size calc checked in!

zranger1:
### Current Status 8/19/2024
- **This branch is now usable for Mothership.**  Just don't use the global Explode effect and any NDI input cases that involve mapping (see the previous status comment for details)
- Default resolution is set to 320x320, which allows up to 102,400 pixels at very high performance. This is plenty of pixels for Mothership, but use the --resolution command line argument if you need more for a combined model.

zranger1:
### Current Status 8/22/2024
This branch is now generally useable with the following conditions:
- The **'Twist'** effect which swapped output x and z axes is **not available** in this configuration.  The control button has been disabled.
- The global **'Explode' effect and NDI input and output are working**
- The default resolution is set to **320x320, which allows up to 102,400 pixels.**  Remember to use the `--resolution` command line argument if you need run a combined model with more pixels.
-  **I have not been able to test with the latest TouchDesigner integrations** though, so unless you can test thoroughly, I'd still recommend running the main branch on lighting-1 for normal TE operations.

Barring major problems, this is where I'll leave this branch 'till post-burn.   If you can use it on the main car, you'll see sharper images and smoother motion, so if you can get the TD work tested, it's definitely worth trying.

.run/Titanic's End Dynamic.run.xml
andrewlook: ok great! maybe it's worth committing another run config specifically for mothership?

zranger1: We shouldn't need a new run config. The default setting now has enough pixels for either car.  If you're running both TE and Mothership on the same instance of LX though, you'll need more.   In that case, I'd do something like `--resolution 640x320` b/c it has about the right aspect ratio for a two model setup w/a little space between the models.

src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -0,0 +1,315 @@
...
+  private static final int FIRST_UNRESERVED_TEXTURE_UNIT = 3;
+  private int nextTextureUnit = FIRST_UNRESERVED_TEXTURE_UNIT;
+
+  // map of texture hash codes to texture data
+  // The hash code uniquely identifies the texture, the texture unit number is
+  // used to bind the texture to the GPU, and the other data is used for
+  // cache management. NOTE: Used ConcurrentHashMap here because the normal
+  // HashMap doesn't let you remove elements while iterating over it, which
+  // we need to do when dealing with model changes.
+  private final Map<Integer, CachedTextureInfo> textures = new ConcurrentHashMap<>();
```
jkbelcher, line 54: @zranger1 the iterate/remove operation is single threaded, right?  If so it seems like we could avoid the ConcurrentHashMap here.

zranger1: Thanks, @jkbelcher!   
I used ConcurrentHashMap because normal HashMap's iterator really wouldn't do the remove-while-iterating thing at all.   It wasn't a multithreading thing, just that the obvious thing didn't work, so I took the next-most-obvious thing that did.

In retrospect,  I think I was thinking about it too narrowly -- since access to this is single threaded, a better, faster way would be to use a non-final normal hash map, and on model changes just build a new one using the elements of the old one that we want to keep.  That's normally going to be a very short list - fewer than 10 items, usually only 2 or 3.    Will make this change in the next pass.

zranger1: Revised to use normal HashMap in 25583a1

andrewlook (approved):
I've been testing this and I think we're pretty much ready to merge it! Just wanted to say thanks to @zranger1 and @jkbelcher for the monumental effort here and the huge leap forwards in our shader stack!

PR #539, opened 2024-08-14
Palette mgr cleanup: Director APC pushes once (remove "pinned" UI), and eliminate complementary/triadic options

andrewlook:


zranger1 (approved):
I'll have to trust you on the actual APC, but I do like the new immediate push behavior!

PR #541, opened 2024-08-14
Mothership - Startup looks + new shaders

zranger1:
A set of "looks" that we can use for Mothership startup, and a couple of new/modified patterns especially for Mothership.

The startup looks are in the project file MShipStartupLooks.lxp.
They're stored as named snapshots (below the palette in the **Global** panel)

This project has only the view-facing side of the Mothership lit, which I think will give a more accurate view of how patterns will look in the field.   Also, for best viewing in Chromatik, go to the **Camera** section at the bottom of the **Model** panel and change the **Point Size** from 3.00 to 1.5.   Again, I think this will give a more realistic picture of how the LEDs on Mothership will look.   (No warranty here though -- can't be certain 'till we see it light up!) 

There are has two channels:
- A composite channel containing all the patterns needed to create the looks
- An effects channel, which is set up with SpaceExplosionFX to make the cyan laser pew (you can do it at any point by pressing **WowTrigger** on that channel).

(Video will be available on TE Slack shortly)

### DIFF
--- a/Models/Mothership_car.lxm
+++ b/Models/Mothership_car.lxm
@@ -1,6 +1,6 @@
 {
-  "version": "1.0.1.TE.1-SNAPSHOT",
-  "timestamp": 1723319742744,
+  "version": "1.0.1.TE.2-SNAPSHOT",
+  "timestamp": 1723666671027,
   "fixtures": [
     {
       "jsonFixtureType": "Mothership/window/Ramp",
@@ -107,7 +107,7 @@
         "enabled": true,
         "brightness": 1.0,
         "identify": false,
-        "mute": false,
+        "mute": true,
         "solo": false,
         "tags": "",
         "fixtureType": "Mothership/window/Slice01P"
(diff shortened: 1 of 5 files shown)


Models/Mothership_car.lxm
```diff
@@ -1435,7 +1435,7 @@
         "enabled": true,
         "brightness": 1.0,
         "identify": false,
-        "mute": false,
+        "mute": true,
```
andrewlook, line 1438: what do these "mute" params mean? I'm assuming @jkbelcher will be tweaking stuff in this `lxm` file so not a blocker, just wanted to check my understanding

andrewlook (approved):
these are looking fantastic!

Models/Mothership_car.lxm
zranger1: Good catch!   We'll want to unmute those for performance.   I turned off the back half of the Mothership model while writing patternsto get closer to seeing what people looking at the real Mothership will see.  I found developing with the whole model enabled to be a little deceptive b/c you get light from LEDs in the back that you definitely wouldn't be able to see.

zranger1: And this almost certainly isn't the final Mothership model.  There are a bunch of segments with incomplete IP address in there too.  More work for @jkbelcher!

PR #543, opened 2024-08-15
remove giggle pixel UI, deprecate giggle code, move autopilot UI to bottom

andrewlook:
image


zranger1 (approved):
Tested & working - Global panel w/new ordering looks great.

PR #546, opened 2024-08-16
Add FxLaserCharge Mothership pattern

zranger1:
#### New FxLaserCharge pattern for Mothership 
Shows a sparkly laser charging progress ring, eerily similar to @jvyduna 's demo.  

- Easy to control manually or via OSC.   Just a one shot press of the WowTrigger control starts the event which then runs to completion. 
- Wow1 controls the (exponential) curve shape
- Wow2 controls the duration of the charge event in seconds.
- Uses the pattern control color, respects the global palette.

#### Small but Useful Fixes Department:
Also added check when drawing panel backings that improves performance on models that have no panel backings.  Previously, this could cause an an 'invisible' exception that Chromatik handled with out reporting ON EVERY FRAME.  UGH!

### DIFF
--- a/.run/Titanic's End Dynamic.run.xml
+++ b/.run/Titanic's End Dynamic.run.xml
@@ -1,10 +1,11 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Titanic's End Dynamic" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="temurin-17.0.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="openjdk-18" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="LXStudio-TE" />
-    <option name="PROGRAM_PARAMETERS" value="dynamic" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
+    <option name="PROGRAM_PARAMETERS" value="dynamic --resolution 640x640" />
+    <option name="VM_PARAMETERS" value="-ea -Djava.awt.headless=true " />
     <method v="2">
       <option name="Make" enabled="true" />
     </method>
--- a/.run/Titanic's End Dynamic.run.xml
+++ b/.run/Titanic's End Dynamic.run.xml
@@ -1,11 +1,10 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Titanic's End Dynamic" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="openjdk-18" />
-    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
+    <option name="ALTERNATIVE_JRE_PATH" value="temurin-17.0.8" />
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="LXStudio-TE" />
-    <option name="PROGRAM_PARAMETERS" value="dynamic --resolution 640x640" />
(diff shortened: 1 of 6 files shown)


.run/Titanic's End Dynamic.run.xml
```diff
@@ -1,10 +1,11 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Titanic's End Dynamic" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="temurin-17.0.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="openjdk-18" />
```
andrewlook, line 3: ah maybe we don't want to commit this run config just to be safe? I haven't checked on the latest JDK support (would be nice if we could use something newer than Temurin) but maybe we keep the old one just to be safe?

src/main/java/titanicsend/pattern/jon/FxLaserCharge.java
```diff
@@ -0,0 +1,87 @@
...
+  GLShaderFrameSetup setup =
+      new GLShaderFrameSetup() {
+        @Override
+        public void OnFrame(GLShader s) {
+          // state of explosion visual
+          double progress = 0;
+          double eventDuration = 2 * getWow2();
+
+          // If the WowTrigger button is pressed and we're not
+          // already running, start the event
```
andrewlook, line 64: nice! is the idea that it starts running on its own and waits a while after going black, so wowtrigger can kick it off explicitly at the right time?

andrewlook (approved):
Tested locally and it's working great! left a couple comments but ready to merge whenever!

.run/Titanic's End Dynamic.run.xml
zranger1: Yeah, that file shouldn't even be in there.  I'll fix it.   It's a Windows thing: I have to change the JVM and get rid of the --XStartOnFirstThread flag every time I check out a branch.   I just forgot to discard it before checkin.

src/main/java/titanicsend/pattern/jon/FxLaserCharge.java
zranger1: It's always running, doing essentially nothing 'till you press WowTrigger.   (As long as the "progress" variable in Java is 0, only black will be displayed.)   When WowTrigger is pressed, the elapsed time starts changing, which feeds the exponential curve formula.   It runs 'till the output value reaches 1, then sets everything back to the idle state.   I love tiny state machines!

.run/Titanic's End Dynamic.run.xml
andrewlook: no worries, thanks for the fix!

src/main/java/titanicsend/pattern/jon/FxLaserCharge.java
andrewlook: oh cool! for me, i think it started running right upon file open, before I actually pressed wowTrigger. is that expected? (maybe the MShipLooksJeff showfile had some active state in it, that was causing it to auto-start?) I can double-check if that's how it behaved - just wanted to double-check if it's the type of thing we want to be silent until it's explicitly triggers.

andrewlook: confirmed offline that when I delete the pattern and re-add it to the channel, it doesn't start until I hit wowTrigger 👍

andrewlook (approved):
tested locally! merging now.

src/main/java/titanicsend/pattern/jon/FxLaserCharge.java
zranger1: I think the issue is that there are two copies of this pattern -- the oneJeff made in the "composite" channel, which runs constantly, and mine, which lives in the "effects" channel and does nothing 'till you press WowTrigger1 in that channel.  They look very much the same.  To test, just disable the "composite" channel, and try the one in "effects."

PR #547, opened 2024-08-16
Beacons for dynamic model + Gamepad engine

jkbelcher:
DmxEngine (ie Beacons) modified to work with dynamic model.  DmxModel class can be specified in an LXF, therefore beacons can be added and removed like other fixtures.  There's an intricate tracking of LX buffers that creates DMX buffers and makes them available to patterns, so in theory beacon patterns can be run on multiple channels and blended.  The blending has not been extensively tested and I expect for this year we'll run a single dedicated Beacons channel.

Added an example pattern where beacon color is synced to the global palette.  Demo video shows color following our Director midi controller.

Implemented the Director fader for beacons into beacon models.  This happens behind the scenes of the pattern where the channel fader is applied to the DMX fixture's Dimmer parameter.

Added a Gamepad engine to flush out on the gamepad support we added to Chromatik this summer.  The BeaconGame pattern demonstrates gamepad control and color sync.

Of note the DmxEngine is still a very early version.  It should work for this year but still needs refinement and polish.

[https://www.youtube.com/watch?v=MFjWLs3A_zY](https://www.youtube.com/watch?v=MFjWLs3A_zY)

### DIFF
--- a/src/main/java/titanicsend/dmx/DmxEngine.java
+++ b/src/main/java/titanicsend/dmx/DmxEngine.java
@@ -25,6 +25,7 @@ import heronarts.lx.mixer.LXChannel;
 import heronarts.lx.mixer.LXGroup;
 import heronarts.lx.mixer.LXMasterBus;
 import heronarts.lx.mixer.LXMixerEngine;
+import heronarts.lx.model.LXModel;
 import heronarts.lx.output.LXOutput;
 import heronarts.lx.studio.TEApp;
 import java.util.ArrayList;
@@ -33,10 +34,13 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+
+import heronarts.lx.utils.LXUtils;
 import titanicsend.app.dev.DevSwitch;
 import titanicsend.dmx.model.DmxModel;
 import titanicsend.dmx.model.DmxWholeModel;
 import titanicsend.dmx.parameter.DmxParameter;
+import titanicsend.model.TEWholeModel;
 
 /**
  * DmxEngine manages the dmx buffers and mixes the dmx frames. Much of the heavy lifting here is a
@@ -45,7 +49,10 @@ import titanicsend.dmx.parameter.DmxParameter;
  * Roles: - Maintains a reference of one DmxModelBuffer per LXBuffer for consumption by
  * patterns/effects. - Acts as a DmxMixer
  */
-public class DmxEngine implements LXLoopTask {
+public class DmxEngine implements LXLoopTask, TEWholeModel.TEModelListener {
+
+  private static final String META_TAG_DMX_FIXTURE = "dmxFixture";
+  private static final String META_KEY_DMX_CLASS = "dmx";
 
   private static final boolean ENABLE_DEBUG = false;
 
@@ -58,6 +65,8 @@ public class DmxEngine implements LXLoopTask {
   protected final LX lx;
 
   protected final DmxWholeModel dmxWholeModel;
+
+  protected final TEWholeModel modelTE;
 
   private final MixerListener mixerListener = new MixerListener();
 
@@ -87,6 +96,8 @@ public class DmxEngine implements LXLoopTask {
     this.lx = lx;
 
     this.dmxWholeModel = TEApp.wholeModel;
+    this.modelTE = TEApp.wholeModel;
+    TEApp.wholeModel.addListener(this);
 
     // Set up global add blend
     this.addBlend = new AddBlend(lx);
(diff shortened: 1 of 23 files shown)


src/main/java/titanicsend/dmx/DmxEngine.java
```diff
@@ -696,4 +627,149 @@ public static void debug(String location, DmxBuffer[] dmxAll) {
...
+    public Frame(DmxWholeModel model) {
+      setModel(model);
+    }
+
+    public void setModel(DmxWholeModel model) {
+      if (model instanceof DmxWholeModel) {
+        DmxWholeModel dmxWholeModel = (DmxWholeModel) model;
+
+        this.model = dmxWholeModel;
+        //if (this.main == null) {
```
andrewlook, line 667: just curious, what's the difference if this is commented out vs. not?

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
```diff
@@ -0,0 +1,190 @@
...
+        addParameter("Prism1", this.prism1);
+        addParameter("Prism1Rotate", this.prism1rotation);
+        addParameter("Prism2Rotate", this.prism2rotation);
+        addParameter("Focus", this.focus);
+        addParameter("Shutter", this.shutter);
+        addParameter("Dimmer", this.dimmer);
+        addParameter("Frost1", this.frost1);
+        addParameter("Frost2", this.frost2);
+        addParameter("ptSpd", this.ptSpeed);
+        addParameter("Control", this.control); // Use caution!
```
andrewlook, line 55: why the need for caution with this one?

src/main/java/titanicsend/gamepad/GamepadEngine.java
```diff
@@ -0,0 +1,398 @@
...
+    // To get the latest mappings, download a new copy of
+    // https://github.com/mdqinc/SDL_GameControllerDB/blob/master/gamecontrollerdb.txt
+
+    // Load gamepad mappings from a file
+    String mappings = loadGamepadMappingsFromFile();
+
+    if (!LXUtils.isEmpty(mappings)) {
+      ByteBuffer mappingBuffer = MemoryUtil.memUTF8(mappings);
+
+      if (glfwUpdateGamepadMappings(mappingBuffer)) {
```
andrewlook, line 66: worth maybe wrapping this in a try/catch? to be clear I don't fully understand what's going on under the hood here, but looks like the kind of thing that has the potential to throw errors

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
```diff
@@ -0,0 +1,190 @@
...
+
+        this.gamepad = TEApp.gamepadEngine.createGamepad();
```
andrewlook, line 57: Is there a reason this pattern needs to create its own `Gamepad` instance, rather than using some kind of global singleton initialized in `TEApp.gamepadEngine`? Like would this get weird if two instances of the pattern create their own gamepad instances, etc?

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
```diff
@@ -0,0 +1,190 @@
...
+    private static final int ptSpeedTick = 5;
+
+    @Override
+    public void onGamepadParameterChanged(LXParameter p) {
+      if (p == this.gamepad.axisLeftX) {
+        gamepadAxisToBeaconParameter(this.pan, this.gamepad.axisLeftX);
+      } else if (p == this.gamepad.axisRightX) {
+        gamepadAxisToBeaconParameter(this.tilt, this.gamepad.axisRightY);
+      } else if (p == this.gamepad.dpUp) {
+        if (this.ptSpeed.getDmxValue() < (225-ptSpeedTick)) {
```
andrewlook, line 77: what's the significance of 225 here?

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
```diff
@@ -0,0 +1,190 @@
...
+
+    private int shutterValue = 32;
+
+    @Override
+    public void run(double deltaMs) {
+        float rightTrigger = this.gamepad.axisRightTrigger.getNormalizedf();
+        if (rightTrigger < .1) {
+            shutterValue = 32;
+        } else {
+            shutterValue = (int) ((rightTrigger - .1) * (95-75) / (1 - .1) + 75);
```
andrewlook, line 102: may want to resolve some of the fixed computations here and/or add parentheses to make order of ops super clear

andrewlook (approved):
This is looking great! I added some non-blocking comments (mostly questions to make sure I understand). But ready to merge this whenever!

I tested locally to make sure LX runs. Added the beacon fixtures to see what happens - it seems like they're positioned around the origin - my guess is that doesn't matter since we're not going to rely on a chromatik render to see what they're doing, just curious if there's anything in particular we'll want to do when adding them to the Mothership showfile.

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
jkbelcher: DMX values are 0-255.  The way these fixtures use those values are all over the map.  In this case 225 is the max value for the range.  There may be values above it that are special.

jkbelcher: Ideally special values like this get pulled into public static variables on the class.  Just no time at the moment.

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
jkbelcher: Sure, fine with more parentheses here.  I'm preferring to leave these magic numbers typed out so you can at least read them and compare them to the DMX profile sheet to see what's going on.  On a cleanup sometime they could be changed for public static.

src/main/java/titanicsend/gamepad/GamepadEngine.java
jkbelcher: Yep, good call.  For all I know it throws an error if the db is corrupt.

src/main/java/titanicsend/dmx/DmxEngine.java
jkbelcher: In static world the top-level models don't change so comparing main to null was a way to see if the buffers were initialized.  In dynamic world it needs to re-create everything when the model is changed.  I didn't yet think through the impact of taking this out on Static model world, so left it in as a reminder.

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
jkbelcher: Control channels on DMX fixtures can do wild stuff including configuration changes.  The value on them should not be changed unless you're going to something for a particular reason.

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
jkbelcher: There are 16 SystemGamepad instances to match the 16 slots provided by GLFW.  The Gamepad engine is a global singleton that creates those and they exist for the duration of runtime.    The Gamepad instance created here in the pattern is like a proxy.  You can point it at a different slot by changing the Input number.  This makes it really convenient to initialize one or more gamepads from a pattern and then retarget them without moving around your listeners.

This was a fast implementation and it would be fun to pry at it eventually and see if there are better methods.  I was going for the idea similar to a game console, where everyone picks up a controller and there's a dialog "player 1 press start"... and pressing start identifies the first controller.  "player 2 press start"... now that controller index is identified as player 2.    It's nice to reassign them in the game without having to swap physical controllers with the person next to you.

jkbelcher:
> Added the beacon fixtures to see what happens - it seems like they're positioned around the origin - my guess is that doesn't matter since we're not going to rely on a chromatik render to see what they're doing, just curious if there's anything in particular we'll want to do when adding them to the Mothership showfile.

Yes we can set the X,Y,Z of the beacon fixture to be anywhere.  We'll set them to match their position on the car in case we want the color to come from the pattern as it passes over that spot.

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
andrewlook: makes sense 👍

src/main/java/titanicsend/dmx/DmxEngine.java
andrewlook: thanks for clarifying! seems like we're 100% committed to dynamic model for the burn anyways

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
andrewlook: got it, thanks 👍

src/main/java/titanicsend/gamepad/GamepadEngine.java
andrewlook: makes sense - at least that'll blow up right at startup time so it wouldnt be a surprise mid-show

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
andrewlook: awesome! cant wait to try this out. very handy that the input numbers are configurable

src/main/java/titanicsend/dmx/pattern/BeaconGamePattern.java
andrewlook: all good, makes total sense!

PR #549, opened 2024-08-16
Added single panel model

sina-cb:
This PR add a new optional model to the ModelFileWriter which is used for the random strobe effect from touchdesigner.


andrewlook (approved):
Nice! thanks for adding the documentation - I was a bit confused by the overloaded method calls but the docstring really helps explain things.

I ran this locally, executed the `ModelFileWriter` in a channel and it worked. It did cause a significant lag, so I'm thinking we may want to comment it out of the pattern registry before showtime just to be safe - but it seems like it's working as expected!

Merging now.

PR #550, opened 2024-08-16
Introducing the TD Stable Diffusion Pattern

sina-cb:
This pattern is a direct derivative of the TDNdiPattern class, but it also handles OSC communication between the pattern and the accompanying TouchDesigner project.

The Stable Diffusion pattern will do the following in order to control the stable diffusion generation performed in TouchDesigner:
1. Using Wow1 parameter to switch between noises
2. Using Wow2 parameter to switch between prompts
3. Handles registration and deregistration of up to 2 stable diffusion patterns in the Chromatik channels.

The other common parameters will be used to perform the common manipulations such as spin, angle, x, y.

TODO Improvements:
- Currently the view will be used to change the overall color theme of the pattern, this can be replaced with a different parameters


src/main/java/titanicsend/osc/CrutchOSC.java
```diff
@@ -35,497 +35,512 @@
 import heronarts.lx.parameter.LXParameter.Polarity;
 import heronarts.lx.parameter.LXParameterListener;
 import heronarts.lx.pattern.LXPattern;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
 public class CrutchOSC extends LXComponent
-    implements LXOscComponent, LXMixerEngine.Listener, LXPalette.Listener {
```
andrewlook, line 43: looks like there was a 2 space -> 4 space change here, which makes it hard to read this diff - would you mind running `mvn spotless:apply` before these changes (you can just commit this file and reset the rest, since lots of our files are un-linted right now)?

sina-cb: Fixed it

andrewlook (approved):

PR #552, opened 2024-08-17
shader: stem-aware TriangleInfinity

andrewlook:
Version of TriangleInfinity with the geometry simplified, to avoid the mid-screen discontinuity of TriangleInfinity.

There are a controllable number of triangles (iQuantity), and the iteration of triangle mod 4 determines which stem get multiplied into it.

Controls
- Wow1 controls the level of stem overrides (wow1 = 0 means stems are ignored)
- Wow2 controls the rotation of the inner triangles (hooking it up to a stem modulator is fun)
- FreqReact - how much to use the raw waveform
- LevelReact - bass reactivity
- WowTrigger - makes the animation get trippy


zranger1 (approved):
Looks great! I thought it was particularly good on faster tunes w/speed and quantity turned up a bit, and stem reactivity on full.  This should be fun!

PR #553, opened 2024-08-17
Mothership driving pattern

jkbelcher:
Dedicated driving pattern.  Uses two colors: The first color is intended for headlights.  The second color is to be visible while keeping reflectivity low behind the driver.

Add the tag "drive" to fixture files for color1, and the tag "run" as in "running lights" to fixture files for color2.

We'll choose the windows on playa.

I also added a TECategory class and adjusted the patterns tagged "Utility".  Sorry this should have been in a different PR, just got carried away.

### DIFF
--- a/Fixtures/Mothership/window/Slice02P.lxf
+++ b/Fixtures/Mothership/window/Slice02P.lxf
@@ -46,13 +46,13 @@
   ],
 
   "components": [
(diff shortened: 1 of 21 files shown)


jkbelcher:
image

zranger1 (approved):
Having the base two-color pattern gives extra flexibility.  Nice!

PR #554, opened 2024-08-17
Don't reset camera if it is already correct for dynamic model

jkbelcher:
Camera was resetting on every file open, unnecessary.


andrewlook (approved):
oh hell yes, thank you! this has been annoying me but I hadn't known where this was happening

PR #555, opened 2024-08-18
Director Effect: auto-add and lock

jkbelcher:
Monitors for file changes.  After a New file or Open file operation the DirectorEffect is added if it was missing.  Then Enabled and Lock states are turned on.

This should be enough to keep a user from accidentally deleting the effect.  I'm refraining from the Absolute level where we re-create the effect if the current instance is deleted, then we wouldn't have a way around it in a special circumstance.

### DIFF
--- a/src/main/java/titanicsend/app/director/Director.java
+++ b/src/main/java/titanicsend/app/director/Director.java
@@ -1,16 +1,18 @@
 package titanicsend.app.director;
 
+import java.io.File;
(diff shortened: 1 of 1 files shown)


andrewlook (approved):
tested locally, works great! Thanks for thinking ahead here, particularly if we're shuffling showfiles around last minute I think this will really come in handy.

PR #557, opened 2024-08-20
AutopilotJKB baseline

jkbelcher:
Legacy autopilot system from Joule.  I'd like to merge this for reference and then make TE-specific changes in a future PR.

This is just the framework, it doesn't add components to the runtime yet.


zranger1 (approved):
Code is clean and in this PR, doesn't impact the operation of TEApp.   Looking forward to seeing it work -- this seems about the perfect level of not-too-distracting reactivity for Mothership.

PR #558, opened 2024-08-21
SuperMod midi controller

jkbelcher:
Updated the SuperMod midi controller to use audio stems modulators.  

Imports the supermod repo:
https://github.com/jkbelcher/SuperMod/tree/te

Tested as working with midi controllers here.  To test you'll need an APCmini and a MFT.

This is going to be fun to use!  Expect we can adjust the button assignments on site as needed.


src/main/java/titanicsend/audio/AudioStemModulator.java
```diff
@@ -122,10 +123,12 @@ public AudioStemModulator(String label) {
 
     addParameter("stem", this.stem);
     addParameter("emaMs", this.emaMs);
-    addParameter("noiseMode", this.noiseMode);
+    addParameter("outputMode", this.outputMode);
     addParameter("accRate", this.accRate);
     addParameter("shape", this.waveshape);
     addParameter("slope", this.slope);
+
+    addLegacyParameter("noiseMode", this.outputMode);
```
andrewlook, line 131: why do we need to keep `noiseMode` around?

jkbelcher: to avoid breaking any presets Will has exported for BM

andrewlook (approved):

PR #561, opened 2024-08-23
Fixed race condition between NDI feed and the NDI patterns

titanicsend:
Fixed race condition between NDI feed and the NDI patterns.

Previously we were using a buffer shared by NDI and the NDI patterns in LX. This caused race conditions between the NDI and LX pattern both trying to access the FrameData buffer.

This PR makes a copy of the frame data and uses that to set the frame if there is no new data received on NDI.


jkbelcher (approved):
This adds a buffer for the incoming pixel data for a "hold last frame" effect.  I wouldn't call this a race condition as it wasn't multiple threads operating on one resource.

PR #562, opened 2024-08-23
Added new project file for BM24

worldveil:
Basically just reduced to single of each High, Lo, and Edges and also loaded the User preset library I created from justin's new feature.


andrewlook (approved):
Solid!

PR #567, opened 2024-08-24
Simplify Effect

jkbelcher:
Wrote this in the fall, tuned it up just now to work with color.

Gain control will need follow-up work.


jkbelcher:
https://www.youtube.com/shorts/z8CwAh7HpmY

worldveil (approved):

PR #570, opened 2024-08-25
Update GLX to fix UI2dContainer.removeAllChildren()

jkbelcher:
Latest GLX from upstream.  Attempting to fix errors Will is seeing when applying user presets.


fakeplant (approved):
approved by look after local testing

PR #572, opened 2024-08-26
Added bootup channel template project file

sina-cb:
This PR adds a new project which has a template set of channels for the bootup sequence.

The bootup sequence is using 2 extra channels now: 1. BOOTUP-NDI 2. BOOTUP

This is to allow us to fade between the NDI and the shader patterns during the bootup sequence.


andrewlook (approved):
tested locally - will add these channels to BM2024_{TE, Mship}.lxp

PR #589, opened 2024-11-05
Location texture caching for 3D branch

zranger1:
#### Adds long-promised caching of view coordinate textures to the 3D branch.
With this PR, the 3d branch is ready to merge with main.  What it does:
- Uses a **lot** less memory, particularly when using many channels and patterns, since the coordinate texture for each view is stored only once (instead of per-pattern as in the current implementation).
- Much faster view switching, since view coord textures, once loaded, are kept resident on the GPU.  No coordinate data has to be copied when you switch views.
- Lower CPU load because it eliminates quite a lot of the data movement involved in setting up the coordinate textures.
- lower GPU load.  Less loading/unloading of textures, plus the shader engine is now able to avoid computing pixel values for points not in the view.  (Previously, it ran the pattern fragment shader over its whole internal buffer.)  Depending on what the pattern is doing, this can be a huge win for sparse views, like "Edges".

Notes for the Future:
**Question**:  It is now possible, with very little performance impact, to support full 3d rotation/translation/scale/skew operations on the model coordinates.  This is something that has come up a few times as a feature request in the Chromatik forums.  Certainly it'd be fantastic for working with physically animated sculptures.  I didn't do it here, b/c I'm not sure (a) we have a great use case yet, and (b) how the heck we'd control it.   Do we want this feature in, just on general principles?

### DIFF
--- a/src/main/java/titanicsend/pattern/glengine/TextureManager.java
+++ b/src/main/java/titanicsend/pattern/glengine/TextureManager.java
@@ -5,6 +5,8 @@ import com.jogamp.opengl.util.GLBuffers;
 import com.jogamp.opengl.util.texture.Texture;
 import com.jogamp.opengl.util.texture.TextureData;
 import com.jogamp.opengl.util.texture.TextureIO;
+import heronarts.lx.LX;
+import heronarts.lx.model.LXModel;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -0,0 +1,315 @@
...
+    // Sanity check: make sure the system resolution has enough points
+    // to hold the model.
+    // This should really never fail, but we're checkin' anyway.
+    if (numPoints > maxPoints) {
+      LX.error(
+        "GLEngine resolution ("
+          + maxPoints
+          + ") too small for number of points in the model ("
+          + numPoints
+          + ")");
```
andrewlook, line 153: if we hypothetically did hit this error, should we actually bail out of the function / return early, or is okay to just log it and continue executing?

src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -0,0 +1,315 @@
...
+    // return the texture's information block
+    return t.textureUnit;
+  }
+
+  /**
+   * Create a coordinate texture from the normalized coordinates of the given model
+   * and bind it to the next available texture unit.  If the view's coordinate texture
+   * already exists,return the bound texture unit number.
+   */
+  public CachedTextureInfo createCoordinateTexture(LXModel model) {
```
andrewlook, line 109: just so I understand, will this get called once per chromatik "view" (with a model containing only the relevant subset of points)?

or will this get called once per full chromatik model (set of fixtures), and do we somehow filter the concept of chromatik views another way?

andrewlook (approved):
This is awesome, thank you so much for doing this!

So if I understand correctly, the texture for model coords will always be as long as `model.points.length * 3`, and for the "default view" (all points), it's laid out like:

```
x1  y1  z1  x2  y2  z2 ...
```

The number of points stays static unless we load/unload a project or load/unload a model file/fixture file.

Suppose a we select "edges" view, and if point 2 is part of that view but point 1 was not, it would look like:

```
NaN NaN NaN x2  y2  z2 ...
```

and in the main method of `_template.fs`, we detect the empty coords and exit early, saving ourselves the compute.

On the backend, we keep the textures loaded into GPU, but we swap in the active view as a uniform into `lxModelCoords`, which treats it as a `sampler2D`.

Is my understanding of that roughly right?

zranger1:
@andrewlook, yes that's pretty much the idea.  Here's more detail:
- since we can't know how many points a project has 'till it's loaded, and even then it can be changed in real-time  by editing the model in Chromatik,  the maximum number of points is controlled by the --resolution parameter.   By default it is currently 102400 points (320x320).  This is enough for either TE or the Mothership w/room to spare (but not both at the same time).
- a view's "active" points are always loaded first in the texture (starting at index 0) in order corresponding with their order in the Chromatik points list.  This makes transfer from shader to Chromatik much easier.    
- the NaNs, if any, wind up at the end of the buffer.  This is good from a performance standpoint, since the GPU isn't really totally parallel.  It grabs tasks to process in groups.  If all the tasks in a group finish at wildly different times, some of the cores are left idle and can't be rescheduled.   So we try to keep all the long tasks and all the short tasks grouped together.
- You've got it -- In OpenGL, textures loaded to the GPU and "ready to go" are associated with Texture Units.   To use these in a shader, all you have to do is set up a sampler2D, and at frame time, tell it via the uniform which texture unit to associate with that sampler.

PR #590, opened 2024-12-06
3D Branch Docs + Cleanup

zranger1:
Mostly about the docs, plus a couple of efficiency improvements that I've been wanting to do for a while.

### Documentation Update
Shader documentation for authors in /resources/docs/ShaderFrameWork.md has been updated with all the
3D renderer's changes and enhancements.

### Refactored Audio Calculations
Most of the audio data used by shaders and other patterns is now updated in the engine, once per frame, rather than
being recalculated for every running pattern on every frame.   

### Implemented Shared Uniform Blocks
Another time/memory saver. Uniform blocks allow all shaders to share a single set of uniforms - in this case two blocks are shared:

- PerRunUniforms:  Things like resolution that are set at startup and persist through an entire run.
- PerFrameUniforms:  In our case, mostly audio data, which is only updated once per frame but is needed by all shaders.

Previously each shader needed its own copy of this data, set as a uniform variable.  It's not really that much data, but it winds up being a lot of per-shader function calls to collect that data and move it to the GPU.  (I think it wound up being around 12 variables, w/2 function calls each to set up the uniforms. So for each shader, at 60 fps, that's 1440 function calls/sec that we're no longer doing.)

With uniform blocks, the data is only moved once per frame, as a unit.  No per-frame work at all is required from the individual shaders.

### DIFF
--- a/.run/Titanic's End Dynamic.run.xml
+++ b/.run/Titanic's End Dynamic.run.xml
@@ -4,7 +4,7 @@
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="LXStudio-TE" />
     <option name="PROGRAM_PARAMETERS" value="dynamic Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
+    <option name="VM_PARAMETERS" value="-ea -Djava.awt.headless=true " />
     <method v="2">
       <option name="Make" enabled="true" />
     </method>
--- a/.run/Titanic's End Dynamic.run.xml
+++ b/.run/Titanic's End Dynamic.run.xml
@@ -1,10 +1,11 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="Titanic's End Dynamic" type="Application" factoryName="Application">
-    <option name="ALTERNATIVE_JRE_PATH" value="temurin-17.0.8" />
+    <option name="ALTERNATIVE_JRE_PATH" value="temurin-17" />
+    <option name="ALTERNATIVE_JRE_PATH_ENABLED" value="true" />
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="LXStudio-TE" />
     <option name="PROGRAM_PARAMETERS" value="dynamic Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -Djava.awt.headless=true " />
+    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
     <method v="2">
       <option name="Make" enabled="true" />
     </method>
--- a/resources/shaders/framework/template.fs
+++ b/resources/shaders/framework/template.fs
@@ -2,29 +2,34 @@
 
 out vec4 finalColor;
 
-// standard shadertoy
+// shared uniforms that are set once per run
+layout (std140) uniform PerRunBlock {
+  vec4 iMouse;
+  vec2 iResolution;
+};
+
+// shared uniforms that are set once per frame
+layout (std140) uniform PerFrameBlock {
+// Legacy TE Audio uniforms
+    uniform float beat;
+    uniform float sinPhaseBeat;
+    uniform float bassLevel;
+    uniform float trebleLevel;
+
+    uniform float volumeRatio;
+    uniform float bassRatio;
+    uniform float trebleRatio;
+
+// Values from audio stem splitter
+    uniform float stemBass;
+    uniform float stemDrums;
+    uniform float stemVocals;
+    uniform float stemOther;
+};
+
+// standard shadertoy uniforms (updated per frame for each
+// running pattern.)
 uniform float iTime;
-uniform vec2 iResolution;
-uniform vec4 iMouse;
-
-// TE Audio
-uniform float beat;
-uniform float sinPhaseBeat;
-uniform float bassLevel;
-uniform float trebleLevel;
-
-uniform float volumeRatio;
-uniform float bassRatio;
-uniform float trebleRatio;
-
-uniform float levelReact;
-uniform float frequencyReact;
-
-// Current values from audio stems
-uniform float stemBass;
-uniform float stemDrums;
-uniform float stemVocals;
-uniform float stemOther;
 
 // TE Colors
 uniform vec3 iColorRGB;
@@ -43,6 +48,8 @@ uniform float iBrightness;
 uniform float iWow1;
 uniform float iWow2;
 uniform bool iWowTrigger;
+uniform float frequencyReact;
+uniform float levelReact;
 
 // Shadertoy audio channel
 uniform sampler2D iChannel0;
@@ -158,7 +165,7 @@ void main() {
     finalColor = _blendFix(finalColor);
     #else
     // Old EDC blending - force black pixels to full transparency, otherwise
-    //use shader provided alpha
+    // use shader provided alpha
     finalColor.a = ((finalColor.r + finalColor.g + finalColor.b) == 0.0) ? 0.0 : finalColor.a;
     #endif
     #endif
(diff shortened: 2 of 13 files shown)


.run/Titanic's End Dynamic.run.xml
```diff
@@ -4,7 +4,7 @@
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="LXStudio-TE" />
     <option name="PROGRAM_PARAMETERS" value="dynamic Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
```
jkbelcher, line 7: @zranger1 Was this intentional?

src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -8,297 +8,497 @@
...
+
+        // Do the same for the per-frame uniform block and its initial data
+        gl4.glBindBuffer(GL4.GL_UNIFORM_BUFFER, uniformBlockHandles[1]);
+        gl4.glBufferData(GL4.GL_UNIFORM_BUFFER, perFrameUniformBlockSize, perFrameUniformBlock, GL4.GL_DYNAMIC_DRAW);
+        gl4.glBindBufferRange(GL4.GL_UNIFORM_BUFFER, perFrameUniformBlockBinding, uniformBlockHandles[1], 0, perFrameUniformBlockSize);
+    }
+
+    // Update once-per-frame audio data and all the calculated
+    // values derived from it.
+    private void updateAudioFrameData(double deltaMs) {
```
jkbelcher, line 333: Thanks for cleaning this out of TEAudioPattern.  It was totally redundant for every pattern to replicate the math.  In general I think GLEngine is not the home for this.  It's a general TE utility that should be run once per frame and could be consumed by other things, not just shader patterns.  That makes it a LoopTask, but it needs to run before the GLEngine so needs to be added to the loop task list first.  Having it here enforces the execution order which is probably why you did it.  I'm fine merging this for now and we can build it a proper home on a subsequent PR.

src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -8,297 +8,497 @@
...
+    private static double sinPhaseBeat = 0.0;
+    private static double volume = 0.0;
+    private static double bassLevel = 0.0;
+    private static double trebleLevel = 0.0;
+    private static double volumeRatio = 0.0;
+    private static double bassRatio = 0.0;
+    private static double trebleRatio = 0.0;
+
+    // audio and related uniform block buffer parameters
+    private FloatBuffer perRunUniformBlock;
```
jkbelcher, line 70: Very nice, I like the organization around the perRun and perFrame uniforms.

src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -262,13 +260,22 @@ public int useTexture(GL4 gl4, String textureName) {
    * textures map.
    */
   public void clearCoordinateTextures() {
+    ArrayList<Integer> keysToRemove = new ArrayList<>();
```
jkbelcher, line 263: Looks good, thanks for taking out the ConcurrentHashMap.

src/main/java/titanicsend/pattern/TEAudioPattern.java
```diff
@@ -84,30 +76,30 @@ protected void run(double deltaMs) {
    */
   protected void computeAudio(double deltaMs) {
     // Instantaneous normalized (0..1) volume level
-    volumeLevel = Math.max(levelOffset,eq.getNormalized());
+    volumeLevel = GLEngine.getVolume();
```
jkbelcher, line 79: As noted elsewhere, this audio data doesn't need to live in GLEngine.  This is fine for now.

jkbelcher (commented):
Overall looks good, nice cleanup work here.  Thanks for the additional verbose intro.  I made a few comments, once we sort those out should be ready to merge.

.run/Titanic's End Dynamic.run.xml
zranger1: No, this is a screwup I made when testing on Windows and Linux.   Only macOS requires the `-XstartOnFirstThread` flag, so I take it out for the other platforms and (usually) make sure to put it back before pushing changes.

Will be reverted to match main in next commit.

src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -8,297 +8,497 @@
...
+        // Do the same for the per-frame uniform block and its initial data
+        gl4.glBindBuffer(GL4.GL_UNIFORM_BUFFER, uniformBlockHandles[1]);
+        gl4.glBufferData(GL4.GL_UNIFORM_BUFFER, perFrameUniformBlockSize, perFrameUniformBlock, GL4.GL_DYNAMIC_DRAW);
+        gl4.glBindBufferRange(GL4.GL_UNIFORM_BUFFER, perFrameUniformBlockBinding, uniformBlockHandles[1], 0, perFrameUniformBlockSize);
+    }
+
+    // Update once-per-frame audio data and all the calculated
+    // values derived from it.
+    private void updateAudioFrameData(double deltaMs) {
+        double levelMin = 0.01;
```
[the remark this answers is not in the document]
zranger1: Done! In next commit.

src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -8,297 +8,497 @@
...
+
+            // start listening for model changes
+            lx.addListener(this);
+
+            // set running flag once initialization is complete
+            isRunning = true;
+        }
+
+        // On every frame, after initial setup
+        if (isRunning) {
```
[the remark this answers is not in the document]
zranger1: It's false until Chromatik has finished its initialization and we can successfully create an OpenGL context, and true forever afterward.   This arrangement just makes sure we don't start running the main part of the loop until Chromatik is ready.

src/main/java/titanicsend/pattern/glengine/GLEngine.java
zranger1: Agree totally  -- we should do this in a PatternAudioEngine or something like that.   You're right though, I did it this way for now to make sure the timing between audio frame processing and pattern execution was as tight as possible, and didn't add any extra latency.

src/main/java/titanicsend/pattern/glengine/GLEngine.java
zranger1: Hoping this makes it easy and fast for us to add new uniforms for better color handling and whatever else we can think of!

src/main/java/titanicsend/pattern/glengine/TextureManager.java
zranger1: :-)

src/main/java/titanicsend/pattern/TEAudioPattern.java
zranger1: Will make a note to fix this.  I'm sure we'll find lots of extra functionality to add to a separate audio processing loop task!

src/main/java/titanicsend/pattern/glengine/GLEngine.java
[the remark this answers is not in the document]
jkbelcher: Hmm.  What am I missing?  I'm reading this as (shorthand):
```
loop() { 
// Not called until engine is initialized
if (no canvas) {
  create canvas;
  isRunning = true;
}
if (isRunning) { // How would isRunning be false here?

}
```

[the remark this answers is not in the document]
zranger1: You're right.  This is an unnecessary leftover from an early initialization strategy that I just never examined closely again.  Thanks for catching it!   Fixed now in justin/3d branch.

resources/shaders/framework/template.fs
```diff
@@ -63,6 +70,16 @@ uniform sampler2D iChannel3;
 
 // returns the normalized model coordinates for the current fragment
 vec4 _getModelCoordinates() {
+    // create 3d rotation matrix rotating around the x axis by iRotationAngle
+    // TODO: Yes, this works, now we need to do it for real using LXMatrix
```
andrewlook, line 74: wow, so cool! thanks for leaving this note :)

andrewlook (commented):
thanks so much for going the extra mile and documenting all this! 🙏

PR #593, opened 2025-01-26
Improved Color Gradients Everywhere

zranger1:
### What's New
- All color gradient functions will now use swatches with up to 5 colors (Chromatik's max)
- Shaders now have the option to use color gradients generated from the active swatch.  (just `#include "include/colorspace.fs"` and call `getGradientColor(v) `)
- Both java and shader gradients will respect the color control's offset parameter
- The "dark TE" gradient is not currently supported in shaders.
- Java patterns now have the option to do color blending in the Oklab color space.  This gives a smoother gradient, with fewer brightness and color artifacts than blending in RGB or HSV.  HSV-min is still the default for now though.  You'll have to switch to Oklab from the pattern's color control UI.
- Legacy behavior is unchanged.  This should not break any patterns.

### About Oklab:  
Oklab is an Lab perceptual color space, which means that the distance between two points in the space correlates with human perception of the distance between the colors. 

Oklab is adapted from the CIELAB standard, using tables based on the CIEDE2000 color distance metric.  The original paper by Bjorn Ottosson is here: (https://bottosson.github.io/posts/oklab/)  

Having this perceptual distance metric means that we can interpolate between two colors in Lab space and humans will see the steps as approximately even.   This gives us:

- Smooth gradients with minimal banding
- Few or no gray intermediate tones between colors
- Very few visible color artifacts (unlike HSV)

In addition, the transform between Oklab and RGB spaces is relatively inexpensive compared to other Lab color spaces, so it works well for our real-time use. 
 
Note that this is not a perfect solution - human color perception is complex and depends on many variables, both biological and environmental.   But Oklab is fast, makes nice looking gradients and has very reasonable worst-case behavior.


Projects/jon test.lxp
jkbelcher, line 1: Can remove this from commit?

jkbelcher (commented):
Blue->Yellow blending looks much better!  Thanks for working on this!  Added a few comments.

Also - running the TEGradientUtils pattern chews up a ton of CPU over here, are you seeing that?

resources/shaders/include/colorspace.fs
```diff
@@ -15,3 +31,108 @@ vec3 rgb2hsv(vec3 c) {
...
+// Return the RGB color at the given index in the current palette.
+vec3 getPaletteColor(int index) {
+    return iPalette[int(mod(index, iPaletteSize))];
+}
+
+// the default gradient palette interpolator. Returns a color from the
+// current palette, or the first system color if the pattern is set to static
+// color mode.
+vec3 getGradientColor(float h) {
+    return (iPaletteOffset >= 0.0) ? getGradientColor_oklab(fract(h + iPaletteOffset)) : iColorRGB;
```
andrewlook, line 132: just curious - how does this relate to palette size? like if the palette has 3 colors, will 0-0.33 lerp between the first 2 colors, 0.34-0.66 lerp between 2nd and 3rd colors, and 0.67-1.0 lerps between the third and first colors?

Or do we not expose palettes larger than 2 colors to the shader engine yet?

Sorry for the basic question, just working on wrapping my mind around this!

src/main/java/titanicsend/color/TEColorParameter.java
```diff
@@ -260,6 +261,9 @@ private int getGradientColorFixed(float lerp, TEGradient gradient) {
       case HSVCW:
         bf = LXGradientUtils.BlendMode.HSVCW.function;
         break;
+      case OKLAB:
```
andrewlook, line 264: worth making this the default value for this param? I'd be in favor of that

andrewlook (approved):
Looking good to me! Excited to play with this 🙌

src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -377,7 +423,14 @@ public GLEngine(LX lx, int width, int height, boolean isStaticModel) {
     ySize = height;
 
     maxPoints = xSize * ySize;
-    TE.log("GLEngine: Rendering canvas size: " + xSize + "x" + ySize + " = " + GLEngine.maxPoints + " total points");
+    TE.log(
```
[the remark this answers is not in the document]
zranger1: fixed manually.  We'll see if it sticks - the linter still seems to want to indent long string continuations by 4 spaces.

src/main/java/titanicsend/pattern/glengine/GLPatternControl.java
```diff
@@ -33,6 +34,9 @@ public void setUniforms(GLShader s) {
         (float) (0xff & LXColor.blue(col)) / 255f);
     s.setUniform("iColor2HSB", LXColor.h(col) / 360f, LXColor.s(col) / 100f, LXColor.b(col) / 100f);
 
+    boolean usePalette = pattern.getControls().color.colorSource.getEnum() != TEColorParameter.ColorSource.STATIC;
+    s.setUniform("iPaletteOffset", (float) (usePalette ? pattern.getControls().color.getOffsetf() : -1f));
```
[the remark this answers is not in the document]
zranger1: fixed

Projects/jon test.lxp
zranger1: Done!

src/main/java/titanicsend/color/TEColorParameter.java
zranger1: Done in latest commit!

src/main/java/titanicsend/lx/LXGradientUtils.java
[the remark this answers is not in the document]
zranger1: Let's talk about an efficient way to do this at the next software meeting.   I'm not totally confident that I'll do the right thing here w/r/t Chromatik's architecture.

(Not seeing any significant perf difference between blending methods w/TEGradientUtils on either M2 or M3 Max.   Is that still happening?)

src/main/java/titanicsend/pattern/glengine/GLPatternControl.java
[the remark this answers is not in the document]
zranger1: Fixed!

src/main/java/titanicsend/pattern/glengine/GLEngine.java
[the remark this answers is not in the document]
andrewlook: i think my IDE settings somehow lost the auto-linter - maybe after we merge the next few PR's we can do a big bulk linter run. Sorry I dropped the ball on linter configs - I'll see if we can set up a github action or something to apply it automatically for us

andrewlook (approved):
🚀

resources/shaders/include/colorspace.fs
andrewlook: awesome, thanks for explaining!

andrewlook:
> Blue->Yellow blending looks much better! Thanks for working on this! Added a few comments.
> 
> Also - running the TEGradientUtils pattern chews up a ton of CPU over here, are you seeing that?

Sounds like calling `getGradientColor()` in java-land using OKLAB color space is going to be CPU-intensive for now - I noticed @zranger1 left a TODO on the java OKLAB calculation about looking into SIMD/JDK Vector utils/JBLAS that would help speed it up.

PR #601, opened 2025-03-23
refactor: move static helpers out to GLPreprocessorHelpers, decouple them from state

andrewlook:
I found the core logic of GLPreprocessor a bit hard to grok, so I thought it'd be helpful to take anything that's not directly responsible for modifying state, and make it a static method of a sibling helper class - the goal being that `GLPreprocessor`'s main processor method is moved up top.

One or two other small changes were involved - making the pragma parsers return ShaderConfigs rather than appending them to a list. Also the check for DRIFT opcode is moved out into `parsePragmas`, which lets `parseControl` be a static method.

### DIFF
--- a/src/main/java/titanicsend/pattern/glengine/GLPreprocessor.java
+++ b/src/main/java/titanicsend/pattern/glengine/GLPreprocessor.java
@@ -1,15 +1,16 @@
 package titanicsend.pattern.glengine;
 
+import static titanicsend.pattern.glengine.GLPreprocessorHelpers.*;
+
 import heronarts.lx.parameter.BooleanParameter;
-import heronarts.lx.parameter.BoundedParameter;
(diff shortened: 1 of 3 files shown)


zranger1 (approved):
@andrewlook, sorry for the delay.  This is great!  It'll be nice to give people the option of using VSCode for shader development.  And I really appreciate having tests for the preprocessor.   

The new code is much easier to understand than the original too.   When I wrote it, we were supporting two independent, not-quite-compatible paths for creating shaders - the "old" year 1 mechanism and the new system we have now.  The old system went away last year, and it's fantastic to have the preprocessor cleaned up as well.  

Runs fine here w/the exception that ByteBuddy is unhappy w/Java 21 on Windows, so I switched back to 20 for testing.

PR #602, opened 2025-04-08
Upgrade to new Audio Stems plugin using git submodule

jkbelcher:
Change from the original AudioStems components to the new version which is now a standalone plugin in a separate repo.

Stem labels and OSC paths can now be adjusted in the config file `audioStems.json`.  This file can be edited to match settings in the [stem analyzer application], to mesh with new flexibility implemented by @worldveil and @zranger1 .

This is the first TE repo component to be transitioned into a standalone Chromatik plugin.  Ideally we'll continue to push the architecture towards independent plugins and eventually TEApp could be an uber-plugin instead of a custom build.


jkbelcher:
There are multiple ways to do things when nesting plugins within Chromatik.  Here's a quick primer on the choices that were made so far:

**Submodule JAR location**
The are two ways to make a [submodule Chromatik plugin] available to the application:
1.  Compile the plugin into a JAR in the `Packages/` folder. This is how standalone plugins are loaded.
2. Add the plugin to the classpath flags at application load time.
I chose option 2, because it gives strict control over the loading order.  We can enforce that the submodule plugins are loaded prior to the TEApp plugin which can be critical for initialization sequence.

**Loading a plugin at all vs Direct component initialization**
You'll notice things like `lx.engine.registerComponent("audioStems", ...)` and `lx.registry.addModulator(AudioStemModulator.class)` have been removed from TEApp.  This is optional; we could keep them in TEApp instead of loading the AudioStems plugin.  For now I think it's better compartmentalization if the submodule handles its own registration.  Any submodule changes will be picked up automatically rather than needing to be mirrored into TEApp.

jkbelcher:
Closing, will redo this as a TE module instead of a separate repo.

jkbelcher:
Replaced by #609

PR #603, opened 2025-04-09
Refactor: maven parent module setup

andrewlook:
TODO list:
- [x] verify run configs / intellij instructions
- [x] move `Projects`/`Presets`/`Fixtures`/etc
- [x] verify local repo reference (via "Import Module") and demo how others can do it


jkbelcher:
Looking good so far!

jkbelcher:
Might need to split the top level `.gitignore`?  And move hidden files like `.lxpreferences` and `.devSwitch`.

andrewlook:
hmm I don't think we need to split `.gitignore` - if `.gitignore` includes `Autosave/` or `.lxpreferences`, it'll also gitignore that in subdirectories. Seems to be working fine for me.

I added a couple commits:
- in bash scripts, dynamically extract latest version from pom file (also update the jar name to use `te-app` instead of `LXStudio-TE`)
- update `LaunchTE.app` to point at the right bash scripts
- move `.app` helpers to a subdir
- remove `TEOnnxRunner` (now that we have VJLab)

jkbelcher:
Looks like the folder moves needed to be `git mv Projects te-app/` instead of `git mv Projects ./te-app`  As-is the contents were moved but the original directory still exists.

andrewlook:
closing in favor of #606 which used `git mv Projects te-app/` instead of `git mv Projects ./te-app`

PR #604, opened 2025-04-10
Add management section for User Presets under Left Pane - Content

jkbelcher:
Replaces the previous Open/Save links that were in the pattern context menu.  The originals were a quick solution on playa.  This new location should make more sense for top-level actions.

image

These classes will get a cleanup pass sometime later.


src/main/java/titanicsend/preset/UserPresetLibrary.java
```diff
@@ -22,7 +22,7 @@
 public class UserPresetLibrary implements LXSerializable {
   private final LX lx;
 
-  private static final String defaultFileName = "newFile.userPresets";
+  private static final String defaultFileName = "Presets/UserPresets/newFile.userPresets";
```
andrewlook, line 25: one little idea here: maybe if this were a static _method_, we could do something like:

```java
String.format("Presets/UserPresets/{}.userPresets", System.getProperty("user.name"));
```

So if people accidentally commit something, we can figure out which user it came from

andrewlook (dismissed):
This is looking awesome! thanks for getting the ball rolling here

src/main/java/titanicsend/preset/UserPresetLibrary.java
jkbelcher: I like it.  Changed!

jkbelcher:
@andrewlook This one is updated to the new folder structure whenever you're ready for it.

andrewlook (approved):

PR #609, opened 2025-04-13
Audio Stems upgrade and conversion to maven module

jkbelcher:
- Moved Audio Stems into a maven module.
- Ported changes from [https://github.com/jkbelcher/AudioStemsPlugin](https://github.com/jkbelcher/AudioStemsPlugin)
- Groups audio stem functions into a Chromatik plugin, laying groundwork for the module to be deployed independently as a Chromatik Package JAR.
- Stem labels and OSC paths can now be adjusted in the config file audioStems.json. This file can be edited to match settings in the [stem analyzer application], to support new flexibility implemented by @worldveil and @zranger1 .
- Modulator uses the original package name `titanicsend.audio`, to avoid breaking project files and presets.
- All existing project files and presets should load to the correct stem, even with the new dynamic stem source.
- Logs will no longer get flooded when receiving an unrecognized OSC stem path.

image


jkbelcher:
Recommend to NOT squash merge, to preserve history on files that were `git mv`'d and then modified

audio-stems/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    <artifactId>audio-stems</artifactId>
+    <version>0.1.0-SNAPSHOT</version>
+    <name>Audio Stems Plugin</name>
+    <description>A Chromatik plugin that receives audio stem information over OSC and makes it available for real-time display and mapping.</description>
+    <inceptionYear>2025</inceptionYear>
+
+    <properties>
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
andrewlook, line 22: I'll double-check, but I think you might be able to just inherit these properties from the parent pom

audio-stems/src/main/java/titanicsend/audio/AudioStemModulator.java
```diff
@@ -113,21 +74,25 @@ public String toString() {
...
 
   public AudioStemModulator(String label) {
     super(label);
 
-    addParameter("stem", this.stem);
+    addParameter(
+        "stem",
+        this.stem =
+            AudioStems.get()
+                .newSelector("Stem", "Which audio stem is the source for this modulator"));
```
andrewlook, line 90: minor code style thing - given the fixed indentation rules we're using, i might split this into two lines

```java
this.stem = AudioStems.get().newSelector("Stem", "Which audio stem is the source for this modulator"));
addParameter("stem", this.stem);
```

audio-stems/src/main/java/titanicsend/audio/AudioStems.java
```diff
@@ -0,0 +1,203 @@
...
+  private Stem[] selectorObjects = new Stem[] {null};
+  private String[] selectorOptions = new String[] {null};
+  private final List<Selector> selectors = new ArrayList<>();
+
+  public final CompoundParameter gain =
+      new CompoundParameter("Gain", 0, -1, 2).setUnits(Units.PERCENT_NORMALIZED);
+
+  public AudioStems(LX lx) {
+    super(lx, "audioStems");
+    if (current != null) {
```
andrewlook, line 47: hmm, depending on how we're expecting people to use this, I might think about adopting a more deliberate singleton pattern here if we really want to prevent multiple instances of this from running. As it's written now, this code would still call the superclass' constructor `super(lx, "audioStems")` before throwing this exception.

What I'd consider doing instead:

```java

  private static AudioStems current;

  // Using 'synchronized' ensures that if there's a race condition / multiple simultaneous callers,
  // `current == null` will be true only once.
  public static synchronized AudioStems get(LX lx) {
    if (current == null) {
      current = new AudioStems(lx);
    }
    return current;
  }

  private AudioStems(LX lx) {
    super(lx, "audioStems");
    addParameter("gain", this.gain);
    loadConfig(lx);
    this.lx.engine.osc.addListener(this);
  }  
```

audio-stems/src/main/java/titanicsend/audio/AudioStemsPlugin.java
```diff
@@ -0,0 +1,51 @@
...
+
+  private AudioStems audioStems;
+
+  public AudioStemsPlugin(LX lx) {
+    LOG.log("AudioStemsPlugin(LX) version: " + VERSION);
+  }
+
+  @Override
+  public void initialize(LX lx) {
+    lx.engine.registerComponent("audioStems", this.audioStems = new AudioStems(lx));
```
andrewlook, line 21: if we end up making AudioStems constructor private, `AudioStems.get(lx)` should work here

audio-stems/src/main/java/titanicsend/audio/AudioStemsPlugin.java
```diff
@@ -0,0 +1,51 @@
...
+
+import heronarts.lx.LX;
+import heronarts.lx.LXPlugin;
+import heronarts.lx.studio.LXStudio;
+
+@LXPlugin.Name("Audio Stems")
+public class AudioStemsPlugin implements LXStudio.Plugin {
+
+  // This string must be manually updated to match the pom.xml version
+  private static final String VERSION = "0.1.2-SNAPSHOT";
```
andrewlook, line 11: there's a streamlined way to do this - 

add this to the pom (the build time thing is optional):

```xml
<plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>build-helper-maven-plugin</artifactId>
    <version>3.3.0</version>
    <executions>
        <execution>
            <id>timestamp-property</id>
            <goals>
                <goal>timestamp-property</goal>
            </goals>
            <configuration>
                <name>build.time</name>
                <pattern>yyyy-MM-dd HH:mm</pattern>
                <locale>en_US</locale>
            </configuration>
        </execution>
    </executions>
</plugin>
<plugin>
    <groupId>org.codehaus.mojo</groupId>
    <artifactId>templating-maven-plugin</artifactId>
    <version>1.0.0</version>
    <executions>
        <execution>
            <id>generate-version-class</id>
            <goals>
                <goal>filter-sources</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

then make a class called `Version`:

```java
package com.yourpackage;

public final class Version {
    private Version() {}
    
    public static final String VERSION = "${project.version}";
    public static final String BUILD_TIME = "${build.time}";
}
```

then you can just import that version class and use it like this:

```java
logger.info("Application version: {}", Version.VERSION);
```

te-app/pom.xml
```diff
@@ -120,6 +120,12 @@
             <groupId>com.jkbstudio</groupId>
             <artifactId>supermod</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.titanicsend</groupId>
+            <artifactId>audio-stems</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
```
andrewlook, line 127: Often in multi-module maven setups I'd keep all the version numbers synchronized (so everything uses the same version number as the parent pom).

But, if we expect to be on separate release cycles with some of the modules (like maybe this plugin), we could still simplify the configuration a little bit - I'd consider adding a property to the root pom file, `{audio-stems.version}` (or something), then we can use that version number here and in `audio-stems/pom.xml`

andrewlook (commented):
This is looking great.

The main pieces of feedback I have are:

1. `audioStems.json`
2. Singleton instance management / coupling

## `audioStems.json`

~I noticed that there's one copy in the audio-stems module under `src/main/resources`. Depending on how this jar gets bundled as a plugin for others, this might end up on the classpath and potentially mask the versions of this JSON file used in the end-user's app. I also noticed that we have a version of this in `te-app/audioStems.json`, as opposed to `te-app/src/main/resources/audioStems.json`.~

~I would consider overloading `AudioStem`'s constructor (or singleton initializer) to optionally accept a custom filename (to look up in the integrator project's resource path), or maybe even a full file path (so the integrator project could define a non-standard location for it). In any case, I'd love to be able to make a copy like `teAudioStems.json`, then in `TEApp` I could do something like `AudioStemsPlugin.registerComponents(lx, "teAudioStems.json")` and I'd be 100% sure that my json file isn't accidentally getting resource-shadowed by `audio-stems/src/main/resources/audioStems.json`.~

Update: I misunderstood how this was working - I just re-read `loadConfig`, and it looks like it's only using the classpath resource loading to grab the default version from the library's `src/main/resources`, and putting the contents in the importing module's main directory.

## Singleton instance management / coupling

I made a note about singleton / constructor stuff in `AudioStems` and in `AudioStemsPlugin`. There are probably a few different ways to approach encapsulating this safely, but it might be smart to think through how integrating applications will be accessing this.

I wonder if the touchpoints in `te-app` should be doing something like `AudioStemsPlugin.get()` instead of `AudioStems.get()` - then maybe the API surface is a bit cleaner, if we could enforce that all accesses go through the plugin class?

audio-stems/src/main/java/titanicsend/audio/AudioStems.java
andrewlook: An alternative approach here could be that we could eliminate the `static AudioStems current` and `AudioStems get(LX lx)` from this class, and instead have `AudioStemsPlugin` maintain the singleton `AudioStems` instance that is accessed by code using the audio stems (e.g. `GLEngine`). I could see that being a little bit simpler, since it's a bit inconvenient to need to call `.get(lx)` (meaning that we'd need to have `lx` available anywhere we call this, just to ensure it gets constructed correctly).

audio-stems/src/main/java/titanicsend/audio/AudioStemsPlugin.java
andrewlook: actually there's another way to do this with resource filtering, that's a bit easier: https://github.com/titanicsend/LXStudio-TE/pull/610/commits/b652b36493ad4b616496af404ac2453834f3aeaa

audio-stems/pom.xml
jkbelcher: Sounds good, thanks.

audio-stems/src/main/java/titanicsend/audio/AudioStemsPlugin.java
jkbelcher: Had been wondering about a way to do this, thanks!

te-app/pom.xml
jkbelcher: Yes, potential separate release cycle.  And since we practically never update the TE App version, I want to be able to update this module version more frequently.

jkbelcher: I actually started with a property in the parent pom, and maven throws some big warnings if you try to use a property in the <version> tag.  Something along the lines of "One of these days we're going to disallow this entirely because people keep screwing it up."

audio-stems/src/main/java/titanicsend/audio/AudioStemModulator.java
jkbelcher: This kind of junk makes me grumble about the linter.  But yes I'll take your suggestion [8c04e158](https://github.com/titanicsend/LXStudio-TE/pull/609/commits/8c04e15899b6148ad25b557ffc738fd92b0264c5).

audio-stems/src/main/java/titanicsend/audio/AudioStems.java
jkbelcher: Based partially on LX convention and partially my own opinion I'm declining the changes on this one.  Explanations are on your example commit.
Not too worried about calling the super constructor before throwing an exception.  The super constructor itself will throw an exception if the same component key is attempted to be registered twice.  Either way, calling this twice will crash something (possibly the entire app) and make the module not functional.  I'd rather do the safety check once in the constructor and get it out of the way, rather than inline with every call to `get()`.
And yes, definitely wouldn't want to pass the LX instance to every `get()` call just so it could be used once for initialization.

jkbelcher:
> There are probably a few different ways to approach encapsulating this safely, but it might be smart to think through how integrating applications will be accessing this.

Fortunately I already worked through these details exactly by developing it as a standalone plugin.  It's dialed in to work both as 1) a compiled JAR imported from the ~/Chromatik/Packages folder, and 2) as a sub-plugin like we're doing here.

audio-stems/src/main/java/titanicsend/audio/AudioStemsPlugin.java
jkbelcher: Went a different route with this.

andrewlook (approved):

PR #612, opened 2025-04-20
refactor: eliminate staticModel

andrewlook:
Seems like we've got good mileage under out belts with dynamic model, and we're about ready to start clearing out stuff related to static model!


te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -273,13 +272,13 @@ private void loadDefault(String path, JsonElement defaultElement) {
       LX.error(
           "Unable to restore default value, parameter " + path + " is not LXListenableParameter");
       return;
-    } else if (parameter instanceof StringParameter || parameter instanceof FunctionalParameter) {
```
jkbelcher, line 276: Looks like this should be left in?

.run/Titanic's End.run.xml
```diff
@@ -3,7 +3,7 @@
     <option name="ALTERNATIVE_JRE_PATH" value="temurin-17.0.8" />
```
jkbelcher, line 3: We can delete the other run config if we modify this line to temurin-21.

jkbelcher (commented):
RIP the epic static model.

Just a couple minor questions, otherwise looks good to go.  Thanks for combing through this.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
andrewlook: I just removed this because intellij told me this condition is always false when reached:

Screen Shot 2025-04-21 at 8 35 28 PM

I think it's because the condition above, `!(parameter instanceof LXListenableParameter)` - if I'm understanding correctly,

- `LXListenableParameter` and `FunctionalParameter` are both abstract classes
- neither is a child of the other, both implement `LXParameter`
- a java class can only `extend` one abstract class

.. so any `FunctionalParameter` would be caught by the previous `else if` branch. Does that sound right?

Sorry for mixing this in with a larger refactor btw - in the future I'll try to make cleanups like this in their own PR without a ton of other stuff going on.

.run/Titanic's End.run.xml
```diff
@@ -1,13 +1,13 @@
...
-    <module name="te-app" />
-    <option name="PROGRAM_PARAMETERS" value="Projects/CoachellaMaster.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
-    <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/te-app" />
-    <method v="2">
-      <option name="Make" enabled="true" />
-    </method>
-  </configuration>
+    <configuration default="false" name="Titanic's End" type="Application" factoryName="Application">
+        <option name="ALTERNATIVE_JRE_PATH" value="temurin-21"/>
```
jkbelcher, line 3: This commit looks good, although fyi the linter touched every line.  Will you double check the linting here to make sure it's right?  It's leaving no space between the last quote and the />, also is using 4 space indentation.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
jkbelcher: Ok on second look yes FunctionalParameter should get caught by the above.

jkbelcher (dismissed):
Cleared for merge, just double-check the linting on the run config.

andrewlook:
> Cleared for merge, just double-check the linting on the run config.

Good call - I think my Intellij "format on save" was running, but we don't have a preferred set of XML format settings committed so it used the defaults. No space before `/>` is standard, but I think we wanna stick with 2 spaces.

For now, I just added an `.editorconfig` where we can put basic formatting settings (indent size, line length, etc) for anything not supported by spotless: https://github.com/titanicsend/LXStudio-TE/pull/612/commits/1268ebe17e3413f8a253e2d1c2777e202adcae53

That reminds me, later I want to add a few things to our spotless config:
- [pom files](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#maven-pom)
- [markdown](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#markdown)

Less important, but might as well add these too:
- [json](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#json)
- [shell](https://github.com/diffplug/spotless/blob/main/plugin-maven/README.md#shell)

.run/Titanic's End.run.xml
andrewlook: done, wrote a longer comment reply below.

jkbelcher:
> That reminds me, later I want to add a few things to our spotless config:

Let's do those file types one at a time when you get to them.

jkbelcher:
>  No space before /> is standard

This conflicts with the format IntelliJ uses to write the run config.  It seems like we should not lint generated files or at least not conflict with the application's format.

jkbelcher (approved):

PR #613, opened 2025-04-22
logging cleanup: use format strings over string concatenation

andrewlook:
I got a bit annoyed by how auto-formatting looked with some of our logging statements, e.g.

```java
TE.log(
          "*** Instantiating all "
              + this.registry.patterns.size()
              + " patterns in registry to channel "
              + channel.getLabel()
              + " ***");
```

I find it less noisy to see this instead:

```java
TE.log(
          "*** Instantiating all %d patterns in registry to channel %s ***",
          this.registry.patterns.size(), channel.getLabel());
```

It's also slower performance-wise to do java string concatenation (as it allocates more memory and causes more GC thrash). Probably negligible, but can't hurt to try and streamline things.

So I (really, Claude code lol) went through and replaced all the instances I could find with format-string style logging.

A few call-outs / future improvements
- I recognize that for things like `UserPresets` / `DevSwitch` that we may end up refactoring into externally-usable plugins, it might not make sense to use `TE.log` and it'd be better to keep `LX.log`. (It'd be great if `LX.log()` had an overloaded option to provide format-style args).
- We could also add `TE.warning` to wrap `LX.warning`.
- I noticed we don't use `lx.pushError` that much - I'd love to know when it'd make sense to use this (would that be a good way to display user-readable errors?)


te-app/src/main/java/titanicsend/util/TE.java
```diff
@@ -12,6 +12,10 @@ public static void err(String format, Object... arguments) {
     LX.error(String.format(format, arguments));
   }
 
+  public static void err(Throwable x, String format, Object... arguments) {
```
jkbelcher, line 15: I would be inclined to stick with "error" to match the LX method, but it looks like we're already using "err" below so the change would take a broader refactor.

jkbelcher (approved):
This looks good.  I appreciate the subtle performance improvement.

jkbelcher:
> I noticed we don't use lx.pushError that much - I'd love to know when it'd make sense to use this (would that be a good way to display user-readable errors?)

I'm actually not familiar with `lx.pushError`

jkbelcher:
> it might not make sense to use TE.log and it'd be better to keep LX.log. 

I've been down this road with plugins and am happy with the following solution:  Each plugin contains a file "LOG" which has pass-throughs to the LX logging methods that insert the module prefix.  This mimics the existing log prefixes from the LX, GLX, etc layers.  [Example here](https://github.com/titanicsend/LXStudio-TE/blob/3a3f3521d6ccc8e0dd087cf72ef438db3a3d1121/audio-stems/src/main/java/titanicsend/audio/LOG.java).

> (It'd be great if LX.log() had an overloaded option to provide format-style args).

This seems reasonable to request from upstream.

te-app/src/main/java/titanicsend/util/TE.java
andrewlook: yeahhh I was thinking the same thing, this already annoyed me - I'll make that change before merging

jkbelcher (approved):

PR #617, opened 2025-04-26
New midi control registration style using attributes

jkbelcher:
First pass at updating midi controllers to the new registration style.  This is untested!


andrewlook (approved):
Approved - didn't solve my problem but it's working as written and this seems to bring us closer to being in line with latest way of doing things from upstream chromatik codebase 👍

PR #620, opened 2025-04-27
Grid Fixtures: lxf, lxp, and .run configs for 3 grid dimensions

andrewlook:
Given the momentum towards digging into shader engine / resolving mixing engine issues, I thought it'd be useful to save some runConfigs + lxf's + lxp's for common configurations (also this'll make it easier for non-engineers to run Chromatik at home on a TV or something).

For the record, I'm using 1280x240 with 6 or 7 channels live on an M3 Max Pro macbook with 96gb RAM, and consistently getting 60FPS. (That's what I'm using for my concert tomorrow). So this is in a usable state already.


jkbelcher (changes requested):
Have thought about this as well.  A few changes I'd like to see for the fixtures:

- Add a common base fixture HRGrid:
  - Parameters for width & height in pixels
  - Parameters for width & height in feet
- Move the common tag to the base fixture.  Use something other than `panel` to avoid confusion with TE panels
- For the fixed-size HRGrids:
  - One child component that is the base fixture
  - Add the size to the label so it's unique after being added to Fixtures list
  - Pass fixed values for pixel width & height to the base fixture
  - Add pass-through parameters for width & height in feet.  Maybe default the size to a common use case like a TV or wall projector.

In the long run I'd like to avoid custom run configs, but for now it looks like they are a quick way to handle the --resolution program arguments.

andrewlook:
> Have thought about this as well. A few changes I'd like to see for the fixtures:
> 
> * Add a common base fixture HRGrid:
>   
>   * Parameters for width & height in pixels
>   * Parameters for width & height in feet
> * Move the common tag to the base fixture.  Use something other than `panel` to avoid confusion with TE panels
> * For the fixed-size HRGrids:
>   
>   * One child component that is the base fixture
>   * Add the size to the label so it's unique after being added to Fixtures list
>   * Pass fixed values for pixel width & height to the base fixture
>   * Add pass-through parameters for width & height in feet.  Maybe default the size to a common use case like a TV or wall projector.
> 
> In the long run I'd like to avoid custom run configs, but for now it looks like they are a quick way to handle the --resolution program arguments.

That's great feedback, thank you - I actually have gotten confused multiple times on the grid fixture and which variable means height vs. width.

Been thinking about the `--resolution` thing too - I think using an env var or java system arg would be preferable vs. CLI arguments, so there's at least less parsing / logic in the main method. Don't think that gets around the run configs issue, but I'll think about whether there's a better solution to aim for there.

andrewlook:
@jkbelcher this is my first time really writing fixtures, but I think i got this mostly working.

The only thing I don't really understand how to do is the height/width in feet - I see that there's a `scale` parameter on the fixtures, but I don't really understand how that translates to feet. Maybe we could do it in a follow-up PR, or if you know how to do it I could amend this one.

te-app/Fixtures/Grids/hr_grid/HRGrid.lxf
```diff
@@ -0,0 +1,34 @@
...
+      label: "Height",
+      description: "Height in Pixels"
+    }
+  },
+  components: [
+    {
+      type: "strip",
+      numPoints: "$width",
+      instances: "$height",
+      spacing: 1,
```
jkbelcher, line 30: Our model uses inches as the base units.  It's a total of `$widthFt*12` inches width, and has `$width-1` spaces between horizontal points. So to calculate horizontal spacing would be something like:
`spacing: "$widthFt*12/($width-1)"`

te-app/Fixtures/Grids/hr_grid/HRGrid.lxf
```diff
@@ -0,0 +1,34 @@
...
+      y: "$instance"
```
jkbelcher, line 31: `y: "$instance*$heightFt*12/($height-1)"`

te-app/Fixtures/Grids/grid_512x2048.lxf
```diff
@@ -0,0 +1,13 @@
+{
+  label: "Grid 512x2048",
```
jkbelcher, line 2: This seems like an unusual resolution... is this common?  If it's a one-off for a show you can just add the child fixture HRGrid to the model.

te-app/Fixtures/Grids/grid_1280x240.lxf
```diff
@@ -0,0 +1,13 @@
+{
+  label: "Grid 1280x240",
```
jkbelcher, line 2: Is this a common resolution for something?

jkbelcher (changes requested):
Parent->Child fixtures look good.  Left some guidance on the sizing.

For LXPs, seems like 1 or 2 is enough?  It's pretty easy to start a new project and add a grid fixture, so feels like somewhat diminishing returns to clutter up the Projects folder.

I feel like run configs is similar, either none or 1 with lots of headroom. The --resolution tag just has to be large enough to accomodate all the pixels total from all the fixtures in the model, it doesn't have to match exactly.

te-app/Fixtures/Grids/grid_512x2048.lxf
andrewlook: This is in here because I think it's useful to experiment with super-wide aspect ratios (which is a common use case for projection setups like City of Gods or our planned SF summer show).

te-app/Fixtures/Grids/grid_1280x240.lxf
andrewlook: This is in here because I think it's useful to experiment with super-wide aspect ratios (which is a common use case for projection setups like City of Gods or our planned SF summer show).

te-app/Fixtures/Grids/grid_512x2048.lxf
andrewlook: good call, this was a one-off, deleting

te-app/Fixtures/Grids/hr_grid/HRGrid.lxf
andrewlook: thanks, done

te-app/Fixtures/Grids/hr_grid/HRGrid.lxf
andrewlook: cool, thank you - where would I find the info that our model uses inches as base units? Is that explicitly represented in the LXM/LXF's for TE anywhere? Or is it arbitrary as far as chromatik is concerned, and we just chose inches as a convention?

te-app/Fixtures/Grids/grid_1024x1024.lxf
```diff
@@ -0,0 +1,13 @@
+{
+  label: "Grid 1024x1024",
+  tags: [
+    "grid"
```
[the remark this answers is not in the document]
andrewlook: It is required - I just added `width_1024`, `height_1024` as the tags because I couldn't think of anything more descriptive than `grid`. Open to ideas.

andrewlook:
Thanks for the feedback - went through and made some changes.

Killed off 2 of the resolutions but kept 3. I wanted it to be really easy to pop open a run config/project with different aspect ratios (since it's a bit of friction to mess with importing fixtures/etc).

I wanted one very wide aspect ratio (could be updated to a higher pixel count post-GPU mixer) so I kept 1280x240, and then there's also 1024x1024 and 1280x720. Feels reasonable to me. After we land GPU mixer, I'll bump up the `--resolution` to something like 5000x5000 to have enough room to accommodate all the pixels (but I don't want to do that now and slow things down)

te-app/Fixtures/Grids/grid_1024x1024.lxf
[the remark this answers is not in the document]
jkbelcher: Looks good.  I started a thread to ask why these are required.

te-app/Fixtures/Grids/hr_grid/HRGrid.lxf
jkbelcher: We chose inches as a convention, because when doing layout in person it's typical to be using an american english tape measure and it's convenient if those units can go straight into the model.

>  Is that explicitly represented in the LXM/LXF's for TE anywhere? 

That should at least have been a comment in `TE.lxf`... i see it's not there currently.

> Or is it arbitrary as far as chromatik is concerned

Yes. I'd call it a legacy missing feature that Chromatik doesn't have a Units scale on fixtures and models.  I implemented a version in my SVG Import plugin.  Expecting it to come to Chromatik at some point.

te-app/Fixtures/Grids/grid_1280x240.lxf
jkbelcher: Sounds reasonable.

jkbelcher:
> After we land GPU mixer, I'll bump up the `--resolution` to something like 5000x5000 to have enough room to accommodate all the pixels (but I don't want to do that now and slow things down)

Well it's not completely free.  That memory is still allocated in every GPU buffer, of which there are many in the gpu mixer branch.

jkbelcher:
Pushed a couple changes:

- Added pass-through parameters for the grid sizes in feet, those might as well still be adjustable from the parent fixture.
- Changed feet parameters to `float` instead of `int`

jkbelcher (dismissed):
Overall looks good.

Possible OCD follow ups:
- `HRGrid.lxf` is in a folder "`hr_grid`". Those could use the same styling to reduce visual clutter.
- The grid fixtures are lower case but every other fixture we have is capitalized.

andrewlook:
> Overall looks good.
> 
> Possible OCD follow ups:
> 
> * `HRGrid.lxf` is in a folder "`hr_grid`". Those could use the same styling to reduce visual clutter.
> * The grid fixtures are lower case but every other fixture we have is capitalized.

Good call, went ahead and did the OCD changes so they don't nag at us later: https://github.com/titanicsend/LXStudio-TE/pull/620/commits/a9ff1066d3fd8a385252e9d9a8f048582bfbd199

jkbelcher (approved):

PR #624, opened 2025-05-12
GPU Mixer (in progress)

jkbelcher:
Progress on a GPU-based mixer. Initial testing with 10 shader channels puts the render+blending capability into the millions of model points.

This starts with a heavy refactoring of the GLEngine, which was done in bite-sized commits in case someone wants to review the trail.

The core LX library mod is proof-of-concept, not final, as it's a quick and dirty edit to bypass the default CPU-based blending.

### Features + Goals

- [x] Basic channel blending (fader level)
- [x] Multiple channel blend types (Add and Multiply implemented so far)
- [x] Pattern transitions
- [x] Channel groups
- [x] Views
- [ ] All other channel blend types
- [ ] Channel composite mode (blends all patterns)
- [ ] GPU Effects (scaffolding is in place)
- [ ] Java patterns
- [ ] Java effects
- [ ] Cue+Aux buses
- [ ] A/B crossfader buses
- [ ] Improve data transfer speed from LX engine -> UI
- [ ] Update MultipassDemo, remove legacy ByteBuffer backbuffers

### Known Issues

- [ ] Getting some CPU spikes during pattern transitions (has been improved - more testing needed)


te-app/resources/shaders/blendAdd.fs
```diff
@@ -0,0 +1,28 @@
...
+// Blend destination texture: the bus from previous channels
+uniform sampler2D iDst;
+
+// Blend source texture: total output for one channel
+uniform sampler2D iSrc;
+
+// Amount of source to be applied, e.g. channel fader
+uniform float level;
+
+void main() {
```
andrewlook, line 14: tiny code organization note - we may want to make a subfolder in `resources/shaders` for these compute/system shaders

te-app/src/main/java/titanicsend/ndi/NDIPattern.java
```diff
@@ -137,54 +138,49 @@ public void createTextureForFrame() {
...
-              // pass the video frame texture to the shader
-              s.setUniform("gain", gain.getValuef());
-              s.setUniform("ndivideo", texture);
-            }
-          }
-        });
+    addShader("ndidefault.fs", this::setUniforms, buffer);
+  }
+
+  private void setUniforms(GLShader s) {
```
andrewlook, line 144: nice, this pattern of `setUniforms` is relatively common so this is a much cleaner way to structure it 👍

te-app/src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -42,8 +41,8 @@ public class GLEngine extends LXComponent implements LXLoopTask, LX.Listener {
   // not affect rendering performance.  The default size is sufficient for
   // even very large models, but can be increased if necessary.
   // TODO - make this configurable per pattern or effect.
-  private static final int mappedBufferWidth = 640;
-  private static final int mappedBufferHeight = 640;
+  @Deprecated private static final int mappedBufferWidth = 640;
+  @Deprecated private static final int mappedBufferHeight = 640;
```
andrewlook, line 45: should we update the comment above this with what needs to be done in order to fully remove these deprecated values?

te-app/resources/shaders/blendAdd.fs
jkbelcher: Absolutely.  Just need to get them loading from a subfolder.

te-app/src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -1001,6 +1007,11 @@ public static void main(String[] args) {
       try {
         TEApp lx = new TEApp(flags);
 
+        // Enable experimental GPU mixer
+        if (gpuMixer) {
+          lx.engine.renderMode = LXEngine.RenderMode.GPU;
```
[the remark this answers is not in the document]
jkbelcher: https://github.com/jkbelcher/LX/tree/te-gpu

te-app/src/main/java/titanicsend/pattern/glengine/GLEngine.java
```diff
@@ -78,35 +77,33 @@ public class GLEngine extends LXComponent implements LXLoopTask, LX.Listener {
...
   // Data and utility methods for the GL canvas/context.
   private GLAutoDrawable canvas = null;
   private GL4 gl4;
 
-  private LXModel model;
-
-  public LXModel getModel() {
-    return model;
-  }
-
```
andrewlook, line 93: should we have a `ModelListener` hook on this class? I've noticed sometimes Chromatik will crash if I have a grid fixture loaded, and then import Mothership (error message about the rendering engine not having enough space allocated for 300,000 points or something to that effect)

te-app/src/main/java/titanicsend/pattern/glengine/GLEngine.java
jkbelcher: I just wanted to ask @zranger1 if we could nix the mappedBuffer stuff entirely.

te-app/src/main/java/titanicsend/pattern/glengine/GLEngine.java
jkbelcher: It had a modelListener that was moved to `TextureManager`.  In the long run, yes, we'd like for GLEngine point size to be adjustable so we can't overrun it.  But as ZRanger noted in the comments that'll take a bunch of memory management... so for now you have to give yourself enough room with the CLI --resolution.

te-app/src/main/java/titanicsend/pattern/glengine/GLShader.java
```diff
@@ -211,127 +218,81 @@ public static ByteBuffer allocateMappedBuffer(int width, int height) {
...
+  public boolean isInitialized() {
+    return this.initialized;
   }
 
-  // Shader initialization that requires the OpenGL context
-  // Normally called each time the pattern is activated or
-  // when the effect is enabled.
-  public void init() {
+  /** Shader initialization that requires the OpenGL context. Call once. */
+  public final void init() {
```
andrewlook, line 255: related to above comment about synchronizing this check - another lightweight approach might be

```java
public final void init() {
  if (!this.initialized) {
    // ...
    this.initialized = true;
  }
}
```

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
```diff
@@ -12,133 +16,169 @@
...
+    s.setUniform("iTime", (float) getTime());
+
+    // get current primary and secondary colors
+    // TODO - we're just grabbing swatch colors here.  Do we need to worry about modulation?
+    int col = getColor1();
+    s.setUniform(
+        "iColorRGB",
+        (float) (0xff & LXColor.red(col)) / 255f,
+        (float) (0xff & LXColor.green(col)) / 255f,
+        (float) (0xff & LXColor.blue(col)) / 255f);
```
andrewlook, line 149: Idea for a future PR - I wonder if we should cache these? If we're doing this bit-bashing for every shader on every frame, maybe a little Guava LRU cache would save a few cycles

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderPattern.java
```diff
@@ -1,129 +1,236 @@
...
-    }
+  public static final int NO_TEXTURE = -1;
+
+  // list of shaders to run
+  private final List<TEShader> mutableShaders = new ArrayList<>();
+  public final List<TEShader> shaders = Collections.unmodifiableList(this.mutableShaders);
+
+  private boolean modelChanged = true;
+
+  private static class Uniforms {
```
andrewlook, line 29: Maybe `TEStandardUniforms` or `StandardUniforms` would be a slightly better name here, hinting at the possibility that pattern authors can still extend and add their own uniforms? (Also it's a bit confusing given that there's e.g. `Uniform.Float3` referencing a `Uniform` class that lives in a different place

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderPattern.java
```diff
@@ -1,129 +1,236 @@
...
 
-    public ShaderInfo(GLShader shader, GLShaderFrameSetup setup) {
-      this.shader = shader;
-      this.setup = setup;
-    }
+  public static final int NO_TEXTURE = -1;
+
+  // list of shaders to run
+  private final List<TEShader> mutableShaders = new ArrayList<>();
+  public final List<TEShader> shaders = Collections.unmodifiableList(this.mutableShaders);
```
andrewlook, line 25: is the idea of having an immutable "view" of the underlying mutable list just that we want public "read" access to the list of shaders without letting outside accessors modify the list?

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderPattern.java
```diff
@@ -1,129 +1,236 @@
...
+    this.uniforms.iColorHSB.setValue(
+        LXColor.h(col) / 360f, LXColor.s(col) / 100f, LXColor.b(col) / 100f);
+
+    col = calcColor2();
+    this.uniforms.iColor2RGB.setValue(
+        (float) (0xff & LXColor.red(col)) / 255f,
+        (float) (0xff & LXColor.green(col)) / 255f,
+        (float) (0xff & LXColor.blue(col)) / 255f);
+    this.uniforms.iColor2HSB.setValue(
+        LXColor.h(col) / 360f, LXColor.s(col) / 100f, LXColor.b(col) / 100f);
```
andrewlook, line 174: can any of this code be shared with `GLSahderEffect`?

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderPattern.java
```diff
@@ -1,129 +1,236 @@
...
-      s.shader.onActive();
+    for (TEShader shader : this.shaders) {
+      shader.onActive();
     }
   }
 
   @Override
   public void onInactive() {
     // fix exception on slow startup
+    // TODO(jkb): remove this null check?
```
andrewlook, line 221: if we do keep this null check, mini code style nit that we should put curly braces around the `return`.

te-app/src/main/java/titanicsend/pattern/glengine/TEShader.java
```diff
@@ -0,0 +1,303 @@
...
+    public int unit;
+  }
+
+  private final ArrayList<TextureInfo> textures = new ArrayList<>();
+
+  // Render buffers: ping-pong FBOs and textures
+  private PingPongFBO ppFBOs;
+
+  // the GL texture unit to which the current view model coordinate texture is bound.
+  private int modelCoordsTextureHandle = -1;
```
andrewlook, line 36: does -1 mean un-initialized? then if it's a positive integer, that's a reference to the texture channel number, or something else?

te-app/src/main/java/titanicsend/pattern/glengine/TEShader.java
```diff
@@ -0,0 +1,303 @@
...
+   *     that need to read the previous frame. If null, a buffer will be automatically allocated.
+   * @param uniformSource callback that will set uniforms on this shader
+   */
+  public TEShader(
+      LX lx, FragmentShader fragmentShader, ByteBuffer frameBuf, UniformSource uniformSource) {
+    this(lx, fragmentShader, frameBuf, List.of(uniformSource));
+  }
+
+  /**
+   * Create new shader object with default backbuffer
```
andrewlook, line 73: if I understand correctly, doesn't this have _no_ backbuffer (as opposed to "default" backbuffer)? or are those statements equivalent?

te-app/src/main/java/titanicsend/pattern/glengine/TEShader.java
```diff
@@ -0,0 +1,303 @@
...
+   *     that need to read the previous frame. If null, a buffer will be automatically allocated. *
+   * @param uniformSources list of callbacks that will set uniforms on this shader
+   * @param textureFilenames (optional) texture files to load
+   */
+  public TEShader(
+      LX lx,
+      String shaderFilename,
+      List<UniformSource> uniformSources,
+      ByteBuffer frameBuf,
+      String... textureFilenames) {
```
andrewlook, line 117: I wonder if it's worth having a config object pattern, if we find ourselves overloading a constructor this much.

```java
new TEShader(
 TEShaderConfig.new(lx)
  .withFilename(shaderFilename)
  .withUniformSources(sources)
  .withBuffer(frameBuf)
  .withTextures(filename1, filename2)
)
```

Just mentioning this b/c sometimes when there are a lot of possible argument orderings to a constructor, it gets a little confusing to read the code elsewhere and wonder what the third arg means.

te-app/src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -21,50 +22,47 @@
...
-   * string, which is then used to look up the texture unit number in our static textures map. It
-   * is, as mentioned, almost certainly reliable for our use case, but not absolutely guaranteed to
-   * generate a unique code for every possible filename.
-   */
-  private int stringToHash(String s) {
-    int hash = 0;
-    for (int i = 0; i < s.length(); i++) {
-      hash = (hash << 5) - hash + s.charAt(i);
+  public void initialize(GL4 gl4) {
+    if (this.initialized) {
```
andrewlook, line 49: similar synchronization suggestion as above

te-app/src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -74,166 +72,88 @@ private int stringToHash(String s) {
...
+    int width = this.glEngine.getWidth();
+    int height = this.glEngine.getHeight();
+    final int enginePoints = width * height;
+    final int modelPoints = model.points.length;
+    // Sanity check: make sure the system resolution has enough points to hold the model
+    if (modelPoints > enginePoints) {
+      LX.error(
+          String.format(
+              "GLEngine resolution (%d) too small for number of points in the model (%d)",
+              enginePoints, modelPoints));
```
andrewlook, line 98: I wonder if it's worth dropping a more instructive error message here, like adding `... re-run with higher --resolution WxH`

Side note: I'm trying to form a mental model of how GLEngine resolution relates to, for instance, the number of points on the fixture I'm importing. I understand when I'm importing a grid fixture, but my mental model falls apart when I have a grid fixture and then import mothership

te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
```diff
@@ -0,0 +1,475 @@
...
+
+    // Patterns have been looped
+
+    // Perform GPU mixing
+
+    // Set the target CPU buffer
+    this.glMasterBus.setMain(main);
+
+    // Recursive run the buses
+    this.glMasterBus.blend(this.blackBackground);
```
andrewlook, line 124: what's the significance of `blackBackground = -1` here?

te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
```diff
@@ -0,0 +1,475 @@
...
+    protected abstract int blendContents();
+
+    protected final int loopEffects(int dst, List<LXEffect> effects) {
+      for (LXEffect effect : effects) {
+        if (effect instanceof GLShaderEffect glShaderEffect) {
+          /*
+                    glShaderEffect.setInput(dst);
+                    glShaderEffect.run();
+                    dst = glShaderEffect.getRenderTexture();
+          */
```
andrewlook, line 236: is this supposed to be commented out?

te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
```diff
@@ -0,0 +1,475 @@
...
+
+    private final LXChannel channel;
+
+    private final BlendShader transitionShader;
+
+    public GLChannel(LXChannel channel) {
+      super(channel);
+      this.channel = channel;
+
+      this.transitionShader = new BlendShader(lx, "blendDissolve.fs");
```
andrewlook, line 401: awesome!

te-app/src/main/java/titanicsend/pattern/jon/ArcEdges.java
```diff
@@ -55,16 +55,13 @@ public ArcEdges(LX lx) {
...
-          @Override
-          public void OnFrame(GLShader s) {
-            // Here, we update line segment geometry
-            // Shader uniforms associated with a context stay resident
-            // on the GPU,so we only need to set them when something changes.
-            if (updateGeometry) {
-              sendSegments(s, saved_lines, LINE_COUNT);
-              updateGeometry = false;
-            }
+        (s) -> {
```
andrewlook, line 58: just so I understand, this lambda is a `UniformSource` - is that right?

andrewlook:
super impressive work man! Just took a first pass through, sorry if some of the questions are a little basic - hoping to find time later to dig in deeper and test it out myself locally. Also please LMK if there's anything I can do to help, like if you want me to fire up a profiler and try to pinpoint those CPU spikes you're seeing on transitions I'd be happy to do that, or anything else that'd be helpful!

te-app/resources/shaders/blendAdd.fs
```diff
@@ -0,0 +1,28 @@
+#version 410
+
+out vec4 fragColor;
+
+// Blend destination texture: the bus from previous channels
+uniform sampler2D iDst;
+
+// Blend source texture: total output for one channel
```
worldveil, line 8: q on naming: why is the previous channel the destination and the next channel the source?

te-app/src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -128,8 +131,8 @@ public class TEApp extends LXStudio {
   // model points allowed. May be changed via the startup
   // command line argument --resolution=WIDTHxHEIGHT
   // (Default allows roughly 102,000 points.)
-  public static int glRenderWidth = 320;
-  public static int glRenderHeight = 320;
+  public static int glRenderWidth = 1920;
+  public static int glRenderHeight = 1200;
```
worldveil, line 135: hell yeah

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
worldveil: `maybe a little Guava LRU cache` - how does that work?

te-app/src/main/java/heronarts/lx/studio/TEApp.java
andrewlook: aggressive defaults! 💪😎

te-app/resources/shaders/blendAdd.fs
andrewlook: my (uninformed) guess: maybe this has to do with the recursive nature of blending? I still need to step through the execution a bit to understand this better too

jkbelcher:
> super impressive work man! Just took a first pass through, sorry if some of the questions are a little basic - hoping to find time later to dig in deeper and test it out myself locally. Also please LMK if there's anything I can do to help, like if you want me to fire up a profiler and try to pinpoint those CPU spikes you're seeing on transitions I'd be happy to do that, or anything else that'd be helpful!

Great! Yeah feel free to take a look at the transition spikes with a profiler.  You can start a few channels with a short cycle time like 5 seconds and a transition time ~1 second.  Might help to compare it to a run where most of the time is *in* transition, such as using 10 second transitions and a short cycle, to see if it's the whole transition that is the problem or just the setup/teardown.

te-app/resources/shaders/blendAdd.fs
jkbelcher: Anyone reading along can jump on this task if they want.

te-app/src/main/java/titanicsend/pattern/glengine/GLShader.java
jkbelcher: Given it's single-threaded, style-wise I prefer setting `initialized` to true immediately after the check.  It groups the flag behavior all in one place which is fast to read on a scan, also removes any question about whether a partial initialization state could happen from entering the init loop but failing part way through.

And I prefer explicitly throwing an error if already initialized, rather than silently skip.  Initialization is a single-time action and if someone attempts it again that's a bug and should throw an error.  I added the `isInitialized()` method which can be checked to avoid this.

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
jkbelcher: `GLShaderEffect` has not been updated yet.  These lines were just moved over from another class that I nixed.  This class will need a full makeover pass when shader effects get implemented.

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderPattern.java
jkbelcher: Yep, likely.  As noted above `GLShaderEffect` hasn't been addressed yet.

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderPattern.java
jkbelcher: Agreed.  And I just nixed it here: 3680e0c8.  The colors array is always set before a pattern is called.

te-app/src/main/java/titanicsend/pattern/glengine/TEShader.java
jkbelcher: Legacy constructors, nixed: d9b4eea3

Every shader has a backbuffer texture now.  The option of passing in a backbuffer is legacy.  After `MultipassDemo` is modernized we can nix this entirely.  I'll add it to the Task list.

te-app/src/main/java/titanicsend/pattern/glengine/TEShader.java
jkbelcher: There was a slight oddness around the sequencing of setUniforms, because child classes can't do `.withUniformSource(this::setUniforms)` inline in the constructor, because they can't pass their method by reference before they are created.
So the child constructor calls need to stack in order, but the config setUniforms need to be *after* all those, in which case they can't get added in the top-level constructor.  So for now I'm delaying adding the config setUniforms until the init() call.  A little weird, but works.

te-app/src/main/java/titanicsend/pattern/glengine/TextureManager.java
jkbelcher: All on the main engine thread.

te-app/src/main/java/titanicsend/pattern/glengine/TextureManager.java
jkbelcher: LXEngine buffer sizes are dynamic.  Every buffer is sized to equal the sum of all points from all fixtures.  As you add fixtures (or adjust a parameter that changes the number of points), the buffers are discarded and re-allocated in the new total size.

GLEngine buffer sizes (resolution width * height) are not dynamic.  Until now we just set them at a value barely big enough to handle all the fixtures we need.  It's just work that hasn't been done yet.

jkbelcher: Updated error message: a914d821

te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
jkbelcher: -1 is an impossible texture handle value, only lives there until initialization.  Clarified: 6d25aa61

te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
jkbelcher: Yes, shader effects are not implemented yet.

te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
jkbelcher: It's using the shader `mix()` function for now.

te-app/src/main/java/heronarts/lx/studio/TEApp.java
jkbelcher: Just for testing convenience... we might not leave it this way all the time.

jkbelcher:
FYI @andrewlook, @zranger1 : CPU spikes during transitions were improved here: [c44bbc3](https://github.com/titanicsend/LXStudio-TE/pull/624/commits/c44bbc301b06840a64495e1478cd51b4d90c8380).  The pattern `onInactive()` override was doing a slow clear of CPU buffer which isn't needed.

PR #628, opened 2025-06-06
Update BGFX renderer Ids

zranger1:
Justin, simple PR in case it saves you time!  This fixes issues on Windows/Linux w/BGFX renderer detection and panel/vertex labels.

- Adds a BGFXRendererType enum based on the BGFX source
- Updates the code in TextRender3d.java to use the new enum so if the BGFX values ever change again, we'll get decent notification, and can just change the enum as necessary.


jkbelcher (approved):
Thanks!

PR #632, opened 2025-06-15
cleanup: delete OldSpiralDiamonds

andrewlook:


zranger1 (approved):
Thanks for taking care of this!    (And now, a brief moment of silence for the passing of this frequently used, but CPU-overloading Java pattern.   May it live on in shader form!)

PR #635, opened 2025-06-15
new shader: polyspiral

andrewlook:


zranger1:
**I love, love, love the clean geometry and movement of this shader!**    A couple of suggestions:

IMO, some of the coolest bits of the shader are hiding in the DJ booth!   Consider borrowing the x axis tiling code from the spiral diamonds shader, and plugging it in to mainImage() right after you normalize the coordinates:

```
     // move to car center on y axis and repeat pattern over x axis
     // at an interval. (0.5 gets you two copies, smaller is more, larger, less)
     // This should duplicate the center of the pattern on each end of the car.  We'll probably want to 
     // disable this or set interval to 1.0 for projection mapping use.
     float interval = 0.5;
     uv.x = mod(uv.x,interval) - 0.5 * interval;
     uv.y += 0.3;
```
(you'll also want to set the default YPOS control offset to 0 for this.)

The line width looks great for rectangular projections.  For use on the car, I'd set it to something higher to smooth things out.   Maybe:
`     const float iLineWidth=0.08;`
Adjust till it makes you happy though!

The easiest way I can see to get palette support working is to 
 `  #include <include/colorspace.fs>`
up near the top of the file, probably after defining PI and so forth.  (Also, be sure to remove the now-duplicated hsv2rgb() from the shader code.)  Then, down near the bottom after you calculate the value of colorT, instead of composing the color using iHue, just use:
   ` vec3 color=getGradientColor(colorT);`
This will get you a bright, smooth gradient.

image

zranger1:
Approved -- check in when you're ready!

andrewlook:
> **I love, love, love the clean geometry and movement of this shader!** A couple of suggestions:
> 
> IMO, some of the coolest bits of the shader are hiding in the DJ booth! Consider borrowing the x axis tiling code from the spiral diamonds shader, and plugging it in to mainImage() right after you normalize the coordinates:
> 
> ```
>      // move to car center on y axis and repeat pattern over x axis
>      // at an interval. (0.5 gets you two copies, smaller is more, larger, less)
>      // This should duplicate the center of the pattern on each end of the car.  We'll probably want to 
>      // disable this or set interval to 1.0 for projection mapping use.
>      float interval = 0.5;
>      uv.x = mod(uv.x,interval) - 0.5 * interval;
>      uv.y += 0.3;
> ```
> 
> (you'll also want to set the default YPOS control offset to 0 for this.)
> 
> The line width looks great for rectangular projections. For use on the car, I'd set it to something higher to smooth things out. Maybe: ` const float iLineWidth=0.08;` Adjust till it makes you happy though!
> 
> The easiest way I can see to get palette support working is to ` #include <include/colorspace.fs>` up near the top of the file, probably after defining PI and so forth. (Also, be sure to remove the now-duplicated hsv2rgb() from the shader code.) Then, down near the bottom after you calculate the value of colorT, instead of composing the color using iHue, just use: ` vec3 color=getGradientColor(colorT);` This will get you a bright, smooth gradient.
> 
> image

Wow thanks for the thoughtful feedback! This is excellent 🙏 I'm going to poke around and make some changes in a followup PR!

PR #636, opened 2025-06-16
refactor(TECommonControls): allow adding extraParams

andrewlook:
I've found that it's not possible to add additional parameters to `TEPerformancePattern` (even if I just want them available "below the fold", outside of the top 16). Mainly this comes up when I'm working on a new shader, and I have a lot of free parameters I need to experiment with (and it gets cumbersome to map each of them into wow1/wow2 iteratively).

Also, with stem-mapping/modulators, we might be able to benefit from extra parameters even if they're not visible on the midi-twisters during a performance.

I think the main problem here was that `TECommonControls` calls `setRemoteControls()` with a fixed list, right after adding all the params to the pattern. Because of that, I needed to pass a `Map<String, LXListenableNormalizedParameter>` up through the superclass, so it's a bit awkward for child patterns to do that instead of have the param as a class variable and then call `addParameter()`.

In an ideal world, maybe when child patterns called `addParameter` in their usual constructor, there could be a way to:
- call `setRemoteControls` again, with the extra parameters appended to the list
- register this param with `GLPatternControl`, so it's made available as a uniform.

Very open to feedback on the approach, just wanted to get an end-to-end prototype working.

Still TODO:
- [ ] wire up `GLPreprocessorHelpers` so that custom `iUniforms` get added to `extraParams` for auto-shaders.

### DIFF
--- a/te-app/src/main/java/titanicsend/pattern/TECommonControls.java
+++ b/te-app/src/main/java/titanicsend/pattern/TECommonControls.java
@@ -8,9 +8,7 @@ import heronarts.lx.parameter.LXListenableNormalizedParameter;
 import heronarts.lx.parameter.LXNormalizedParameter;
 import heronarts.lx.parameter.LXParameter;
 import heronarts.lx.parameter.LXParameterListener;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 import titanicsend.color.TEColorParameter;
 import titanicsend.color.TEGradientSource;
 import titanicsend.pattern.jon.TEControl;
@@ -25,6 +23,9 @@ public class TECommonControls {
 
   private final HashMap<TEControlTag, TEControl> controlList =
       new HashMap<TEControlTag, TEControl>();
+
+  // package-protexted, so it's accessible to TEPerformancePattern
+  final Map<String, LXListenableNormalizedParameter> extraParams;
 
   public final Set<LXNormalizedParameter> unusedParams = new HashSet<>();
 
@@ -46,8 +47,10 @@ public class TECommonControls {
         }
       };
 
-  TECommonControls(TEPerformancePattern pat) {
+  TECommonControls(
+      TEPerformancePattern pat, Map<String, LXListenableNormalizedParameter> extraParams) {
     this.pattern = pat;
+    this.extraParams = extraParams;
 
     panic.addListener(panicListener);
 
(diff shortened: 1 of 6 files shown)


te-app/src/main/java/titanicsend/pattern/TECommonControls.java
```diff
@@ -101,12 +104,11 @@ public void buildDefaultControlList() {
     setControl(TEControlTag.QUANTITY, p);
 
     p =
-        (CompoundParameter)
```
andrewlook, line 104: side note: unrelated to this PR, but my IDE tells me that these casts weren't needed.

te-app/src/main/java/titanicsend/pattern/TECommonControls.java
```diff
@@ -364,28 +373,34 @@ public void markUnused(LXNormalizedParameter param) {
...
-          // To be SHIFT, not implemented yet
-        });
+
+    List<LXListenableNormalizedParameter> params =
+        new ArrayList<>(
+            Arrays.asList(
+                // this.color.gradient,
+                getControl(TEControlTag.LEVELREACTIVITY).control,
+                getControl(TEControlTag.FREQREACTIVITY).control,
+                this.pattern.view,
```
andrewlook, line 383: TODO(for later): move "view" below the fold, since it's kind of a footgun for beginners.

te-app/src/main/java/titanicsend/pattern/look/TriangleCrossAudioLevels.java
```diff
@@ -2,15 +2,25 @@
...
+  //  public final CompoundParameter energy;
+
   public TriangleCrossAudioLevels(LX lx) {
-    super(lx, TEShaderView.ALL_POINTS);
+    super(
+        lx,
+        TEShaderView.ALL_POINTS,
+        Map.of(
+            "energy",
+            new CompoundParameter("Energy", .1, 0, 1)
```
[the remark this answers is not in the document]
andrewlook: A couple of ideas here ...

Is there a way to structure this such that `TEPerformancePattern` remains responsible for making the calls to `addParameter()`?

e.g. `public List<LXListenableNormalizedParameter> addCommonControls(TEPerformancePattern pat) {...}`

And then in TEPerformancePattern:

```
// in TEPerfPattern
private boolean haveWeCalledSetupCommonControls;

public void setupCommonControls() {
  List<LXListenableNormalizedParameter> controlParams = this.controls.addCommonControls();
  controlParams.map((LXListenableNormalizedParameter p) -> {
    this.addParameter(p.label, p);
  });

  List<LXListenableNormalizedParameter> extraParams = ... // list all params that have been added which don't match a TEControlTag

  this.controls.setupRemoteControls(controlParams + extraParams); // still handles the device ordering

  haveWeCalledSetupCommonControls = true;
}
```

then it could be used like:

```java
addParameter("energy", energy);
markUnused("quantity);

this.setupCommonControls();
```

maybe for the sake of avoiding a common footgun, TEPerfPattern could have a little check inside addParam / markUnused/etc:

```java
// in TEPerfPattern

public void addParameter(String label, LXListenableNormalizedParameter p) {
  if (this.haveWeCalledSetupCommonControls) {
    throw new RuntimeException("You must call addParameter() before setupCommonControls().");
  }
}
```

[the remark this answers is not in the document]
andrewlook: note to self from Justin - look at how pattern enumerates params for default remote controls

te-app/src/main/java/titanicsend/pattern/TECommonControls.java
jkbelcher: At some point in the past year or so we added [to Chromatik upstream] a bunch of overload methods on parameter classes that cast the return type.  So there will be a lot of old casting floating around that is no longer needed.  IMO take it out if you're already touching the line, or do a standalone cleanup commit to do all of them at once.

PR #638, opened 2025-06-17
Update libs to fix mesh rainbow issue

jkbelcher:
Includes latest Chromatik fix and enables backing panels with panelRow ui meshes.


andrewlook (approved):
wow, that was fast!

PR #640, opened 2025-06-21
Use panelBacking.lxf sub-fixture to reduce UI vertices

jkbelcher:
Previously we were drawing panel backings with one rectangle per panelRow.  It chewed up most of the allowed dynamic vertex buffers from BGFX.

This PR adds a common sub-fixture, `panelBacking.lxf`, which draws the backing as two triangles.

By using a sub-fixture the backing moves with any panel alignment changes and with the `onCar` position toggle.

image


jkbelcher:
Previously 1967 meshes.  Now 68 meshes:

`[GLX 2025/06/21 09:38:03] UIModelMesh count: 68`

zranger1 (approved):
Nice!   Works well in both design and performance modes, through repeated project loads and random fixture editing.    A couple of notes for the future:
- it'd be great if the "Backings" visibility switch in the TE UI worked.  I can't think of a good reason for anybody to ever turn backings off though.
- This works perfectly for our panels.    In the future, should we ask Mark for a "Generate bounding surface" feature in .lxf  that can handle more complex, non-triangle, non-convex, possibly 3d fixture shapes?  (I know that creating and tesselating this surface isn't exactly trivial, but once done, it'll help  everybody who needs backings, structural elements, or possibly diffusers on their fixtures.)

jkbelcher:
> it'd be great if the "Backings" visibility switch in the TE UI worked. I can't think of a good reason for anybody to ever turn backings off though.

Agreed.  Expecting to sweep through and fix that at some point.  Basically need it to loop over the current fixtures and toggle their `showBacking` parameter.  But if you're using the whole car fixture `TE.lxf`, I added a pass-through parameter labeled "Backings" that relays to all the panel backings.

>  a "Generate bounding surface" feature in .lxf that can handle more complex, non-triangle, non-convex, possibly 3d fixture shapes?

 Seems like the OBJ file support handles this use case?

PR #642, opened 2025-06-25
Chromatik libs 2025-06-25

jkbelcher:
Updates to the latest.

Resolves the device modulators errors that were visible at startup:
image


andrewlook (approved):
Nice, thanks for the fix!

One small piece of feedback for future reference: I recommend actually bumping the version number a little bit any time we change the dependency like this, e.g. `1.1.1-TE.2-SNAPSHOT`.

The reason for this is that I've noticed when I'm switching around branches of LX/GLX, for example `te-gpu-main` vs. `te` branch of `jkbelcher/LX`, if the version is `1.1.1-TE.1-SNAPSHOT` in both cases but there's substantially different code under the hood, it makes compile errors pop up a bit more often (since the snapshots can get cached in `.m2` even if the vendorized deps in `./repo` have been updated. So if the version string is different, it's a lot easier to track those things down.

jkbelcher:
Yeah, that occurred to me as I was submitting it.  Good reminder, I'll do that in future updates.  Do you this one recompiled with a version bump?

PR #643, opened 2025-06-26
find/replace jkbstudio. -> studio.jkb. in lxp files

andrewlook:
Follow-up from https://github.com/titanicsend/LXStudio-TE/pull/641 - just running a find/replace on the LXP files so they load properly


jkbelcher (approved):
Thanks!

PR #644, opened 2025-07-08
shader: ice glint

piemonte:
🧊 crystalline ice formations for the deep playa!

i could spend more time tweaking this but i want to create a couple more and circle back.


piemonte:
oops, didn't squash, please ignore

PR #645, opened 2025-07-08
shader: ice glint

piemonte:
🧊 crystalline ice formations for the deep playa!

i could spend more time tweaking this but i want to create a couple more and circle back.

### DIFF
--- a/te-app/resources/shaders/ice_glint.fs
+++ b/te-app/resources/shaders/ice_glint.fs
@@ -1,0 +1,212 @@
+// Ice Glint - crystalline ice formations for the deep playa
+// simulates ice crystal structures with light refraction and glinting
+//
+uniform float iFade;       // {"default": 1.0, "min": 0.0, "max": 2.0}
+uniform float iRotation;   // {"default": 1.0, "min": 0.0, "max": 2.0}
+
+float hash(vec2 p) {
+    return fract(sin(dot(p, vec2(127.1, 311.7))) * 43758.5453);
+}
+
+// 2D hash function for better randomness
+vec2 hash2(vec2 p) {
+    p = vec2(dot(p, vec2(127.1, 311.7)),
+             dot(p, vec2(269.5, 183.3)));
+    return fract(sin(p) * 43758.5453);
+}
+
+// Returns barycentric coordinates (u, v, w) where w = 1 - u - v
+vec3 triangleBarycentric(vec2 p, vec2 a, vec2 b, vec2 c) {
+    vec2 v0 = c - a;
+    vec2 v1 = b - a;
+    vec2 v2 = p - a;
+
+    float dot00 = dot(v0, v0);
+    float dot01 = dot(v0, v1);
+    float dot02 = dot(v0, v2);
+    float dot11 = dot(v1, v1);
+    float dot12 = dot(v1, v2);
+
+    float invDenom = 1.0 / (dot00 * dot11 - dot01 * dot01);
+    float u = (dot11 * dot02 - dot01 * dot12) * invDenom;
+    float v = (dot00 * dot12 - dot01 * dot02) * invDenom;
+
+    if (u >= 0.0 && v >= 0.0 && u + v <= 1.0) {
+        return vec3(u, v, 1.0 - u - v);
+    }
+    return vec3(-1.0);
+}
+
+void mainImage(out vec4 fragColor, in vec2 fragCoord) {
+    vec2 uv = fragCoord / iResolution.xy;
+
+    // calculate aspect ratio
+    float aspectRatio = iResolution.x / iResolution.y;
+
+    vec2 p = (uv - 0.5) * vec2(aspectRatio, 1.0) * 2.0 / iScale;
+
+    float color = 0.0;
+    float maxBrightness = 0.0;
+    float t = iTime * iSpeed * 2.0; // Faster default animation
+
+    float numTriangles = floor(iQuantity * 3.0 + 5.0); // 5 to 35 triangles
+    float minSpacing = max(aspectRatio, 1.0) * 0.3 / sqrt(iQuantity + 1.0);
+
+    // create multiple overlapping layers for depth
+    for (float layer = 0.0; layer < 2.0; layer++) {
+        float layerTime = t + layer * 2.0;
+
+        // generate randomly distributed triangles
+        for (float n = 0.0; n < 50.0; n++) {
+            if (n >= numTriangles) break;
+
+            // create a unique seed for each triangle
+            vec2 triSeed = vec2(n * 137.5, n * 285.7 + layer * 500.0);
+
+            // generate random position for animation
+            float timeOffset = floor(layerTime * 0.1 + hash(triSeed) * 10.0);
+            vec2 randPos = hash2(triSeed + vec2(timeOffset, 0.0));
+
+            // map to screen space
+            vec2 center = (randPos - 0.5) * vec2(aspectRatio, 1.0) * 2.5;
+
+            // add slow drift animation
+            float driftSpeed = 0.1 + hash(triSeed + vec2(3.0, 0.0)) * 0.1;
+            center += vec2(
+                sin(layerTime * driftSpeed + hash(triSeed) * 6.28) * 0.1,
+                cos(layerTime * driftSpeed * 0.7 + hash(triSeed + vec2(1.0, 0.0)) * 6.28) * 0.1
+            );
+
+            // animation
+            float animPhase = hash(triSeed) * 6.28318;
+            float animSpeed = 0.5 + hash(triSeed + vec2(1.0, 0.0)) * 1.0;
+
+            vec2 a, b, c;
+
+            // change the triangle sizes for organic look
+            float sizeVar = 0.5 + hash(triSeed + vec2(4.0, 0.0)) * 0.8;
+            float triSize = minSpacing * sizeVar * 2.5; // Large triangles
+
+            // create various triangle shapes
+            float shapeType = hash(triSeed + vec2(8.0, 0.0));
+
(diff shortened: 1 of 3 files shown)


zranger1:
This is a beautiful shader.   Obviously, you've done things like this before.   I look forward to seeing it on the car!

piemonte:
🙏 @zranger1

jkbelcher:
Just noticed that adjusting the speed parameter causes a jump in animation.  It could be nice if the animation remained smooth allowing the speed to be adjusted while the pattern was live.

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
```diff
@@ -0,0 +1,47 @@
...
+ * <p>a crystalline ice pattern for the deep playa... (features various optical properties that make
+ * it sparkle, refract, and glint)
+ *
+ * <p>controls: - SPEED: Crystallization/melting speed (0-2, default 1) - SIZE: Crystal scale and
+ * detail level (0.1-3, default 1) - QUANTITY: Crystal density and count (1-20, default 10) - WOW1:
+ * Glint intensity and refraction (0-2, default 1) - WOW2: Temperature/color temperature (0-1,
+ * default 0.5)
+ */
+@LXCategory("piemonte shaders")
+public class IceGlint extends GLShaderPattern {
```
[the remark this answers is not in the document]
piemonte: ahhhh, thanks, i wasn't sure the proper way to do this

piemonte:
@jkbelcher thanks so much. will follow up with these fixes.

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
```diff
@@ -0,0 +1,47 @@
...
+
+  /**
+   * Constructor for IceGlint pattern
+   *
+   * @param lx The LX lighting engine instance
+   */
+  public IceGlint(LX lx) {
+    super(lx, TEShaderView.DOUBLE_LARGE);
+
+    controls.setRange(TEControlTag.SPEED, 0, 2, 1);
```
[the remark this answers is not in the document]
piemonte: ah, makes sense

piemonte:
@jkbelcher pushed those fixes! thanks again, def didn't have the right config.

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
```diff
@@ -28,18 +27,18 @@ public class IceGlint extends GLShaderPattern {
...
+
+    controls.setRange(TEControlTag.QUANTITY, 17, 10, 25);
 
-    // configure temp (affects color: 0=cold blue, 1=warm white)
-    controls.setRange(TEControlTag.WOW2, 0, 1, 0.5);
+    // configure glint intensity
+    controls.setRange(TEControlTag.WOW1, 1.5, 0, 3);
 
-    // disable unused controls
-    controls.markUnused(controls.getLXControl(TEControlTag.ANGLE));
```
jkbelcher: Just curious if this was intentionally removed?

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
```diff
@@ -28,18 +27,18 @@ public class IceGlint extends GLShaderPattern {
   public IceGlint(LX lx) {
     super(lx, TEShaderView.DOUBLE_LARGE);
 
-    controls.setRange(TEControlTag.SPEED, 0, 2, 1);
-    controls.setRange(TEControlTag.SIZE, 0.1, 3, 1);
-    controls.setRange(TEControlTag.QUANTITY, 1, 30, 15);
+    controls.setRange(TEControlTag.SPEED, 0.4, 0, 1);
+    controls.setValue(TEControlTag.SPEED, 0.4);
```
jkbelcher, line 31: Minor note this line is redundant so can be removed.  The preceding line sets a default value of 0.4 with a range 0 to 1.  You only need the setValue call if you want the value to be initialized to something *other than* the default.

On the previous review this line was okay because you were setting the default to 0 and then initializing it to 0.4.

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
```diff
@@ -12,10 +12,9 @@
...
- * <p>controls: - SPEED: Crystallization/melting speed (0-2, default 1) - SIZE: Crystal scale and
- * detail level (0.1-3, default 1) - QUANTITY: Crystal density and count (1-20, default 10) - WOW1:
- * Glint intensity and refraction (0-2, default 1) - WOW2: Temperature/color temperature (0-1,
- * default 0.5)
+ * <p>controls: - SPEED: crystallization/melting speed - SIZE: crystal scale and detail level -
+ * QUANTITY: crystal density and count - WOW1: glint intensity and refraction - WOW2:
+ * temperature/color temperature
  */
 @LXCategory("piemonte shaders")
 public class IceGlint extends GLShaderPattern {
```
jkbelcher, line 20: There's one more step, you'll need to add this class to the LX registry so the pattern will show up in the content list (aka Add Pattern list).  In `TEApp.java`, find the appropriate-looking place in the `initialize()` method and add this line:
`lx.registry.addPattern(IceGlint.class);`
... and also import the class at the top of the file.

Previously it was visible in the list because `ShaderPanelsPatternConfig` child classes of type `ConstructedShaderPattern` are auto-added to the registry.

jkbelcher (changes requested):
A few more comments. Note the standalone class needs to be added to the registry to show up in the content list.

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
piemonte: ah, makes sense! i had tested the config in the "Shared..." file. and copy/pasted it back over into the IceGlint.java file once i thought it was good. 

this is something i have been hitting though in my setup as i've been testing everything out of that shared config file. 😆🙏

will get these resolved this evening. hopefully PR that other one i'm doing too.

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
piemonte: totally, will remove it 🙏

te-app/resources/shaders/ice_glint.fs
```diff
@@ -0,0 +1,212 @@
+// Ice Glint - crystalline ice formations for the deep playa
+// simulates ice crystal structures with light refraction and glinting
+//
+uniform float iFade;       // {"default": 1.0, "min": 0.0, "max": 2.0}
```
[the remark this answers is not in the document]
piemonte: 💯

te-app/resources/shaders/ice_glint.fs
```diff
@@ -0,0 +1,212 @@
...
+            float sizeVar = 0.5 + hash(triSeed + vec2(4.0, 0.0)) * 0.8;
+            float triSize = minSpacing * sizeVar * 2.5; // Large triangles
+
+            // create various triangle shapes
+            float shapeType = hash(triSeed + vec2(8.0, 0.0));
+
+            if (shapeType < 0.25) {
+                // equilateral triangles
+                float angle = hash(triSeed + vec2(2.0, 0.0)) * 6.28318;
+                angle += sin(layerTime * animSpeed * 0.2 + animPhase) * 0.1 * iRotation;
```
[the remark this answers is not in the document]
piemonte: ah sweet, i'll try!

te-app/resources/shaders/ice_glint.fs
```diff
@@ -0,0 +1,212 @@
...
+                float fadeInOut = pow(sin(fadeTime) * 0.5 + 0.5, 2.0) * maxOpacity;
+
+                // apply WOW1 control
+                // glint line becomes more prominent during fade-out
+                float fadeFactor = sin(fadeTime) * 0.5 + 0.5;
+                float electricBoost = 1.0 + (1.0 - fadeFactor) * electricLine * 0.5;
+                float finalOpacity = fadeInOut * (0.3 + glint * 0.7 * iWow1) * electricBoost;
+
+                maxBrightness = max(maxBrightness, finalOpacity);
+                color += finalOpacity * (1.0 - color * 0.5); // blend
```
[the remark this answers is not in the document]
piemonte: nice idea, will try and see how it looks 🙏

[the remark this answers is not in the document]
piemonte: (that was my intension and i think i just missed passing it, sorry my PR has been a little sloppy)

piemonte:
@jkbelcher @zranger1 pushed an update! thx again!

te-app/src/main/java/titanicsend/pattern/piemonte/IceGlint.java
zranger1: Last picky note on speed,  I promise!

We generally try to make speed symmetrical around zero unless something in pattern math precludes time running backwards.   For this pattern,  the direction doesn't make a huge amount of visual difference, but if you want to support time travel to the past, try:

```
    controls.setRange(TEControlTag.SPEED, 0.0 -1, 1);
    controls.setValue(TEControlTag.Speed, 0.4);
```

Note that in setRange() the default speed is set to zero, but setValue sets the pattern's starting speed to 0.4.   This is another convention that makes it possible for the VJ to hit the PANIC button in the UI and stop all motion.   (But the pattern will still run at speed 0.4 when it is loaded.)

piemonte: oh that's neat!

PR #646, opened 2025-07-12
fix runFile

andrewlook:
1-line change to use the default SDK


jkbelcher (approved):
Can't tell offhand what IDE action enabled this.

andrewlook:
> Can't tell offhand what IDE action enabled this.

@jkbelcher just FYI, this line in the runConfig gets added if someone changes the JDK source in the run config, here: 

image

Tends to happen if someone has a JDK installed with a slightly different name, and overrides the run config locally when developing

PR #650, opened 2025-07-16
Big GLEngine update

jkbelcher:
This is the `gpu` branch containing many shader and shader engine updates.

I think this is fully working with the normal (non-gpu) run config.


jkbelcher:
This might be worth merging without squashing to preserve the iterative history.

te-app/src/main/java/titanicsend/output/GrandShlomoStation.java
```diff
@@ -5,6 +5,8 @@
 import java.net.UnknownHostException;
 import java.util.*;
 
+// Legacy artnet output. No longer used. Can be deleted in a standalone PR.
+@Deprecated
 public class GrandShlomoStation {
```
andrewlook, line 10: RIP GrandShlomoStation! It had a good run 😢 🏆

andrewlook (approved):
🚀 🏎️  INCREDIBLE work here man! I'm so excited about this

PR #661, opened 2025-07-25
Easy hi-res NDI Out with NDIOutFixture

jkbelcher:
New `NDIOutFixture`, like a grid with width & height parameters:
- Auto-detected by the `NDIOutShaderEffect` -> an NDI stream is created with a matching resolution
- Resolution is adjustable while running
- The NDI stream is renameable (`Stream` parameter on the fixture)
- Multiple fixtures can be added, creating multiple streams
- Works with views (set a view on the channel or pattern)

This was *lightly* tested.

Demo vid!:
https://www.youtube.com/watch?v=pZ6IYRlu8Yo

### DIFF
--- a/te-app/src/main/java/titanicsend/effect/NDIOutShaderEffect.java
+++ b/te-app/src/main/java/titanicsend/effect/NDIOutShaderEffect.java
@@ -6,6 +6,7 @@ import heronarts.lx.LX;
 import heronarts.lx.LXCategory;
 import heronarts.lx.LXComponentName;
 import heronarts.lx.model.LXModel;
+import heronarts.lx.parameter.LXParameterListener;
 import heronarts.lx.structure.LXFixture;
 import heronarts.lx.structure.LXStructure;
 import heronarts.lx.studio.LXStudio;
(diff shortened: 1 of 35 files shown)


andrewlook:
just FYI @jkbelcher , after I merged the branch that this was stacked on top of, I merged main into this branch (probably would have been easier if I merged this into lower branch and then merged to main all in one go).

I was pretty careful resolving merge conflict, just wanted to give a heads up that i did that: https://github.com/titanicsend/LXStudio-TE/pull/661/commits/635d004c4caa65d54129fc2800b95314e8600bf1

I'll test more thoroughly soon. cc @sina-cb  as FYI!

te-app/src/main/java/titanicsend/effect/NDIOutShaderEffect.java
```diff
@@ -29,31 +86,127 @@ protected void run(double deltaMs, double enabledAmount) {
...
+    private void modelChanged(LXModel model) {
+      this.shader.setModelCoordinates(model);
+
+      // If our fixture size changed, set the NDI stream resolution to match
+      int width = fixture.widthPixels.getValuei();
+      int height = fixture.heightPixels.getValuei();
+      if (this.width != width || this.height != height) {
+        this.width = width;
+        this.height = height;
+        this.shader.setNdiResolution(width, height);
```
andrewlook, line 200: for later: maybe worth proactively checking the system's `--resolution` and popping up a warning if we're going to exceed it, so it's less scary in the moment if that happens?

PR #662, opened 2025-07-26
Cleaner references to child TELaserTask from UIDirector

jkbelcher:
Minor edit. `laserTask` is a child of `Director` now, so can be accessed directly instead of through global static `TELaserTask.get()`.


andrewlook (approved):
Sweet! Thanks for doing this

PR #664, opened 2025-07-26
Allow midi mapping to Laser color offset

jkbelcher:
Bug sweep: forgot to addParameter() on the laser color, which prevented it from being mappable.

### DIFF
--- a/te-app/src/main/java/titanicsend/lasercontrol/TELaserTask.java
+++ b/te-app/src/main/java/titanicsend/lasercontrol/TELaserTask.java
@@ -35,10 +35,13 @@ public class TELaserTask extends LXComponent {
       new BooleanParameter("SendTempo", DEFAULT_ENABLE_IN_PRODUCTION)
           .setDescription("Send beats and BPM to Pangolin Beyond with OSC");
 
-  public final TEColorParameter color;
-
   public final BeyondCompoundParameter brightness;
+
+  // User-editable laser color and an internal relay helper
+  public final TEColorParameter color;
   private final BeyondColorSync colorSync;
+
+  // Internal helper that is not a registered parameter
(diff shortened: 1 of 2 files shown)


te-app/src/main/java/titanicsend/lasercontrol/TELaserTask.java
```diff
@@ -35,10 +35,13 @@ public static TELaserTask get() {
       new BooleanParameter("SendTempo", DEFAULT_ENABLE_IN_PRODUCTION)
           .setDescription("Send beats and BPM to Pangolin Beyond with OSC");
 
-  public final TEColorParameter color;
-
   public final BeyondCompoundParameter brightness;
+
+  // User-editable laser color and an internal relay helper
+  public final TEColorParameter color;
```
andrewlook, line 41: so is this by default synced with the first entry in our Chromatik palette mgr system, but then we're wrapping it in a colorParam so it can be independently modified by Chromatik operator without changing the global palette? Just want to confirm I'm understanding

andrewlook (approved):

te-app/src/main/java/titanicsend/lasercontrol/TELaserTask.java
jkbelcher: This is the same color parameter we use on patterns.  This instance of it exists for laser color.  Internally it can be set to one of:
- Static (any color), 
- Normal TE gradient (a gradient of the current global color palette)
- Dark TE gradient (a gradient of 1st global color -> black -> 1st global color)

When set to one of the TE Gradients the offset parameter moves the color position within the gradient.  So yes, with Normal TE Gradient and offset=0, the color value equals the first color in the global palette.  If the Offset is adjusted, that changes the color only for lasers.

PR #667, opened 2025-07-27
Palette Manager updates

jkbelcher:
Heavy updates to color palette manager.  Functional but hold off on merge until we can integrate changes with @piemonte .


jkbelcher:
A core strategy was removing extra hue/sat/bright parameters.  They already exist in the aggregate `ColorParameter`.  By referencing only one set of params it's easier to keep everything in sync.  Now position 1 is treated as the leading `LXDynamicParameter`.  Colors at positions 2 and 3 derivatives.

Started into heavy UI changes.  I meant for this to be an expanded view, with a toggle to collapse the list and a toggle to swap the grid for the previous hue/sat sliders.  Will need to follow up later if we want to flush this out.

image

andrewlook (approved):

PR #672, opened 2025-07-28
(WIP) Starting the plumbing for a GlobalEffectsManager

andrewlook:
What I'm hoping to achieve:

- a UI widget showing the top 4-8 effects that a VJ would want to use during a show
- order of those effects should mirror the order in which they're added to the master effects bus
- utility effects (e.g. director) should be excluded
- each relevant "performance effect" can define:
  - one or two parameters which would be useful to communicate at a glance what the effect will do
  - what "trigger" means for it
- Making this shortlist of effects easily visible:
  - UI widget in performance pane
  - Corresponding MIDI driver 

Design:

Abstract class `TEPerformanceEffect` (extends `TEEffect`) to let effects define for themselves how (and whether) they should show up in our global effects manager. MasterBus effects listener (maybe project open listener? Not sure if effects bus fires on project open). The idea was that maybe for a grid showfile we wouldn't want `SimplifyEffect`, so there's some customization rather than my original idea of hardcoding our top list of effects.

Feedback Wanted:

I'm still thinking about the best "trigger"/"toggle" behavior for this. After playing with CDJ's once, I really like how their effects work - you "enable" the effect and it starts blinking, you can dial up the intensity, and then you click the effect button once for it to immediately disable (rather than needing to quickly spin a knob back to zero, with little room for error / looking terrible if it's a micron above zero).


jkbelcher:
I like the DJM style of an effect on/off button separate from the "level" for each effect.

te-app/src/main/java/titanicsend/effect/EdgeSieveEffect.java
```diff
@@ -9,7 +9,7 @@
 import titanicsend.model.TEEdgeModel;
 
 @LXCategory(LXCategory.TEXTURE)
-public class EdgeSieve extends TEEffect {
+public class EdgeSieveEffect extends TEEffect {
```
jkbelcher, line 12: housekeeping

te-app/src/main/java/titanicsend/effect/ExplodeEffect.java
```diff
@@ -39,10 +46,10 @@ public class ExplodeEffect extends GLShaderEffect {
       new BoundedParameter("Slope", 5, 1, 15).setDescription("Steepness of effect/time curve");
 
   public final BooleanParameter tempoSync =
-      new BooleanParameter("Sync", true).setDescription("Sync the effect to the engine tempo");
+      new BooleanParameter("Sync", false).setDescription("Sync the effect to the engine tempo");
```
jkbelcher, line 49: housekeeping

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
```diff
@@ -52,6 +53,21 @@ public GLShaderEffect(LX lx) {
     super(lx);
   }
 
+  @Override
+  protected LXListenableNormalizedParameter primaryParam() {
+    throw new RuntimeException("Subclasses must override");
```
jkbelcher, line 58: At this point you could make GLShaderEffect abstract and leave out these "must override" messages.  But I don't think we need to force it on people... how about patterns/effects implement the interface only if it makes sense for them?

te-app/src/main/java/titanicsend/effect/util/NDIOutShaderEffect.java
jkbelcher, line 1: Can we keep file moves in dedicated PRs to keep them clean and to separate the discussion?

jkbelcher (changes requested):
High take - I think an interface is a better fit for this than injecting into the class hierarchy.  But even then, our ability to include those is limited to classes we control.  Another option would be to build a library.  When registering an effect with the global effects library, you can submit a proxy class that can map an instance of the effect to a the generic methods / parameters for use by the GlobalEffects controller.

The proxy class might be a good place for a generic<T>.

te-app/src/main/java/titanicsend/effect/SimplifyEffect.java
jkbelcher, line 1: This is a bug fix.  It should be a standalone PR.

PR #676, opened 2025-07-29
tweak(defaults): DotVortex panic button should still have speed=0.5, not 0

andrewlook:
Small hiccup i noticed during VJ training: Clicking "Panic" button on dot vortex caused pattern to stop completely (even though when it's added to channel, the default is 0.5).

https://github.com/user-attachments/assets/7e280237-03c1-4760-9d2c-c83d86a8667e

### DIFF
--- a/te-app/src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
+++ b/te-app/src/main/java/titanicsend/pattern/yoffa/config/ShaderPanelsPatternConfig.java
@@ -284,7 +284,7 @@ public class ShaderPanelsPatternConfig {
 
     @Override
     protected void createShader() {
-      controls.setRange(TEControlTag.SPEED, 0, -4, 4).setValue(TEControlTag.SPEED, 0.5);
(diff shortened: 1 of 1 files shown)


andrewlook:
> justin had recommend once that speed range should try to stick from 0 to 1 too, might be OK for this pattern

Oh yeah, or at least -1 to 1 (I think the parameter is a bipolar param by default).

jkbelcher:
Yeah I think we had explained to @piemonte how one *could* initialize the pattern to a different value than the default.  But in practice... yes it's probably more intuitive if the panic button just goes to the initial value. Therefore better to not modify the value out of the gate.

> justin had recommend once that speed range should try to stick from 0 to 1 too

> Oh yeah, or at least -1 to 1 (I think the parameter is a bipolar param by default).

The important thing when setting range on one of our bipolar params such as Speed is to keep the range symmetrical around zero.  So (-4,4) or (-1, 1) or (-8,8) are okay.  If they're not symmetrical then zero is not at the center knob position which is harder to read in the UI or on the MFTs.

PR #678, opened 2025-07-29
Cleanup(Effects): naming consistency and creating 'util' sub-pkg

andrewlook:
No changes to functionality, only moves/renames


te-app/src/main/java/titanicsend/effect/util/NDIOutShaderEffect.java
jkbelcher, line 1: Hmm I'm not sure `effect.util` is the best place for these NDI effects.  I think of utility packages/classes as providing helper functions to multiple other classes, often through `static` methods.  These NDI effects are standalone effects that can be directly added to a device.

If you want a new home for these, how about grouping them by feature in the `titanicsend.ndi` package?

PR #679, opened 2025-07-29
GlobalEffectsManager (no UI yet), and making some effects implement TEPerformanceEffect

andrewlook:
`GlobalEffectsManager` (better name welcome, this is WIP)

## Proposed UI

- UI widget showing the top 8 effects that a VJ would want to use during a show
- order of those effects should **be hardcoded in the GlobalEffectsManager** ~mirror the order in which they're added to the master effects bus~
- Making this shortlist of effects easily visible:
  - UI widget in performance pane
  - Corresponding MIDI driver 

## Proposed Trigger/Toggle Behavior 

After playing with CDJ's once, I really like how their effects work:
- when available in the current project but **disabled**, the effect's toggle button is lit up (so it's easy to see in the dark)
- when **enabled**, the toggle button is blinking (on MIDI device and in UI)

When enabled, you can dial up the intensity, and then you click the effect button once for it to immediately disable (rather than needing to quickly spin a knob back to zero, with little room for error / looking terrible if it's a micron above zero).

## Proposed Design

Each relevant "performance effect" can **implement `TEPerformanceEffect` interface to define:
- one or two parameters which would be useful to communicate at a glance what the effect will do
- what "trigger" means for it.
  - For most effects, it might be as simple as `effect.enable()`, so `trigger()` has a default implementation in the interface and is not required by implementors (but they still have a hook to define any custom behavior)

`GlobalEffectsManager` has a hardcoded List of `Class<? extends TEPerformanceEffect>`, with a length corresponding to the number of available slots. Similar to `TEPerformancePattern`/`TECommonControls`, if an effect isn't present we'll have its slot empty, so we can maintain muscle memory for which effects live where on the MIDI controller.

### DIFF
--- a/te-app/src/main/java/heronarts/lx/studio/TEApp.java
+++ b/te-app/src/main/java/heronarts/lx/studio/TEApp.java
@@ -46,6 +46,7 @@ import titanicsend.app.dev.UIDevSwitch;
 import titanicsend.app.director.Director;
 import titanicsend.app.director.DirectorEffect;
 import titanicsend.app.director.UIDirector;
+import titanicsend.app.effectmgr.GlobalEffectManager;
 import titanicsend.audio.AudioStemModulator;
 import titanicsend.audio.AudioStems;
 import titanicsend.audio.AudioStemsPlugin;
@@ -173,6 +174,7 @@ public class TEApp extends LXStudio {
     private DevSwitch devSwitch;
     private final Director director;
     private final PresetEngine presetEngine;
+    private final GlobalEffectManager effectManager;
 
     // objects that manage UI displayed in 3D views
     private UI3DManager ui3dManager;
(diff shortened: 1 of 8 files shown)


te-app/src/main/java/titanicsend/effect/SimplifyEffect.java
```diff
@@ -7,10 +7,7 @@
 import heronarts.lx.effect.LXEffect;
 import heronarts.lx.model.LXModel;
 import heronarts.lx.model.LXPoint;
-import heronarts.lx.parameter.CompoundParameter;
-import heronarts.lx.parameter.DiscreteParameter;
-import heronarts.lx.parameter.EnumParameter;
-import heronarts.lx.parameter.LXParameter;
+import heronarts.lx.parameter.*;
```
jkbelcher, line 10: Hmm why are these imports changing around?

te-app/src/main/java/heronarts/lx/studio/TEApp.java
```diff
@@ -201,6 +203,7 @@ public Plugin(LX lx) {
       gamepadEngine = new GamepadEngine(lx);
       this.presetEngine = new PresetEngine(lx);
       this.presetEngine.openFile(lx.getMediaFile("Presets/UserPresets/BM24.userPresets"));
+      this.effectManager = new GlobalEffectManager(lx);
```
jkbelcher, line 206: we should explicitly call this.effectManager.dispose() below

andrewlook: done

PR #686, opened 2025-08-02
Shader effects in CPU mode, more informative shader error messages

jkbelcher:
- Enables shader effects in CPU mode.
- Cleans up backbuffer texture clearing for shader effects in both CPU and GPU modes
- Fixes this GL error on `SustainEffect` in CPU mode:
`UNSUPPORTED (log once): POSSIBLE ISSUE: unit 1 GLD_TEXTURE_INDEX_2D is unloadable`
- Avoids a subtle LX bug where an effect is briefly run if a project file is opened that contains the disabled effect.

New helpful warnings:
- Adds warning about uniform values that are set from java code but not found in the shader program.  *These are suppressed for now because it's a lot of them... perhaps the shader compiler is removing unused ones?
`<WARNING> Uniform 'iSpeed' not found in shader program electric.fs`
- Adds warning when an invalid texture is bound to a sampler2D uniform.  GL was already throwing a vague error:
`UNSUPPORTED (log once): POSSIBLE ISSUE: unit 1 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable`
... Now we are anticipating it and added this extra information:
`Shader 'sustain_effect.fs' uniform error: Missing texture 'iDst' for unit 3, this could be a bug...`


andrewlook (approved):
Tested locally and it's working great!

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
```diff
@@ -106,7 +185,11 @@ protected void run(double deltaMs, double enabledAmount) {
       this.shaders.get(i).setCpuBuffer(null);
     }
     // Set the CPU buffer for the last shader, if using CPU mixer
-    this.shaders.getLast().setCpuBuffer(this.lx.engine.renderMode.cpu ? this.colors : null);
+    if (this.lx.engine.renderMode.cpu) {
+      this.shaders.getLast().setCpuBuffer(this.colors);
+    } else {
+      this.shaders.getLast().setCpuBuffer(null);
```
andrewlook, line 191: just so i understand - is it important to `setCpuBuffer(null)` (as opposed to not calling this)? I'd assume it's clearing something out, just wanted to confirm.

Do we expect that this should always be `null` when we're in GPU mode?

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
```diff
@@ -132,6 +215,8 @@ public int getRenderTexture() {
 
   private void initializeUniforms(GLShader s) {
     // Keep direct references to each Uniform, saves hashmap lookup.
+
+    // TODO: confirm shader template for TE Effect matches these uniforms:
```
andrewlook, line 219: Traced through the code a bit - as far as I can tell, we have `#define TE_EFFECTSHADER` defined in a few places, but we don't have anything conditional happening based on it in `template.fs` or anywhere else.

Tracing through `explode_effect.fs` as an example, it's loaded up through our existing `GLPreprocessor` and it's having `template.fs` applied to it, just the same as all of our shader patterns. So TLDR nothing different is happening for Effects vs. regular shader patterns.

cc @zranger1 to confirm

te-app/src/main/java/titanicsend/pattern/glengine/GLShader.java
```diff
@@ -730,6 +737,11 @@ public Uniform getUniform(String name, UniformType type) {
     if (uniform == null) {
       // First time accessing this uniform, create a new object.
       int location = this.gl4.glGetUniformLocation(this.shaderProgram.id, name);
+      if (location == Uniform.LOCATION_NOT_FOUND) {
```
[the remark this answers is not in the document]
zranger1: On unused uniforms in shaders:    This mostly involves the standard uniforms, so I think the best thing to do here is just make sure this path gets handled as quickly as possible.   Obviously, not every shader is going to use every standard uniform.  Pretty sure we'd lose more time checking each shader than it takes to just set them though.   

FWIW, the GLSL compiler is super aggressive about pruning unused variables and code - all these uniforms are explicitly declared in the framework, but the compiler eats them -   so there is no cost to having unused  uniforms during shader execution.

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
zranger1: There really is almost no difference between effect and regular shaders.  

There used to be a bit more -- TE_EFFECTSHADER was originally used to tell the engine that the shader was an effect so that it would create an internal rectangular backbuffer (called mappedBuffer) that wound up with a copy of the rendered points, mapped to their normalized coords in 2D space.  

You could then use mappedBuffer to render effects that needed to use the "neighborhood" around a given pixel,  like blur, sharpen, line detection and other convolutional things.    The "Explode" effect used this feature.

For now, it is much harder to do that sort of effect.  I've tried a couple of tricky radial neighborhood things, but finding neighbors in 3D requires an exhaustive search + distance comparison for each pixel.   Even done once, at startup, it'd be slow for significant numbers of pixels.    (it's like O(n^2) to do this, I think?)

te-app/src/main/java/titanicsend/pattern/glengine/GLShader.java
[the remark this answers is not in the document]
jkbelcher: Good angle.  We should just swallow these if they are a standard uniform, but otherwise throw the informative warning message.

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
jkbelcher: Good read.  The `else {}` here is unnecessary.

te-app/src/main/java/titanicsend/pattern/glengine/GLShaderEffect.java
jkbelcher: Thanks for confirming.  It would be nice for effects to get a slightly different template, adding `iDst` or whatever we want to name the input texture.  And removing any uniforms that don't have a source parameter.

jkbelcher: @zranger1 I'm okay bringing mappedBuffer back in.  I wasn't totally sure of the purpose so I sidelined it during the remodel to get things working.  But I left a placeholder for it in the uniforms and it's easy to tack on another reserved texture unit.

PR #687, opened 2025-08-02
small refactor: encapsulate checkForPreview in GLAbstractChannel

andrewlook:
Just trying to make a small change to confirm my (still limited) understanding of the code.

I noticed that `checkForPreview()` was passed an `LXAbstractChannel` and a `GLBus` (in all cases, it was a `GLAbstractChannel`) - but the `GLAbstractChannel` had an internal reference to that same `LXAbstractChannel`, so I thought it might be easier to understand if the channel-specific behavior was encapsulated in the thing that already has a reference to the channel.

Feel free to reject this if I'm missing something though, maybe you had ideas for ways to use this that don't fit the narrow constraints above.

### DIFF
--- a/te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
+++ b/te-app/src/main/java/titanicsend/pattern/glengine/mixer/GLMixer.java
@@ -166,20 +166,6 @@ public class GLMixer implements LXMixerEngine.Listener, LXMixerEngine.PostMixer 
     this.gl4.glBindFramebuffer(GL_FRAMEBUFFER, 0);
   }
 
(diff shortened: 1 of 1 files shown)


andrewlook:
> The bigger picture is there are other buses that could be previewed, just hasn't been implemented yet in GPU mode. Those are the A/B crossfader buses, which are unique buses that do not inherit from `LXAbstractChannel`.
> 
> Given that the "list" (currently one) of bus(es) to be previewed lives on the `GLMixer` object, I think it makes sense for the method managing that list to also live on the `GLMixer` object. The logic of managing the list requires a scope broader than `GLAbstractChannel`.

Cool, that makes total sense - figured I might be missing something. Closing this PR

PR #690, opened 2025-08-03
udpated default speed for oldmatrixscroll

marianebk:


te-app/src/main/java/titanicsend/pattern/will/shaders/OldMatrixScroll.java
```diff
@@ -23,7 +23,7 @@ public OldMatrixScroll(PatternTarget target) {
 
     TECommonControls ctl = pattern.getControls();
 
-    ctl.setRange(TEControlTag.SPEED, 0, -4, 4); // speed
+    ctl.setRange(TEControlTag.SPEED, 0, 0.4, 4); // speed
```
andrewlook, line 26: @marianebk actually this function's API is:

```
ctl.setRange(TEControlTag.SPEED, <DEFAULT>, <MIN>, <MAX>)
```

So in this case, you updated the min value. (And the "speed" parameter should always be bipolar because of how our MIDI controls are set up, so the min and max should have equal magnitude and opposite sign, so that 0 is exactly in the center of the parameter's range).

But if you change the code to this, that would be perfect:

```
ctl.setRange(TEControlTag.SPEED, 0.4, -4, 4);
```

andrewlook (approved):
🎊 Awesome! thanks for this 🙏

PR #692, opened 2025-08-04
UILabels: add edges and vertices

andrewlook:
Bit of a hack to get this working - "registered" the vertices from within `TEEdgeModel` and `TEPanelModel`, using their metadata to identify the vertices then save them in a global lookup. Maybe the global lookup should live on `TEWholeModel` instead?

Just wanted to get a POC of this working in case it makes life easier for folks testing edges at the warehouse, I know Andy mentioned this would be helpful but I didn't want to add it to anyone else's plate.

image

Screen Shot 2025-08-03 at 10 46 44 PM

### DIFF
--- a/te-app/src/main/java/titanicsend/model/TEWholeModelDynamic.java
+++ b/te-app/src/main/java/titanicsend/model/TEWholeModelDynamic.java
@@ -299,6 +299,7 @@ public class TEWholeModelDynamic implements TEWholeModel, LX.Listener {
       this.mutableVertexes.add(vertex);
       this.vertexesById.put(v, vertex);
     }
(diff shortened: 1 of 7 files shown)


te-app/src/main/java/titanicsend/model/TEWholeModelDynamic.java
```diff
@@ -299,6 +299,7 @@ private TEVertex findOrCreateVertex(int v) {
       this.mutableVertexes.add(vertex);
       this.vertexesById.put(v, vertex);
     }
+    System.out.println("Vertex " + v + ": " + vertex);
```
jkbelcher, line 302: @andrewlook This debug trace can be taken out.

PR #696, opened 2025-08-04
Use custom color parameter UI in performance mode

jkbelcher:
Brings in our nice color picker in Performance mode:
image

Made possible by a new Chromatik feature as of April this year.

Includes minor edits:
- Used UIKnob.HEIGHT for height, so the parameter label ends up in the correct place.
- Changed constructor option from (w, h) to (x, y) because we only modify the position, never the size.

A couple caveats to sort out later:
- The pop-up goes off the right side of the screen in performance mode.  I have some thoughts on fixing this in upstream.
- If you edit the remote controls (not normally done on TE), all subparameters get the same UI widget.  This might need an upstream change as well.


jkbelcher:
Draft upstream change: 34be5eb

PR #700, opened 2025-08-06
UserPresets parameter refinements

jkbelcher:
This was a complex one, especially with the UI interactions.  I'll leave comments on the PR as to why things were needed.

### DIFF
--- a/te-app/src/main/java/titanicsend/pattern/TEPattern.java
+++ b/te-app/src/main/java/titanicsend/pattern/TEPattern.java
@@ -3,8 +3,6 @@ package titanicsend.pattern;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import heronarts.lx.LX;
-import heronarts.lx.LXComponent;
-import heronarts.lx.LXPresetComponent;
 import heronarts.lx.Tempo;
 import heronarts.lx.audio.GraphicMeter;
 import heronarts.lx.color.LinkedColorParameter;
@@ -67,40 +65,16 @@ public abstract class TEPattern extends DmxPattern {
     this.sua = this.modelTE.getPanel("SUA");
     this.sdc = this.modelTE.getPanel("SDC");
     this.presetSelector =
-        PresetEngine.get().getLibrary().get(this).newUserPresetParameter("Presets");
+        PresetEngine.get().getLibrary().get(this).newUserPresetSelector("Presets");
 
     addParameter(KEY_PRESET_SELECTOR, this.presetSelector);
     addParameter("setDefaults", this.captureDefaults);
 
     this.presetListener =
-        (selector) ->
-            lx.engine.addTask(
-                () -> {
-                  UserPreset preset = presetSelector.getObject();
-                  LXComponent pattern = selector.getParent();
-
-                  // If preset is null, restore defaults
-                  if (preset == null) {
-                    if (pattern instanceof TEPattern) {
-                      // For TEPattern / TEPerformancePattern, use overridable restoreDefaults()
-                      // method. We can't check for instanceof TEPerformancePattern without creating
-                      // a circular dependency.
-                      ((TEPattern) pattern).restoreDefaults();
-                    } else {
-                      // Otherwise, reset each parameter.
-                      for (LXParameter param : pattern.getParameters()) {
-                        param.reset();
-                      }
-                    }
-                  } else {
-                    // This should never happen, but check the type and log an error just in case.
-                    if (pattern instanceof LXPresetComponent) {
-                      preset.restore((LXPresetComponent) pattern);
-                    } else {
-                      TE.error("Unable to restore preset for non-LXPresetComponent");
-                    }
-                  }
-                });
+        (p) -> {
+          UserPreset preset = presetSelector.getObject();
+          PresetEngine.get().applyPresetDelayed(preset, this);
+        };
     this.presetSelector.addListener(this.presetListener);
   }
 
@@ -290,45 +264,13 @@ public abstract class TEPattern extends DmxPattern {
     }
   }
 
-  /**
-   * Called by the listener to restore from a UserPreset.
-   *
-   * <p>If called with a null preset, "restore defaults". Subclasses like TEPerformancePattern can
-   * override restoreDefaults, e.g. to replicate the behavior of panic button.
-   *
-   * @param preset
-   */
-  public void restore(UserPreset preset) {}
-
-  /**
-   * Called to restore default parameters for the "default" (null) preset. Can be overridden by
-   * subclasses (e.g. TEPerformancePattern can replicate the "panic" button functionality). NOTE: we
-   * don't need to exclude KEY_PRESET_SELECTOR here (the way we do when restoring an arbitrary
-   * preset) because the selector's default value will be selected by "reset()" on the parameter.
-   * Context [here](https://github.com/titanicsend/LXStudio-TE/pull/685#discussion_r2254597754)
-   */
-  public void restoreDefaults() {
-    for (LXParameter p : this.getParameters()) {
-      // If a custom default was captured/stored for this parameter,
-      // Set the value to the stored default.
-      if (this.defaults.containsKey(p.getPath())) {
-        p.setValue(this.defaults.get(p.getPath()));
-      } else {
-        // Otherwise, just call parameter.reset() to restore the default value
-        // given to Chromatik when param was created.
-        p.reset();
-      }
-    }
-  }
-
   @Override
   public void save(LX lx, JsonObject obj) {
     super.save(lx, obj);
     obj.add(KEY_DEFAULTS, toObject(lx, this.defaults));
+
     // Ensure non-serializable params not serialized.
-    if (obj.has(KEY_PRESET_SELECTOR)) {
-      obj.remove(KEY_PRESET_SELECTOR);
-    }
+    removePresetSelector(obj);
   }
 
   public static JsonObject toObject(LX lx, Map<String, Double> map) {
@@ -341,6 +283,9 @@ public abstract class TEPattern extends DmxPattern {
 
   @Override
   public void load(LX lx, JsonObject obj) {
+    // In case a preset selector was serialized by accident, remove it
+    removePresetSelector(obj);
+
     super.load(lx, obj);
 
     this.defaults.clear();
@@ -379,6 +324,15 @@ public abstract class TEPattern extends DmxPattern {
           "Invalid format loading default parameter value %s from JSON value: %s",
           path,
           defaultElement);
+    }
+  }
+
+  private void removePresetSelector(JsonObject obj) {
+    if (obj.has(KEY_PARAMETERS)) {
+      final JsonObject parametersObj = obj.getAsJsonObject(KEY_PARAMETERS);
+      if (parametersObj.has(KEY_PRESET_SELECTOR)) {
+        parametersObj.remove(KEY_PRESET_SELECTOR);
+      }
     }
   }
 
--- a/te-app/src/main/java/titanicsend/pattern/TEPerformancePattern.java
+++ b/te-app/src/main/java/titanicsend/pattern/TEPerformancePattern.java
@@ -80,11 +80,6 @@ public abstract class TEPerformancePattern extends TEAudioPattern {
     this.constructed = true;
   }
 
-  @Override
-  public void restoreDefaults() {
-    TEPerformancePattern.this.getControls().onPanic();
-  }
-
   public TECommonControls getControls() {
     return controls;
   }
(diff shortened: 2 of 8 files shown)


te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -67,40 +65,16 @@ protected TEPattern(LX lx) {
     this.sua = this.modelTE.getPanel("SUA");
     this.sdc = this.modelTE.getPanel("SDC");
     this.presetSelector =
-        PresetEngine.get().getLibrary().get(this).newUserPresetParameter("Presets");
+        PresetEngine.get().getLibrary().get(this).newUserPresetSelector("Presets");
```
jkbelcher, line 68: I think most references had transitioned to the Selector word, this one seemed legacy.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -67,40 +65,16 @@ protected TEPattern(LX lx) {
...
 
     addParameter(KEY_PRESET_SELECTOR, this.presetSelector);
     addParameter("setDefaults", this.captureDefaults);
 
     this.presetListener =
-        (selector) ->
-            lx.engine.addTask(
```
jkbelcher, line 77: Bumped the engine task wrapper to PresetEngine.  It is specific to UserPresets and could be needed for something else (effects).

te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -67,40 +65,16 @@ protected TEPattern(LX lx) {
...
-                  if (preset == null) {
-                    if (pattern instanceof TEPattern) {
-                      // For TEPattern / TEPerformancePattern, use overridable restoreDefaults()
-                      // method. We can't check for instanceof TEPerformancePattern without creating
-                      // a circular dependency.
-                      ((TEPattern) pattern).restoreDefaults();
-                    } else {
-                      // Otherwise, reset each parameter.
-                      for (LXParameter param : pattern.getParameters()) {
-                        param.reset();
```
jkbelcher, line 92: The subtle thing here is that the parameter defaults have already already been modified.  We can call param.reset() on the `TEPerformancePattern` parameters and they will go to the "new" defaults, not the stock defaults.

I don't think it's worth the messy overhead just to call `TEPattern.restoreDefaults()` just to call `onPanic()`.  We're really just resetting all parameters on the pattern, which works fine on `TEPatterns` too.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -290,45 +264,13 @@ protected void captureDefault(LXListenableParameter p) {
...
-   * don't need to exclude KEY_PRESET_SELECTOR here (the way we do when restoring an arbitrary
-   * preset) because the selector's default value will be selected by "reset()" on the parameter.
-   * Context [here](https://github.com/titanicsend/LXStudio-TE/pull/685#discussion_r2254597754)
-   */
-  public void restoreDefaults() {
-    for (LXParameter p : this.getParameters()) {
-      // If a custom default was captured/stored for this parameter,
-      // Set the value to the stored default.
-      if (this.defaults.containsKey(p.getPath())) {
-        p.setValue(this.defaults.get(p.getPath()));
```
jkbelcher, line 315: This is not needed.  Per above, parameter defaults have already been modified.  The call is something like:
`parameter.reset(newDefaultValue);`

te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -290,45 +264,13 @@ protected void captureDefault(LXListenableParameter p) {
...
-    }
-  }
-
   @Override
   public void save(LX lx, JsonObject obj) {
     super.save(lx, obj);
     obj.add(KEY_DEFAULTS, toObject(lx, this.defaults));
+
     // Ensure non-serializable params not serialized.
-    if (obj.has(KEY_PRESET_SELECTOR)) {
```
jkbelcher, line 329: Lost a couple hours to this bug.  The preset selector is under `Parameters`, not at the top level.  The serialization was including the selector value which rendered havoc when it restored a preset and *also* set a new selector value.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -341,6 +283,9 @@ public static JsonObject toObject(LX lx, Map<String, Double> map) {
 
   @Override
   public void load(LX lx, JsonObject obj) {
+    // In case a preset selector was serialized by accident, remove it
+    removePresetSelector(obj);
```
jkbelcher, line 287: In theory these won't sneak by for new users.  But I ended up with some saved in presets during the above bug, and you might have some saved too.  To be safe we'll sweep them from the json before it is restored.  That'll handle project file opens and preset restores.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
```diff
@@ -382,6 +327,15 @@ private void loadDefault(String path, JsonElement defaultElement) {
     }
   }
 
+  private void removePresetSelector(JsonObject obj) {
+    if (obj.has(KEY_PARAMETERS)) {
+      final JsonObject parametersObj = obj.getAsJsonObject(KEY_PARAMETERS);
+      if (parametersObj.has(KEY_PRESET_SELECTOR)) {
```
jkbelcher, line 333: Critical move here: the parameter is under the Parameters key, not at the top.

te-app/src/main/java/titanicsend/pattern/TEPerformancePattern.java
```diff
@@ -474,11 +469,17 @@ protected void run(double deltaMs) {
...
+  // Key from LXDeviceComponent
+  private static final String KEY_REMOTE_CONTROLS = "remoteControls";
+
   @Override
   public void load(LX lx, JsonObject obj) {
+    // Strip out custom remote controls to avoid stale versions overriding the new. (8-25: We've
+    // been using custom remote controls wrong, which causes problems on project files and presets)
+    if (obj.has(KEY_REMOTE_CONTROLS)) {
+      obj.remove(KEY_REMOTE_CONTROLS);
+    }
```
jkbelcher, line 481: Yeah we've been doing it wrong with remote controls.  In a later PR we'll switch from using Custom to using Normal remote controls.  For now we'll rip them out.  Restoring custom remote controls triggers a "remote controls changed" event which causes a few events: device UIs rebuild, performance device UIs rebuild, and the MFT re-sends its parameters.

We had multiple rounds of these refreshes: a user preset would load, trigger the controls changed, then the lines below would change them again.  Lots of time saved to avoid all that, especially on the MFT startup sequence.

te-app/src/main/java/titanicsend/preset/PresetEngine.java
```diff
@@ -119,4 +125,79 @@ public PresetEngine openFile(File file) {
...
+        // Normal operation: restore preset to component
+        preset.restore(presetComponent);
+      } else {
+        throw new IllegalArgumentException("Preset does not match component: " + presetComponent);
+      }
+    } else {
+      // "Default": No preset was supplied, so fully reset the component.
+
+      // Remove all modulations currently acting on the component's parameters
+      if (presetComponent instanceof LXModulatorComponent modulatorComponent) {
```
jkbelcher, line 164: Setting the "Default" took a few things:
- clearing modulations (could be mapped to one of these params from a *different* modulation engine
- Clearing modulators on the device modulation engine
- Resetting all parameters, except the label.

te-app/src/main/java/titanicsend/preset/UserPresetCollection.java
```diff
@@ -73,12 +73,6 @@ public UserPreset addPreset(LXPresetComponent component, JsonObject object) {
     return preset;
   }
 
-  public UserPreset addPreset(LXPresetComponent component, JsonObject object, String label) {
```
jkbelcher, line 76: This was unused by the time I got to it, so took it out.

te-app/src/main/java/titanicsend/preset/UserPresetCollection.java
```diff
@@ -143,6 +137,7 @@ public List<UserPreset> getPresets() {
   public class Selector extends ObjectParameter<UserPreset> {
     public Selector(String label) {
       super(label, presetObjects, presetLabels);
+      setWrappable(false);
```
jkbelcher, line 140: I find it preferable that the UIKnob when controlled by MFT doesn't wrap.

te-app/src/main/java/titanicsend/ui/UIMFTControls.java
jkbelcher, line 1: Big upgrade for the UI MFT Controls.  For one:
- The `ConcurrentModificationException` you found is due to the dang `UIKnobs` in the device and performance device UIs.  They all listen to the `Selector` parameter.  When a selector is changed, the listeners fire.  One of the listeners is the one on the pattern that restores a preset.  IF the serialization load() command restores a set of custom remote controls, the UI device controls were getting ripped down and rebuilt.  Disposing the knobs removed a parameter from the collection that was being iterated, which throws an exception. What a pain.

Upgrade #1 was here, the `UIMFTControls` now don't add or remove UI controls unless a parameters actually appears or disappears from the remote controls list.  It keeps track of the previous UI elements and moves them to their new locations (in case the list order changed).

This was great in design mode but the `UIPerformanceDevice` was still trashing its `UIKnobs` and recreating them, causing the exception.

Anyway, it's still nice in design mode to not throw away the `UIKnob`, because you can now click and drag the mouse continuously through preset restores.  Previously when the `UIKnob` was thrown away you would lose your mousedrag tracking.

te-app/src/main/java/titanicsend/preset/UIUserPresetSelector.java
```diff
@@ -0,0 +1,21 @@
...
+
+  public UIUserPresetSelector(UI ui, UserPresetCollection.Selector selector) {
+    this(ui, selector, null);
+  }
+
+  public UIUserPresetSelector(UI ui, UserPresetCollection.Selector selector, LXParameter p) {
+    super(0, 0, selector);
+
+    // The selector currently bonks if used with the command engine for unknown reasons
+    setUseCommandEngine(false);
```
jkbelcher, line 19: `UIKnob` by default uses the command engine to `setNormalized()` on the underlying parameter.  For some reason this was crashing with the Preset Selector.  Chased it for a while and did not solve it.  Seems like it it is working fine just to avoid the command engine with this one.

te-app/src/main/java/titanicsend/pattern/TEPattern.java
andrewlook: ah good catch.

side note for later: it'd be nice to have a convenience function like `PresetEngine.getCollection(this)` as a shorthand for `PresetEngine.get().getLibrary().get(this)`

te-app/src/main/java/titanicsend/preset/PresetEngine.java
```diff
@@ -119,4 +125,79 @@ public PresetEngine openFile(File file) {
...
+   * If a preset is applied from a parameter listener, and the parameter is also a device remote
+   * control (such as a preset Selector), the action needs to be delayed. Otherwise, a
+   * ConcurrentModificationException will occur when listeners of the remote controls list (such as
+   * UI device and UI performance device) perform downstream actions (such as disposing and
+   * rebuilding UI parameter controls).
+   *
+   * @param preset Preset to apply, or null for defaults
+   * @param presetComponent Target component
+   */
+  public void applyPresetDelayed(UserPreset preset, LXPresetComponent presetComponent) {
```
andrewlook, line 139: Potential syntactic sugar: making this...

```java
public static void applyPresetDelayed(UserPreset preset, LXPresetComponent presetComponent) {
    PresetEngine.get().lx.engine.addTask(
        () -> {
          applyPreset(preset, presetComponent);
        });
}
```

then callers like `TEPattern` could just do `PresetEngine.applyPresetDelayed(preset, this)` instead of `PresetEngine.get().applyPresetDelayed(preset, this)`

te-app/src/main/java/titanicsend/pattern/TEPattern.java
andrewlook: oh dammit! I'm sorry i didn't QA this all the way through, that's my bad

te-app/src/main/java/titanicsend/pattern/TEPattern.java
andrewlook: makes sense 👍

te-app/src/main/java/titanicsend/pattern/TEPattern.java
andrewlook: nice, thanks for this

te-app/src/main/java/titanicsend/pattern/TEPattern.java
andrewlook: got it, thanks for calling this out 👍

te-app/src/main/java/titanicsend/preset/PresetEngine.java
```diff
@@ -119,4 +125,79 @@ public PresetEngine openFile(File file) {
     updateLibraryName();
     return this;
   }
+
+  /**
+   * If a preset is applied from a parameter listener, and the parameter is also a device remote
+   * control (such as a preset Selector), the action needs to be delayed. Otherwise, a
```
andrewlook, line 131: how much does it actually need to be delayed? is it important that there's a delay, or just that it's called from engine thread? (and however long the engine thread takes to get to it is enough delay?)

te-app/src/main/java/titanicsend/preset/UIUserPresetSelector.java
andrewlook: what is the command engine? i've seen a couple mentions of LXCommand - is that how the UI normally dispatches actions to the core engine? what's the alternative, just directly manipulating the param values?

te-app/src/main/java/titanicsend/preset/UserPresetCollection.java
andrewlook: agreed

andrewlook (approved):
This is awesome, thanks!

PR #702, opened 2025-08-07
Shader Updates, Round 2

zranger1:
# Shader Visual Updates for 2025, Round 2

## GENERAL 
 - minor fixes & improvements in CarGeometryPatternTools.java

#### IceMelt
New - Meltwater running down glacial ice
- QUANTITY:  How much of the glacier is melting

#### TESolid
- Now compatible w/GPU mode for blazing speed!

#### OutrunGrid
- Enhanced palette support
- LvlReact - controls audio-based animated terrain height
- All other controls: Same as before.

#### FxDualWave
- Now compatible w/GPU mode
- Enhanced palette support
- Spin/Angle 
- Size: Width of waves
- Scale: Number of replicates
- Wow1: Trigger mode ("ONCE" or "RUN")
- Wow2: Palette gradient mix level (0=monochrome, 1=gradient)

#### ArcEdges
- Updated geometry for GPU Mixer
- Updated default settings to look better w/new shader engine.

#### EdgeFall
- Updated geometry for GPU Mixer
- Exploding lines now fade out over time, and there's a brief delay before reset
- pressing WowTrigger once explodes lines, pressing again resets instantly

#### FireEdges
- LVLREACT - Added whole car "bounce" on bass stem.  Try this out if you can.  I think it'll be really dramatic at full size on the car.

#### Kaleidosonic
- Scaled reactivity to sane values
- Improved color selection

### DIFF
--- a/.run/Titanic's End.run.xml
+++ b/.run/Titanic's End.run.xml
@@ -4,7 +4,7 @@
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="te-app" />
     <option name="PROGRAM_PARAMETERS" value="Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea  -Djava.awt.headless=true " />
+    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
     <option name="WORKING_DIRECTORY" value="$PROJECT_DIR$/te-app" />
     <method v="2">
       <option name="Make" enabled="true" />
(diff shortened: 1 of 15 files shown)


.run/Titanic's End.run.xml
```diff
@@ -4,7 +4,7 @@
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="te-app" />
     <option name="PROGRAM_PARAMETERS" value="Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
+    <option name="VM_PARAMETERS" value="-ea  -Djava.awt.headless=true " />
```
andrewlook, line 7: oops, I don't think we want to remove this JVM arg

zranger1: Thanks for catching that!  Next PR, I'll add "Non-Mac Titanics End GPU", etc. run configurations so I won't keep messing this up for Mac users, including myself.

PR #706, opened 2025-08-08
Off Air Parameters

jkbelcher:
A highly requested feature.

Adds a "safety wrapper" around certain parameters to prevent them from being changed while their pattern is mixed into the live output.  Prevents accidental visual jumps during a show.

Conditions that allow the parameter to be modified, because it is not live, include: channel fader is down, channel is disabled, pattern is not active and not being transitioned to on the channel, etc.

Added to the `View` and `PresetSelector` knobs on `TEPerformancePattern`.

Code notes:
- Built on our existing LXVirtualDiscreteParameter
- Off-Air parameters are removed from serialization

### DIFF
--- a/te-app/src/main/java/titanicsend/parameter/OffairDiscreteParameter.java
+++ b/te-app/src/main/java/titanicsend/parameter/OffairDiscreteParameter.java
@@ -1,0 +1,142 @@
+package titanicsend.parameter;
+
+import heronarts.lx.LXComponent;
+import heronarts.lx.effect.LXEffect;
+import heronarts.lx.mixer.LXAbstractChannel;
+import heronarts.lx.mixer.LXMixerEngine;
+import heronarts.lx.mixer.LXPatternEngine;
+import heronarts.lx.parameter.DiscreteParameter;
+import heronarts.lx.parameter.LXListenableNormalizedParameter;
+import heronarts.lx.parameter.LXParameter;
+import heronarts.lx.pattern.LXPattern;
+import titanicsend.model.justin.LXVirtualDiscreteParameter;
+
+/**
+ * Wraps a discrete parameter and prevents the value from changing while the parent device is "live"
+ */
+public class OffairDiscreteParameter<T extends DiscreteParameter>
+    extends LXVirtualDiscreteParameter<T> {
+
+  public OffairDiscreteParameter(String label, T parameter) {
+    super(label, parameter);
+  }
+
+  /**
+   * Determine is this parameter is Off-Air, aka Not contributing to Live Output. JKB note: This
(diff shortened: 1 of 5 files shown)


jkbelcher:
Wouldn't we still want it in Design mode?  Someone could be using design mode on the live car.

Is the issue being able to change them in the dev environment?  We could add a check for master live output.

zranger1:
Tying it to live output would be fine.
Though maybe it'd be better to use the Control Panel, "Developer/Production", state as a flag.  That way you could try out looks in Chromatica or on the live car without switching off channels while in developer mode, and still keep people from doing silly stuff by accident in production.

zranger1 (approved):
Code is simple + nice addition to control capabilities.  Works as advertised!

te-app/src/main/java/titanicsend/parameter/OffairDiscreteParameter.java
```diff
@@ -0,0 +1,142 @@
...
+public class OffairDiscreteParameter<T extends DiscreteParameter>
+    extends LXVirtualDiscreteParameter<T> {
+
+  public OffairDiscreteParameter(String label, T parameter) {
+    super(label, parameter);
+  }
+
+  /**
+   * Determine is this parameter is Off-Air, aka Not contributing to Live Output. JKB note: This
+   * could be sped up by tracking the device and channel.
```
andrewlook, line 26: just to understand what you mean here so I can keep it in the back of my mind, would this look like:
- at constructor time, resolve the parent device and which channel it's in
- set a listener on device/channel, updating a boolean "offAir" variable here?

Maybe those listeners would be global and keep some singleton registry up to date, so each param doesnt repeat the effort?

Not proposing we do this now, just trying to practice "thinking in LX" so I can learn. FWIW I don't think it's massively urgent to optimize this since both view and preset are things that aren't getting modified as often as say, speed or spin, and if I understand correctly we're only doing the extra work when the parameter is updated.

te-app/src/main/java/titanicsend/parameter/OffairDiscreteParameter.java
```diff
@@ -0,0 +1,142 @@
...
+    if (offAir()) {
+      super.decrement();
+    }
+    return this;
+  }
+
+  @Override
+  public LXParameter reset() {
+    if (offAir()) {
+      super.reset();
```
andrewlook, line 138: if i understand correctly, this'll also prevent panic button from updating view/preset while the fader is up?

I think that's actually desirable to prevent a big visual jump, just confirming I understand how it all fits together.

andrewlook (approved):

PR #707, opened 2025-08-08
Don't auto-enable midi surface types if multiple could be in use

jkbelcher:
This was causing problems with the new midi setup.  It would auto-enable multiple surfaces tied to the same midi I/Os, making the surface behavior wonky and dual-reading the inputs.


andrewlook (approved):
looks good, thank you - later we might want to consider removing those hardcoded MIDI device names entirely

PR #708, opened 2025-08-08
Preprocessor Fix

zranger1:
- Preprocessor now ignores #iUniform tags if they don't have an associated TEControlTag.   This enables the use of custom uniforms during development w/the VSCode plugin.  Previously a custom #iUniform would cause an exception to be thrown, and the associated shader would not load.
- Shaders now must have either `#pragma auto` or `#pragma name` to be configured as a TEAutoShader.  Formerly anything with a `#pragma` that we understood was interpreted as an autoshader, which occasionally lead to shaders having duplicate entries in the Chromatik patterns list.  This fix allows the use of our preprocessor directives in shaders created from Java. It eliminates the duplicate entries.  And down the road, with a little more work, it will allow us to change some common control settings from GLSL without restarting the Java app.

### DIFF
--- a/.run/Titanic's End.run.xml
+++ b/.run/Titanic's End.run.xml
@@ -4,7 +4,7 @@
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="te-app" />
     <option name="PROGRAM_PARAMETERS" value="Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
+    <option name="VM_PARAMETERS" value="-ea -Djava.awt.headless=true " />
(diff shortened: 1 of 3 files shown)


.run/Titanic's End.run.xml
```diff
@@ -4,7 +4,7 @@
     <option name="MAIN_CLASS_NAME" value="heronarts.lx.studio.TEApp" />
     <module name="te-app" />
     <option name="PROGRAM_PARAMETERS" value="Projects/BM2024_TE.lxp" />
-    <option name="VM_PARAMETERS" value="-ea -XstartOnFirstThread -Djava.awt.headless=true " />
```
c24t, line 7: IIRC us mac users might still want this arg, what's the reason for removing it in this PR?

c24t (approved):
Looks great, thanks for doing this so fast!

PR #716, opened 2025-08-09
tweak(DotPolka): invert size, and increase the max size a bit

andrewlook:
For the sake of consistency, inverting the size parameter so smaller size is smaller in the shader.

Also bumping up the max size a bit, because it almost becomes a second type of pattern when its zoomed in super huge :) Lots of possibilities here.

https://github.com/user-attachments/assets/bd2ca3ac-4170-405b-a22d-e506601f7a16


c24t (approved):
Sweet! LGTM

PR #719, opened 2025-08-10
Shader Updates, round 3: Cleanup

zranger1:
##A few updates + decluttering

#### **REALLY, REALLY IMPORTANT**
Reduced GPU requirement for SpaceExplosion/SpaceExplosionFX.   

While testing with multiple NDI outs (and yes, the sphere), I noticed that SpaceExplosion was using an incredible amount of GPU.  Volumetric ray marching can do this, and it was OK with just TE, but it might result in frame rate drops at higher pixel counts.   I reduced the number of sub-explosions, then increased the noise sampling frequency to make up for the loss of visual detail.   This version uses about half the GPU of the original.  

### Of Normal Importance
**Decluttering**:
Removed unused Java patterns:  FXEdgeRocket & FrameBrights

**Updated FollowThatStar pattern**.  
You can now make a **LOT** of stars.  They move in flocks.  You can cover the whole screen with them.  Worth playing with to see what it can do - it's set up in CarColorTest.LXP.   I left the tuning set up so over-the-top looks are possible.  Let me know if I need to dial it back.  It can be a lot.  

SIZE:  Size of stars
QUANTITY: Number of stars
SPIN/ANGLE:  Rotation speed of (individual stars)
WOW1: Beat reactivity (Stars pulse and sparkle)
WOW2: Monochromeness (higher == more iColorRGB, lower == more getGradientColor())


andrewlook:
@zranger1 this is looking great!

My only feedback here is that i think i'd like to setup your new star pattern as a separate shader, and keep the original `FollowThatStar`. They're good for very different things!

I love the new star shaders (and I agree that the super-high quantity is really cool, let's definitely keep that). I'd suggest we move `wow1` to `levelReact`, since it's beat reactive (we're somewhat inconsistent about levelReact/freqReact, but I've been explaining in VJ trainings that levelReact is usually bass/beat/maybe volume - so if that knob is present it indicates to new VJ's that something audio-reactive is tied to it).

My main motivation for keeping the original minkowski-distance FollowThatStar is because it looks SO good on the edges. It's one of my favorite moves to have a really slow-moving FollowThatStar with quantity=3 or 4, on an edge view channel - then when a breakdown comes in, drop everything else out and just have this glowing effect traversing the car really slowly.

The new one looks awesome on edges too, just a different-enough purpose that I'd (selfishly) like to have both :)

zranger1:
### Done:
The old `FollowThatStar` is restored

The new pattern is now called `StarSwarm`
- Beat reactivity is on LEVELREACT
- Gradient/Monochrome control is now WOW1

Sorry, I didn't realize we were using `FollowThatStar` for edges.   For all that the Minkowski stars are hard to control, they do have a unique brightness gradient.   The new one is a very different critter: 
image

te-app/src/main/java/titanicsend/pattern/jon/StarSwarm.java
```diff
@@ -0,0 +1,30 @@
...
+import titanicsend.pattern.yoffa.framework.TEShaderView;
+
+@LXCategory("Combo FG")
+public class StarSwarm extends GLShaderPattern {
+
+  public StarSwarm(LX lx) {
+    super(lx, TEShaderView.ALL_POINTS);
+
+    controls
+        .setRange(TEControlTag.QUANTITY, 5, 1, 30)
```
andrewlook, line 16: Note for later - we may want to set the max quantity around 25-27... that's when I noticed significant CPU thrash on my machine. Of course lighting-1 might be a bit different. But i love how the higher end of quantity looks, so I'm torn. I've been training VJ's to recognize the red flashing square indicator though, so I think it's manageable even if we keep it in there :)

zranger1: That's interesting.  Are you running w/NDI outputs as well?  In GPU mode on with just the main TE model, with size & quantity set to full, StarSwarm tops out around 40% here.  CPU mode is a little less.

Running the sphere demo though, I can get the GPU up to well over 90%, fans going crazy.  It probably thermal throttles itself somewhere in there too. 

I guess for the shows where we need hi-res output, at least telling VJs to dial it down a little would be good.  Hoping that the Mac Studio will do better b/c it has a lot of fast cores and importantly, better cooling.

zranger1: (The old FollowThatStar really is beautiful on the edges!)

PR #723, opened 2025-08-13
WIP:  Enable texture sampling for shader effects

zranger1:
Allows our shader effects to sample neighboring pixels, so we can do filtering, domain warping and other effects tricks!

Requires a bit of special handling for retrieving the right pixel.  I'll write up some documentation shortly, but in the meantime, we have the following working examples:
- explode_effect.fs
- distort_effect.fs


te-app/resources/docs/ShaderFramework.md
```diff
@@ -358,21 +346,14 @@ iBackbuffer contains the output of the previous shader pass or frame. Important:
...
-iMappedBuffer is an optional texture containing the data rendered by the previous shader in "normal"
-rectangular buffer form.  It is currently available only to effect shaders, and can be used for convolution
-and other effects that need rectangular "neighborhoods" of pixels.
-
-Note that only pixels that actually exist on the target LED fixture will be colored. All other pixels
-will be zero.  So blur, bloom and similar filters may not have the expected effect.
-
-It is currently not supported in pattern shaders, but will be available as an option at some point in the future.
+But be aware - texelFetch with any other coordinate likely will not get you the pixel you expect.  To get neighboring
+pixels, use the `_getMappedPixel(sampler2D tex, ivec2 coords)` function instead.
```
andrewlook, line 356: This is excellent, thank you for the super clear doc updates

te-app/resources/shaders/distort_effect.fs
```diff
@@ -0,0 +1,91 @@
...
+
+    // Apply warp to the sampling coordinates and translate back to
+    // the range [0, 1] for texture sampling
+    vec2 warpedUV = fract(abs(uv + 0.5 + warpOffset));
+
+    // Convert back to pixel coordinates for texelFetch
+    vec2 warpedPixel = vec2(warpedUV * iResolution.xy);
+
+    // Sample the warped pixel from the backbuffer
+    fragColor = _getMappedPixel(iDst,ivec2(warpedPixel));
```
andrewlook, line 90: Amazing! Thanks so much for pulling this over

te-app/src/main/java/titanicsend/effect/DistortEffect.java
```diff
@@ -0,0 +1,52 @@
...
+    addParameter("size", this.size);
+
+    addShader(
+        GLShader.config(lx).withFilename("distort_effect.fs").withUniformSource(this::setUniforms));
+  }
+
+  private void setUniforms(GLShader shader) {
+    shader.setUniform("size", size.getValuef());
+    shader.setUniform("depth", depth.getValuef());
+    shader.setUniform("speed", speed.getValuef());
```
andrewlook, line 42: I wasnt sure here, do we need speed to update iTime or will that happen internally?

If we don't have iSpeed common control, should i be updating `iTime*iSpeed` inside the shader?

te-app/src/main/java/titanicsend/pattern/glengine/TextureManager.java
```diff
@@ -75,32 +79,61 @@ public void modelGenerationChanged(LX lx, LXModel model) {
   }
 
   /**
-   * Copy a model's normalized coordinates into a special texture for use by shaders. Must be called
-   * by the parent pattern or effect at least once before the first frame is rendered and Should be
-   * called by the pattern's frametime run() function on every frame for full Chromatik view
-   * support.
+   * Writes the index of a model point to a 5x5 pixel neighborhood in the index texture buffer. This
+   * greatly improves the quality of sampling from the model buffer
```
andrewlook, line 83: I was wondering about this... if the resolution of our coords leaves gaps between the pixels, like x=10 and then the next x=20, and then a shader asks for x=11... is that the problem this is solving (make 11 resolve to the actual value at 10 tather than being a missed lookup?)

andrewlook (approved):
Amazing work as usual! Thanks soooo much for this 🙌🙏🙌🙏🙌

andrewlook:
> @andrewlook,  I copied the DistortEffect java and shader here from your PR#689 to make sure that it would work with the new system.  It does.  Looks great!
> 
> 
> 
> This will distort git attribution though, so feel free to take a look at the changes I made here, take what you like, and then remove DistortEffect from this PR so we can check it in w/689.

Thanks for calling that out, but no worries about the git attribution! In the interests of getting things integrated and tested, i'm happy to get this merged in so we can start testing stuff sooner rather than later!

Thanks for saving me the extra step of rebasing / updating distort effect!

te-app/src/main/java/titanicsend/effect/DistortEffect.java
zranger1: `time = iTime * speed` gets updated in the shader code in mainImage().

That's a good reminder though - effects use the common controls variable speed timer.  So if we added a couple of get/set methods to GLShaderEffect,  speed controls in effects could change the rate at which iTime passes the same way as normal patterns do, and we'd have clean speed control w/o visual jumps.

te-app/src/main/java/titanicsend/pattern/glengine/TextureManager.java
zranger1: This solves two problems:
- First, you would not believe how jumbled up and discontiguous our pixel layout is, especially in views.  There are lots of gaps. Blocking out the entire neighborhood really helps with this.
- And then there's  pixel pitch -- the physical (and coordinate space) difference between pixels is different on edges than on the main car body, and neither of those match the pitch of the render target bitmap.  The neighborhood thing gives us much more tolerance for a shader not asking for coordinates exactly, precisely at a pixel's location.

PR #701, opened 2025-08-07
Disable unused shader pattern reactivity controls

c24t:
Cleanup: disable controls for shader patterns that use TE common controls, but don't use the `levelReact` or `frequencyReact` uniforms in shader code.

Spot checked a few of these patterns to make sure the controls are removed in Chromatik: 
image

PR #724, opened 2025-08-13
Disable unused shader pattern wow controls

c24t:
Like #701, but for the `WOW1`, `WOW2`, and `WOWTRIGGER` controls.


te-app/resources/shaders/candy_flip.fs
```diff
@@ -5,6 +5,7 @@
 #include <include/colorspace.fs>
 #pragma TEControl.LEVELREACTIVITY.Disable
 #pragma TEControl.FREQREACTIVITY.Disable
+#pragma TEControl.WOWTRIGGER.Disable
```
zranger1, line 8: Since this shader has a Java class, the controls should be disabled there -- Java settings completely overwrite  autoshader control pragmas.  

The same applies for FaceMelt and HappyChibi.  (And actually, we really need to remove at least the `#pragma name` from all shaders with custom java classes, so they won't wind up being listed twice in the Pattern selector.)

piemonte: thanks ya'll! great catches.

andrewlook (approved):

PR #727, opened 2025-08-14
Add DatsaMoire shader pattern

c24t:
`DatsaMoire` is a Moiré(-ish) interference pattern built from four rotating spirals. The position parameters control the spiral center coordinates, and the wow parameters control which set of spirals are displayed and how they overlap.

https://github.com/user-attachments/assets/34065656-b659-4256-9402-ef48411ff441


andrewlook (approved):
Hell yeah! This is awesome. Merging so I can try this out locally!

PR #735, opened 2025-08-15
(WIP - please don't merge yet) Presets: rename 'open' to 'load', add option to 'merge'

andrewlook:
- "Load": open new library and replace existing library entirely (renamed from "Open")
- "Merge": open library and add all the presets to what's currently loaded

So if we have user-specific preset files we want to integrate, you can load BM24.userPresets, merge Look.userPresets, save BM24.userPresets

image


jkbelcher (approved):
Looks good.  FYI @andrewlook I made a small change: the "current file" now stays the same after an Import, rather than changing to the imported file name.

PR #738, opened 2025-08-16
Smooth Happy Chibi speed changes

c24t:
Drive-by fix for Happy Chibi: prevent speed changes from making the animation jump.

Use just `iTime` instead of `iSpeed` and update control uniforms to match. This should have a similar speed range to the original.


zranger1 (approved):
Looks good!

PR #742, opened 2025-08-18
Autosave presets

worldveil:
This builds on https://github.com/titanicsend/LXStudio-TE/pull/735, so that should be checked and merged first!

Functionality in user presets UI when making new user presets, ie: different starting parameters or stem modulator mappings with VJLab:

1. When user clicks "Add", this new user preset is saved to the collection, just as though "Save" had been clicked in `UIUserPresetsManager` under `Content`. 
2. When user clicks, "Remove" we remove from that same collection file
3. When user renames a preset, the old one is deleted and a new one created with that label, and then saved to collection automatically
4. When a user clicks update, that preset (active one) is similarly saved to collection

This extra complexity was to prevent footguns. The previous logic essentially just saved the entire collection state which meant that saving a single preset would crystalize and save ALL other presets in the open Chromatik project, which isn't ideal if other presets are in various states of development (or you were just tweaking them as a part of playing around). 

This logic intends to make it easy to directly save a single user preset to the collection with "Update" button and eliminate the need to use the global "Save" (which is a blunt tool, anyway).

### DIFF
--- a/te-app/src/main/java/titanicsend/dmx/effect/DmxEffect.java
+++ b/te-app/src/main/java/titanicsend/dmx/effect/DmxEffect.java
@@ -32,7 +32,7 @@ public abstract class DmxEffect extends LXEffect {
    */
   @Override
   public LXLayeredComponent setBuffer(LXBuffer buffer) {
-    this.dmx = DmxEngine.get().getDmxModelBuffer(buffer, this.getChannel());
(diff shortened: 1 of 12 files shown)


jkbelcher:
FYI the file save operation are synchronous and therefore risk a visual lag if performed during a show.  e.g. If you autosave after every "update" or "add" there is a risk of frame drop each time.

PR #743, opened 2025-08-18
Improve Inactive Effect Behavior

zranger1:
Minimizes work done by Explode and Distort effects when inactive.


jkbelcher (approved):
Thanks @zranger1!

PR #745, opened 2025-08-19
Add AutoSave option for user presets

jkbelcher:
Autosave option for User Presets:

image

When enabled, autosaves occur after:
- Add preset
- Update preset
- Remove preset (or backspace delete)
- Rename preset
- Import userPresets library


jkbelcher:
This is an alternative to PR #742

PR #746, opened 2025-08-19
tweak(RhythmicStatic): use wow2 for color mix

andrewlook:
I thought it'd be nice to expose option for gradient lerp vs. single color, so we can stem-map the color mix and pop from monochrome to gradient-lerp.


zranger1 (approved):
Good idea!  Approved.

PR #750, opened 2025-08-20
Look/minilab3 part7

andrewlook:
fix normalized param ranges, and support unplugging and replugging the controller and having it return to the correct pad colors for the current bank (by responding to bank received).

Tried to proactively set bank, but couldnt easily find that in bitwig classes

### DIFF
--- a/te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
+++ b/te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
@@ -262,7 +262,11 @@ public class EffectsMiniLab3 extends LXMidiSurface implements LXMidiSurface.Bidi
   protected void onEnable(boolean on) {
     if (on) {
       register();
-      initialize();
+      try {
+        initialize();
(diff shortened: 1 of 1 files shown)


te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
```diff
@@ -696,6 +707,9 @@ private void bankReceived(boolean isBankA) {
     // echo if we set it with a sysex?
     verbose("Received Bank: " + (isBankA ? "A" : "B"));
     this.isBankA = isBankA;
+    // sometimes upon reconnect, buttons go all white but we do get a sysex for bank received -
+    // use this to send pad colors to stay consistent
+    sendPadColors();
```
jkbelcher, line 712: What's the scenario where this happens?  Haven't been able to repro.

And if this is on reconnect, could it live in the `onReconnect()` method?

jkbelcher (approved):
Merging to start part 8

PR #752, opened 2025-08-20
Quantized Random Strobe

jkbelcher:
Four updates to RandomStrobe, each in its own commit:
- Added quantized launch!  In Tempo Sync mode, run length is 1, 4, or 8 beats, and strobe speed is tempo-based.  In non-tempo sync mode, run length is milliseconds and strobe speed is Hz.
- Combined the internal `Element` classes into a single list.
- Proper handling for `modelChanged` events
- Added Mothership windows to RandomStrobe elements!

And one global update:
- Launch quantization depends on a global value being set in the tempo module.  To keep it consistent for shows I added it to the Control Panel (aka DevSwitch).

Video of quantized launch:
https://www.youtube.com/watch?v=Gegs3XgPoAw

### DIFF
--- a/te-app/src/main/java/titanicsend/effect/RandomStrobeEffect.java
+++ b/te-app/src/main/java/titanicsend/effect/RandomStrobeEffect.java
@@ -44,6 +44,8 @@ import titanicsend.model.TEPanelModel;
 @LXCategory("Titanics End")
 public class RandomStrobeEffect extends TEEffect implements LX.Listener {
 
+  private static final String TAG_MOTHERSHIP_WINDOW = "window";
+
   private abstract class Element {
     protected double offset;
 
@@ -83,6 +85,19 @@ public class RandomStrobeEffect extends TEEffect implements LX.Listener {
     @Override
     protected LXPoint[] getPoints() {
       return this.edge.points;
+    }
+  }
+
+  private class ModelElement extends Element {
+    private final LXModel model;
+
+    private ModelElement(LXModel model) {
+      this.model = model;
+    }
+
+    @Override
+    protected LXPoint[] getPoints() {
+      return this.model.points;
     }
   }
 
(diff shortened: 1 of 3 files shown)


zranger1:
Love the strobe button!  Can't believe (a) I didn't put one in originally, and (b) we've gone this long without it.   

A couple of minor things:
I'm not sure the "Quantize" button in the control panel is working -- after disabling quantize ,the effect will still only start on beat 1 of the next measure, even if I also disable sync in the effect itself

Also, minor qq (just a personal preference really, so up to you and other people with more live experience)  - once started I'd like the effect to stay active as long as the button is down, or at least keep starting new cycles as long as the button is down.  I promise to never hold it down through a whole set!

jkbelcher:
> once started I'd like the effect to stay active as long as the button is down, or at least keep starting new cycles as long as the button is down.

This is a totally reasonable idea, especially if it runs for increments of 1 beat.  Unfortunately we might not have the plumbing to implement it at the moment.  QuantizeTriggerParameter is specifically an extension of TriggerParameter, which always turns itself off immediately after turning itself on.  For the "press and hold" function we would need a momentary BooleanParameter with a quantized start and end.

jkbelcher:
Adjusted the label. What do you think?
image

te-app/src/main/java/titanicsend/effect/RandomStrobeEffect.java
```diff
@@ -139,22 +212,33 @@ public RandomStrobeEffect(LX lx) {
...
     for (TEEdgeModel edge : modelTE.getEdges()) {
-      EdgeElement e = new EdgeElement();
-      e.edge = edge;
+      EdgeElement e = new EdgeElement(edge);
       e.randomizeOffset();
-      this.edgeElements.add(e);
+      this.elements.add(e);
+    }
+
+    for (LXModel mothershipWindow : getModel().sub(TAG_MOTHERSHIP_WINDOW)) {
```
zranger1, line 238: Mothership!!!

zranger1 (approved):
Nice!   Code looks good, and everything works well.  This makes RandomStrobe better, and it's easy to control quantization once you understand what's going on!

Note for next year: it would be nice to allow quantized trigger controls the option of retriggering if the button is held down.

PR #753, opened 2025-08-21
Add new component to send color palette to Resolume

sina-cb:
Add Resolume OKLab gradient publisher; default-disable logging:
- Implemented titanicsend.osc.TEResolumeGradientPublisher and registered it in TEApp.Plugin as resolumePalette.
- Listens to active LXPalette/LXSwatch, binding only to colors 0 and 1; rebinds on swatch changes.
- Coalesces multiple H/S/B parameter events via engine-task scheduling to avoid duplicate handling and ensure updates occur after values settle.
- Generates a 16-color OKLab gradient:
  -- index 0 = black, index 4 = color0, index 15 = color1
  -- smoothed black→color0 segment with ease-out blending to reduce the jump near index 4
- Publishes normalized H/S/B to Resolume OSC:
  -- Base const: /composition/video/effects/colorize16palette/effect/
  -- Per-color paths: /lxcolor{1..16}/{hue|saturation|brightness}
- Introduced enableLogging flag (default false) to suppress palette change logs while keeping OSC publishing active.

### DIFF
--- a/te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
+++ b/te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
@@ -22,6 +22,8 @@ public class TEResolumeGradientPublisher extends LXComponent
   private boolean pendingLog0 = false;
   private boolean pendingLog1 = false;
   private boolean pendingPublish = false;
+  // Global logging switch (disabled by default)
+  private boolean enableLogging = false;
 
   // Base OSC path for the Resolume effect
   private static final String OSC_EFFECT_BASE =
@@ -130,22 +132,24 @@ public class TEResolumeGradientPublisher extends LXComponent
     // (hue, saturation, brightness). Scheduling onto the engine task queue ensures
     // we log only once per user change rather than 2-3 times, and do so on the
     // engine thread after all parameter updates have settled for this tick.
-    if (index == 0) {
-      if (pendingLog0) return;
-      pendingLog0 = true;
-      lx.engine.addTask(
-          () -> {
-            pendingLog0 = false;
-            logColorChange(0, this.color0);
-          });
-    } else if (index == 1) {
-      if (pendingLog1) return;
-      pendingLog1 = true;
-      lx.engine.addTask(
-          () -> {
-            pendingLog1 = false;
-            logColorChange(1, this.color1);
-          });
+    if (enableLogging) {
+      if (index == 0) {
+        if (pendingLog0) return;
+        pendingLog0 = true;
+        lx.engine.addTask(
+            () -> {
+              pendingLog0 = false;
+              logColorChange(0, this.color0);
+            });
+      } else if (index == 1) {
+        if (pendingLog1) return;
+        pendingLog1 = true;
+        lx.engine.addTask(
+            () -> {
+              pendingLog1 = false;
+              logColorChange(1, this.color1);
+            });
+      }
     }
 
     // Also coalesce OSC publishing to one action after changes to either color0 or color1
--- a/te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
+++ b/te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
@@ -58,18 +58,7 @@ public class TEResolumeGradientPublisher extends LXComponent implements LXSwatch
     // Initialize reusable ColorStops
     this.stopBlack.setRGB(LXColor.BLACK);
 
-    if (this.enabled.isOn()) {
-      bindToActiveSwatch();
-      // Debug: Log initialization
-      LX.log(
-          "TEResolumeGradientPublisher initialized - enabled: "
-              + this.enabled.isOn()
-              + ", logging: "
-              + this.enableLogging.isOn());
-      LX.log("  OSC engine available: " + (lx.engine.osc != null));
-    } else {
-      LX.log("TEResolumeGradientPublisher disabled...");
-    }
+    bindToActiveSwatch();
   }
 
   private void bindToActiveSwatch() {
--- a/te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
+++ b/te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
@@ -21,11 +21,11 @@ public class TEResolumeGradientPublisher extends LXComponent implements LXSwatch
   private boolean pendingLog0 = false;
   private boolean pendingLog1 = false;
   private boolean pendingPublish = false;
-  // Global logging switch (disabled by default)
-  private boolean enableLogging = false;
-
   public final BooleanParameter enabled =
-      new BooleanParameter("Enabled", false).setDescription("Enable OSC publishing to Resolume");
+      new BooleanParameter("Enabled", true).setDescription("Enable OSC publishing to Resolume");
+
+  public final BooleanParameter enableLogging =
+      new BooleanParameter("Enable Logging", true).setDescription("Enable color change logging");
 
   // Base OSC path for the Resolume effect
   private static final String OSC_EFFECT_BASE =
@@ -50,9 +50,18 @@ public class TEResolumeGradientPublisher extends LXComponent implements LXSwatch
     super(lx);
     this.activeSwatch = lx.engine.palette.swatch;
     addParameter("enabled", this.enabled);
+    addParameter("enableLogging", this.enableLogging);
     // Initialize reusable ColorStops
     this.stopBlack.setRGB(LXColor.BLACK);
     bindToActiveSwatch();
+
+    // Debug: Log initialization
+    LX.log(
+        "TEResolumeGradientPublisher initialized - enabled: "
+            + this.enabled.isOn()
+            + ", logging: "
+            + this.enableLogging.isOn());
+    LX.log("  OSC engine available: " + (lx.engine.osc != null));
   }
 
   private void bindToActiveSwatch() {
@@ -116,23 +125,25 @@ public class TEResolumeGradientPublisher extends LXComponent implements LXSwatch
     // (hue, saturation, brightness). Scheduling onto the engine task queue ensures
     // we log only once per user change rather than 2-3 times, and do so on the
     // engine thread after all parameter updates have settled for this tick.
-    if (enableLogging) {
+    if (this.enableLogging.isOn()) {
       if (index == 0) {
-        if (pendingLog0) return;
-        pendingLog0 = true;
-        lx.engine.addTask(
-            () -> {
-              pendingLog0 = false;
-              logColorChange(0, this.color0);
-            });
+        if (!pendingLog0) {
+          pendingLog0 = true;
+          lx.engine.addTask(
+              () -> {
+                pendingLog0 = false;
+                logColorChange(0, this.color0);
+              });
+        }
       } else if (index == 1) {
-        if (pendingLog1) return;
-        pendingLog1 = true;
-        lx.engine.addTask(
-            () -> {
-              pendingLog1 = false;
-              logColorChange(1, this.color1);
-            });
+        if (!pendingLog1) {
+          pendingLog1 = true;
+          lx.engine.addTask(
+              () -> {
+                pendingLog1 = false;
+                logColorChange(1, this.color1);
+              });
+        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  // Build a 16-color OKLab gradient with these anchors:
+  // index 0 = black, index 4 = color0, index 15 = color1
+  private void publishGradientIfReady() {
+    if (this.color0 == null || this.color1 == null) return;
+
+    // Prepare ColorStops for OKLab blending
+    LXGradientUtils.ColorStop stopBlack = new LXGradientUtils.ColorStop();
```
jkbelcher, line 185: Since 3 `ColorStops` are needed repeatedly, they can be allocated once in the constructor.  That'll save constant re-allocation in this method which gets called a lot.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  @Override
+  public void colorRemoved(LXSwatch swatch, LXDynamicColor color) {
+    rebindColors();
+  }
+
+  // LXPalette.Listener
+  @Override
+  public void swatchAdded(LXPalette palette, LXSwatch swatch) {}
```
jkbelcher, line 229: The `LXPalette.Listener` methods are not used here, so they can be removed.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+
+  private final LXParameterListener color1Listener =
+      (p) -> {
+        scheduleLog(1);
+      };
+
+  public TEResolumeGradientPublisher(LX lx) {
+    super(lx);
+    // Listen for palette changes and bind to the active swatch
+    lx.engine.palette.addListener(this);
```
jkbelcher, line 45: Not used, can be removed.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+  @Override
+  public void swatchRemoved(LXPalette palette, LXSwatch swatch) {}
+
+  @Override
+  public void swatchMoved(LXPalette palette, LXSwatch swatch) {}
+
+  @Override
+  public void dispose() {
+    unbindFromActiveSwatch();
+    try {
```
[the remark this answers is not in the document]
jkbelcher: Edit: `palette.listener` is not used, so this `removeListener()` line can be taken out.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+
+  private LXSwatch activeSwatch;
+  private LXDynamicColor color0;
+  private LXDynamicColor color1;
+  private boolean pendingLog0 = false;
+  private boolean pendingLog1 = false;
+  private boolean pendingPublish = false;
+  // Global logging switch (disabled by default)
+  private boolean enableLogging = false;
+
```
jkbelcher, line 27: We should have a way to enable/disable this from other places.  Add a `public final BooleanParameter enabled`, then call `addParameter("enabled", this.enabled)` in the constructor.

After you add the parameter, and because you already registered this component with the lx engine in TEApp, the parameter will get saved and loaded with the project file.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+            s,
+            b,
+            (cFromHSB & LXColor.R_MASK) >>> LXColor.R_SHIFT,
+            (cFromHSB & LXColor.G_MASK) >>> LXColor.G_SHIFT,
+            (cFromHSB & LXColor.B_MASK)));
+  }
+
+  private void scheduleLog(int index) {
+    // Debounce/coalesce: a single palette color change emits multiple parameter events
+    // (hue, saturation, brightness). Scheduling onto the engine task queue ensures
```
jkbelcher, line 132: Yeah the multiple parameter changed events for a color change can be a hassle.  What you've done here will work.  Another way to do it [theoretically, not saying you need to change this at the moment], would be:
- Add this class as an engine loop task.  it will get called every frame.
- When colors change (aka in the colorListener), set the pending flags to true.
- in the loop task method, if pending=true, then log/publish/etc.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+  }
+
+  private void rebindColors() {
+    unbindColorListeners();
+    if (this.activeSwatch == null) return;
+    int numColors = this.activeSwatch.colors.size();
+    if (numColors > 0) {
+      this.color0 = this.activeSwatch.getColor(0);
+      try {
```
jkbelcher, line 94: remove `try/catch` block

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+        this.color0.color.addListener(this.color0Listener, true);
+      } catch (Exception ignored) {
+      }
+    }
+    if (numColors > 1) {
+      this.color1 = this.activeSwatch.getColor(1);
+      try {
```
jkbelcher, line 101: remove `try/catch` block

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+      };
+
+  private final LXParameterListener color1Listener =
+      (p) -> {
+        scheduleLog(1);
+      };
+
+  public TEResolumeGradientPublisher(LX lx) {
+    super(lx);
+    // Listen for palette changes and bind to the active swatch
```
jkbelcher, line 44: call `addParameter("enabled", this.enabled)` here in constructor

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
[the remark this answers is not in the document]
sina-cb: Done

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+            publishGradientIfReady();
+          });
+    }
+  }
+
+  private boolean canSendOsc() {
+    return this.lx != null
+        && this.lx.engine != null
+        && this.lx.engine.osc != null
+        && this.lx.engine.osc.transmitActive.isOn();
```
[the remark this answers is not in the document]
sina-cb: Yes! Got it! I was wondering why! :)

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+            hex,
+            h,
+            s,
+            b,
+            (cFromHSB & LXColor.R_MASK) >>> LXColor.R_SHIFT,
+            (cFromHSB & LXColor.G_MASK) >>> LXColor.G_SHIFT,
+            (cFromHSB & LXColor.B_MASK)));
+  }
+
+  private void scheduleLog(int index) {
```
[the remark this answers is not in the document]
sina-cb: Started with just logging the color changes to see what's happening, the function name became a legacy I kept I think :D 

Renamed.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
sina-cb: Done

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
sina-cb: I have not idea how I got this to work, but the agent found a way to do so.

I like to chat to you about this if we have time later and see what's happening with parameters, and why we have multiple events.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
sina-cb: Done

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+        this.color1.color.removeListener(this.color1Listener);
+      } catch (Exception ignored) {
+      }
+      this.color1 = null;
+    }
+  }
+
+  private void rebindColors() {
+    unbindColorListeners();
+    if (this.activeSwatch == null) return;
```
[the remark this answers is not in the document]
sina-cb: Done

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+  private void unbindColorListeners() {
+    if (this.color0 != null) {
+      try {
+        this.color0.color.removeListener(this.color0Listener);
+      } catch (Exception ignored) {
+      }
+      this.color0 = null;
+    }
+    if (this.color1 != null) {
+      try {
```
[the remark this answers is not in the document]
sina-cb: Done

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,246 @@
...
+      } catch (Exception ignored) {
+      }
+      this.activeSwatch = null;
+    }
+    unbindColorListeners();
+  }
+
+  private void unbindColorListeners() {
+    if (this.color0 != null) {
+      try {
```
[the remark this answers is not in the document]
sina-cb: Done

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -53,15 +53,19 @@ public TEResolumeGradientPublisher(LX lx) {
...
-    LX.log("  OSC engine available: " + (lx.engine.osc != null));
+
+    if (this.enabled.isOn()) {
+      bindToActiveSwatch();
+      // Debug: Log initialization
+      LX.log(
+          "TEResolumeGradientPublisher initialized - enabled: "
+              + this.enabled.isOn()
+              + ", logging: "
+              + this.enableLogging.isOn());
```
jkbelcher, line 64: enableLogging will also always be its default value.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -107,43 +111,46 @@ private void logColorChange(int index, LXDynamicColor dynamicColor) {
...
-            "Palette color%d changed: %s hsb(%.1f, %.1f, %.1f) rgb(%d,%d,%d)",
-            index,
-            hex,
-            h,
-            s,
-            b,
-            (cFromHSB & LXColor.R_MASK) >>> LXColor.R_SHIFT,
-            (cFromHSB & LXColor.G_MASK) >>> LXColor.G_SHIFT,
-            (cFromHSB & LXColor.B_MASK)));
+    if (this.enableLogging.isOn()) {
```
jkbelcher, line 114: It's good to have this logging switch but it should be checked before the engine task is every created, so skip that overhead.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -107,43 +111,46 @@ private void logColorChange(int index, LXDynamicColor dynamicColor) {
...
-      } else if (index == 1) {
-        if (!pendingLog1) {
-          pendingLog1 = true;
-          lx.engine.addTask(
-              () -> {
-                pendingLog1 = false;
-                logColorChange(1, this.color1);
-              });
-        }
+    if (index == 0) {
```
jkbelcher, line 137: Check `enabledLogging.isOn()` before these two index ifs

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,255 @@
...
+        && this.lx != null
+        && this.lx.engine != null
+        && this.lx.engine.osc != null;
+  }
+
+  private void sendOsc(String address, float value) {
+    if (canSendOsc()) {
+      lx.engine.osc.sendMessage(address, value);
+    } else {
+      if (this.enableLogging.isOn()) {
```
jkbelcher, line 179: This will never happen.  You're checking for a state where logging is enabled and sending is disabled.  But at the top of colorChanged() you are bailing if sending is disabled and sendOsc will never be reached.

Also wouldn't want to flood the logs even if this could get reached.  I'd just take this log message out.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
jkbelcher: Even with a UI, the parameters will have their default values in the constructor.  The UI attached to this object after it is created and the constructor is complete.

te-app/src/main/java/titanicsend/osc/TEResolumeGradientPublisher.java
```diff
@@ -0,0 +1,235 @@
...
+    if (this.color0 == null || this.color1 == null) {
+      LX.log(
+          "Gradient publish skipped - color0: "
+              + (this.color0 != null)
+              + ", color1: "
+              + (this.color1 != null));
+      return;
+    }
+
+    if (!this.enabled.isOn()) {
```
jkbelcher, line 186: This won't cause any issues but it can be removed for clarity.  The method won't get called if it is disabled.

PR #754, opened 2025-08-21
Enable user presets for random strobe

jkbelcher:
Adding a UI section for user presets to `RandomStrobe`.  A simple one-line change.

Long term it would be nice to make this section collapsible for screen real estate.

image


jkbelcher:
This PR is stacked on #752 .

PR #756, opened 2025-08-21
Fix label width on UIUserPresets

jkbelcher:
Fixed: In the title label "PRESETS: <Pattern Name>", previously the label was getting cut short.

**After:**
image

**Before:**
image


zranger1 (approved):
Code looks good , and definite visual improvement!

PR #760, opened 2025-08-22
Shake Effect

zranger1:
One last thing...  a whole screen "shake" effect.   Can sync with beat and musical division.  Controls work pretty much the same as explode.

This is one of those things that looks cool on the screen but I don't know for sure what it'll be like on the Car, so just putting this out there for you to play with.

Take it if you like it and/or tell me how I can make it better!


te-app/src/main/java/titanicsend/effect/ShakeEffect.java
```diff
@@ -0,0 +1,146 @@
...
+      new BooleanParameter("Manual", true)
+          .setDescription("Enable manual triggering w/trigger button");
+
+  public final EnumParameter<Tempo.Division> tempoDivision =
+      new EnumParameter<Tempo.Division>("Division", Tempo.Division.WHOLE)
+          .setDescription("Tempo division to use in sync mode");
+
+  public final TriggerParameter trigger =
+      new TriggerParameter("Trigger", this::triggerListener)
+          .setMode(BooleanParameter.Mode.MOMENTARY)
```
jkbelcher, line 46: No need to set mode on `TriggerParameter`.  It's specifically made to be clicked once, fire its `runnable` only on the change to `value=TRUE`, and then reset itself to `value=FALSE`.

te-app/src/main/java/titanicsend/effect/ShakeEffect.java
```diff
@@ -0,0 +1,146 @@
...
+              1,
+              new FunctionalParameter() {
+                @Override
+                public double getValue() {
+                  return 2000 * speed.getValue();
+                }
+              }));
+
+  private void triggerListener() {
+    if (manualTrigger.isOn()) {
```
jkbelcher, line 63: This runnable will only get called for the OFF->ON direction, so no need to check `isOn()`.

te-app/src/main/java/titanicsend/effect/ShakeEffect.java
```diff
@@ -0,0 +1,146 @@
...
+import heronarts.lx.parameter.CompoundParameter;
+import heronarts.lx.parameter.EnumParameter;
+import heronarts.lx.parameter.FunctionalParameter;
+import heronarts.lx.parameter.TriggerParameter;
+import titanicsend.pattern.glengine.GLShader;
+import titanicsend.pattern.glengine.GLShaderEffect;
+
+@LXCategory("Titanics End")
+public class ShakeEffect extends GLShaderEffect {
+  double effectDepth;
```
jkbelcher, line 18: private?

jkbelcher (commented):
Added a couple comments on a quick glance.

PR #763, opened 2025-08-23
DJ Lights: independent left/right colors, presets list, and car positions

jkbelcher:
Improvements to DJ lights:
- Independent colors for left/right DJ lights (in Show and Easy patterns)
- Added UI user presets list (also added to Beacon patterns)
- Cleaned up the Saturation parameters to avoid confusion. Still using Eli's aftermarket (extra) desaturation.  No need to expose the internal knob, it doesn't do anything when color comes from the palette.

You'll need to delete the patterns from your project file and re-add them to see the new parameters!

image

### DIFF
--- a/te-app/src/main/java/titanicsend/dmx/pattern/DjLightsEasyPattern.java
+++ b/te-app/src/main/java/titanicsend/dmx/pattern/DjLightsEasyPattern.java
@@ -125,6 +125,6 @@ public class DjLightsEasyPattern extends DjLightsPattern
 
   @Override
   public void buildDeviceControls(UI ui, UIDevice uiDevice, DjLightsEasyPattern device) {
-    UIUtils.buildMftStyleDeviceControls(ui, uiDevice, device);
+    UIUtils.buildTEDeviceControls(ui, uiDevice, device);
   }
 }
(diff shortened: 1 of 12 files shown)


te-app/src/main/java/titanicsend/dmx/pattern/DjLightsEasyPattern.java
```diff
@@ -86,18 +81,24 @@ protected void run(double deltaMs) {
...
-    r -= w;
-    g -= w;
-    b -= w;
-
     int i = 0;
     for (DmxModel d : this.modelTE.getDjLights()) {
       if (d instanceof AdjStealthModel) {
+
+        // Determine Left or Right DJ light by checking fixture tags
+        boolean isRight = d.model.tags.contains("right");
```
andrewlook, line 89: (maybe for a future PR - we could be a bit more defensive here and either do:

```java
int color;
if (d.model.tags.contains("right")) {
  color = this.colorR.calcColor();
} else if (d.model.tags.contains("left")) {
  color = this.colorL.calcColor();
} else {
  throw new InvalidStateException("Invalid DJ Light fixture: must have either 'right' or 'left' tag");
}
int r = ((color >> 16) & 0xff);
// ...
```
 
Or  a lighter-weight version might be:
```java
// Determine Left or Right DJ light by checking fixture tags
boolean isRight = d.model.tags.contains("right");

if (!isRight && !d.model.tags.contains("left")) {
  throw new InvalidStateException("Invalid DJ Light fixture: must have either 'right' or 'left' tag");
}

int color = isRight ? this.colorR.calcColor() : this.colorL.calcColor();
int r = ((color >> 16) & 0xff);
// ...
```

jkbelcher: Reasonable for it to give feedback.  I think soft fail by logging an error would be enough, rather than throwing an exception.

PR #766, opened 2025-08-23
MiniLab3 Part 10: Faders

jkbelcher:
Implemented faders.  Mapped the first one to global audio gain as an example.

Faders have modes including `Scale` and `Pickup` that prevent parameter jump (if the parameter value was changed in the UI and then changed with the fader).

Fader parameters are currently hardcoded in this file rather than going through `GobalEffectsManager`.  Should be good enough if they are always global parameters.

### DIFF
--- a/te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
+++ b/te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
@@ -14,8 +14,11 @@ import heronarts.lx.midi.MidiControlChange;
 import heronarts.lx.midi.MidiNote;
 import heronarts.lx.midi.MidiNoteOn;
 import heronarts.lx.midi.MidiPitchBend;
+import heronarts.lx.midi.surface.LXMidiParameterControl;
 import heronarts.lx.midi.surface.LXMidiSurface;
+import heronarts.lx.parameter.EnumParameter;
 import heronarts.lx.parameter.LXListenableNormalizedParameter;
+import heronarts.lx.parameter.LXParameter;
 import heronarts.lx.utils.ObservableList;
 import java.util.List;
 import titanicsend.app.effectmgr.GlobalEffectManager;
@@ -249,11 +252,24 @@ public class EffectsMiniLab3 extends LXMidiSurface implements LXMidiSurface.Bidi
   private boolean isBankA = false;
   private boolean isDAW = false;
 
+  public final EnumParameter<LXMidiParameterControl.Mode> faderMode =
+      new EnumParameter<>("Fader Mode", LXMidiParameterControl.Mode.SCALE)
+          .setDescription("Fader behavior when controlling parameters");
+
+  private final LXMidiParameterControl[] faders = new LXMidiParameterControl[NUM_FADERS];
+
   // Constructor
 
   public EffectsMiniLab3(LX lx, LXMidiInput input, LXMidiOutput output) {
     super(lx, input, output);
     this.effectManager = GlobalEffectManager.get();
+
+    for (int i = 0; i < NUM_FADERS; i++) {
+      this.faders[i] = new LXMidiParameterControl();
+    }
+    updateFaderMode();
+
+    addSetting("faderMode", this.faderMode);
   }
 
   // Connection
@@ -296,6 +312,8 @@ public class EffectsMiniLab3 extends LXMidiSurface implements LXMidiSurface.Bidi
   private void register() {
     this.isRegistered = true;
 
+    registerFaders();
+
     int slotIndex = 0;
     for (Slot<? extends LXDeviceComponent> slot : this.effectManager.slots) {
       setSlot(slotIndex++, slot);
@@ -313,6 +331,8 @@ public class EffectsMiniLab3 extends LXMidiSurface implements LXMidiSurface.Bidi
 
   private void unregister() {
     this.isRegistered = false;
+
+    unregisterFaders();
 
     this.effectManager.slots.removeListener(this.slotsListener);
     this.effectManager.triggerSlots.removeListener(this.triggerSlotsListener);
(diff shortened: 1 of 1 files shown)


te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
```diff
@@ -714,7 +765,10 @@ private void knobReceived(int knob, int value) {
     parameterSetValue(knob, value);
   }
 
-  private void faderReceived(int fader, int value) {
+  private void faderReceived(MidiControlChange cc, int fader, int value) {
+    this.faders[fader].setValue(cc);
```
andrewlook, line 769: is it worth doing a null-check here? I know we're pre-initializing the faders so they're present even if unassigned, but maybe worth having a check just in case that precondition ever changes?

te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
```diff
@@ -881,6 +935,7 @@ private void parameterSetValue(int parameterIndex, int value) {
   /** Set the value of a global parameter */
   private void globalParameterSetValue(int globalParamIndex, int value) {
     verbose("Global Parameter " + globalParamIndex + ": set to " + value);
+    // Nothing to do here, we already set the value through the LXMidiParameterControl fader
```
andrewlook, line 938: think we should just remove this function? I forget if we had some other rationale for including this function stub

te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
```diff
@@ -469,6 +489,37 @@ private void slotStateChanged(int slotIndex) {
...
+    // Global Audio gain
+    this.faders[0].setTarget(this.lx.engine.audio.meter.gain);
+
+    // Example: Global Audio Stems gain
+    // this.faders[0].setTarget(AudioStems.get().gain);
+  }
+
+  private void unregisterFaders() {
+    for (LXMidiParameterControl fader : this.faders) {
+      fader.setTarget(null);
```
andrewlook, line 519: Hmm normally we use unregister for teardowns/listeners/things that use resources.. is there any possibility of a memory leak or something if we hold onto a reference to the target? Or would it make sense to delete the whole parameter from the array?

My guess is that your thinking here is that maybe if we had a different driver, or different ways of mapping params to the faders, when you disable the driver you wouldn't want to accidentally still have it controlling something if the device gets re-enabled with some different config (not how it works today, but within the realm of possibilities i could imagine). Is that what you had in mind here, in terms of why this teardown is worth doing?

I realize that this probably doesn't have any big consequences, partly just curious to hear how you and Slee think about teardown methods and what's worth cleaning up.

andrewlook (approved):
Wow, this is great! Sorry i didnt manage to fit this one into the burn this year.

I had tried midi mapping the faders directly without code, as an override on our NDI input channel, and they behaved really strangely - jumping from zero to max/etc. Ended up having to abandon that mapping. Thanks for getting this plumbing in place so it's more solid for next time!

andrewlook:
✅ verified locally

https://github.com/user-attachments/assets/d1d6fd16-ede1-4811-961d-3fbc2403ef8d

andrewlook:
One side note: when i tried just midi-mapping the faders directly (before i knew about this PR), i got wonky effects because the absolute min/max values were mapped directly to the midi values. I think theres a way to tweak those ranges in the midi mapping UI widget, but mid-concert I abandoned the midi mapping approach and just used the mouse.

In this example, i tried midi mapping audio range to a fader, and instead of -48 to 48, it only goes like 6 to 96 or something odd. Just wanted to record that gotcha while its fresh in my mind.

https://github.com/user-attachments/assets/175751fe-836d-468e-ad3f-8c80d60f94ca

So i'm very happy you got this native MIDI fader setup working - that'll be much smoother here.

Just for my own learning, if we didnt have this PR, is there some obvious fix i could have made when doing the midi mapping?

te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
jkbelcher: No, it's a final array with all items initialized in the constructor.  The design is that these are created once and do not change.  If someone modifies that assumption they need to do a full sweep of everything that touches the array.  If they don't do a proper sweep they should get a crash on the first run so they'll know they broke an assumption and can look for other unintended consequences.

jkbelcher: This mimics the use of `LXMidiParameterControl` faders in the `APC40Mk2` class.

te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
jkbelcher: I can't recall at the moment.  Maybe leave it in for a while in case we remember?

te-app/src/main/java/titanicsend/lx/EffectsMiniLab3.java
jkbelcher: When one of these `LXMidiParameterControls` is targeting a parameter, it adds a listener to that parameter.  Listeners always need to be removed before application tear-down. 

We may be unregistering because we're disposing, in which case we definitely want to remove all listeners.  Or we may we unregistering because the surface was disabled, in which case we must remove listeners so that we can re-add them if the surface is re-enabled.  When a midi surface is disabled it fully disconnects from targets.

jkbelcher:
> I had tried midi mapping the faders directly without code, as an override on our NDI input channel, and they behaved really strangely - jumping from zero to max/etc. Ended up having to abandon that mapping.

Interesting.  I just did a test and was able to runtime-map one of these faders to a channel level fader.

image

image

jkbelcher:
I think this was just a mix-up on the ranges of the audio parameters.  Gain ranges from -48 to 48, and Range ranges from 6 to 96.  Try mapping to Gain and see if you get the full range.

PR #770, opened 2025-08-24
Refactor names related to textures (low priority)

jkbelcher:
Did a refactor pass (name changes only) to clarify a few things:
- iChannel textures
- ModelCoords and ModelIndex textures

There were places where "coords" was used in reference to both model-derived textures, and places where "coords" was used only in reference to the first texture.  Attempting to create consistency I chose:
- "ModelTextures" refers to both lxmodel-derived textures
- "Coords" refers to the first texture and "Index" refers to the second texture.

### DIFF
--- a/te-app/src/main/java/titanicsend/ndi/NDIOutShader.java
+++ b/te-app/src/main/java/titanicsend/ndi/NDIOutShader.java
@@ -32,7 +32,7 @@ public class NDIOutShader extends GLShader implements GLShader.UniformSource {
   private String ndiStreamLabel = "TitanicsEnd";
 
   private int modelCoordsTextureHandle = UNINITIALIZED;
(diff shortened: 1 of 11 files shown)


zranger1 (approved):
Looks good!  Tested - the iChannel[n] textures and internal coordinate and indexing samplers are all working.   I especially appreciate the cleanup of the messy naming  around the shadertoy-compatible iChannels.  That's been around since before year 1, and it has always been confusing.

Approved, but maybe hold merge 'till after the burn.

andrewlook:
note: i merged the stacked PR https://github.com/titanicsend/LXStudio-TE/pull/773 into this branch, then will rebase it into master (to preserve the 3 resulting commits)

andrewlook:
thanks for thinking ahead here @jkbelcher to make the naming something that is clearer before we keep adding stuff onto this 🙏

PR #776, opened 2025-08-26
Bug fix: suppress command engine with UIOffairParameter

jkbelcher:
We had suppressed the command engine for the `UserPreset` `Selector` by using a custom variation of a `UIKnob`.  This protection was lost when we wrapped it in an `OffairParameter`.

Added the same safety wrapper to `OffairParameter`, but had to make it non-generic to work with the `LXRegistry`.

You could encounter the bug by selecting a non-live pattern and changing the `UserPreset` selector.


jkbelcher:
Agreed, the listbox is not linked to the parameter, but should be fixed eventually.

PR #778, opened 2025-08-26
Visual indicator on UserPresets knob when there are no presets

jkbelcher:
Stacked on #776 .  Merge that one first.

This makes the `Preset` knob look disabled when there are no presets available.  Requested by @andrewlook.

image

### DIFF
--- a/te-app/src/main/java/titanicsend/parameter/UIOffairDiscreteParameter.java
+++ b/te-app/src/main/java/titanicsend/parameter/UIOffairDiscreteParameter.java
@@ -1,0 +1,22 @@
+package titanicsend.parameter;
+
+import heronarts.glx.ui.UI;
+import heronarts.glx.ui.component.UIKnob;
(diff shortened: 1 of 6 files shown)


te-app/src/main/java/titanicsend/parameter/UIOffairDiscreteParameter.java
```diff
@@ -18,5 +19,13 @@ public UIOffairDiscreteParameter(
...
     // There's an issue when this wraps a preset selector. It needs to not use the command engine.
     setUseCommandEngine(false);
+
+    // For visual aid, disable if there is only one option
+    LXParameterListener optionsChangedListener =
+        (p) -> {
+          setEditable(parameter.getOptions().length > 1);
+          redraw();
+        };
+    addListener(parameter.optionsChanged, optionsChangedListener, true);
```
andrewlook, line 29: 🔥

PR #789, opened 2026-03-06
new mothership pattern: afterglow

piemonte:
- Add MothershipAfterglow pattern: white dots pulse outward from window corners along sides, adapted from the edge-based Afterglow for Mothership geometry


piemonte:
https://github.com/user-attachments/assets/fe862791-df81-41ce-876d-35309bde4c48

Mothership.run/Template Application.run.xml
jkbelcher, line 1: Is there something unique about this run config that’s different from existing ones?

te-app/src/main/java/titanicsend/pattern/piemonte/MothershipAfterglow.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Mothership Afterglow Pattern
+ *
+ * <p>Adapted from Afterglow for the Mothership geometry. Creates white dots pulsing outward from
+ * window corners along window sides. Best viewed in deep playa.
+ */
+@LXCategory("Mothership")
+public class MothershipAfterglow extends TEPerformancePattern {
+
+  protected final SawLFO phase =
```
jkbelcher: This can be private - no classes extend this pattern.  If it’s needed by a child class later we can come back and open it up.

jkbelcher (commented):
Just did a quick read.  Am not at the computer to run it yet.

Looking good overall!  Good job doing a model-specific pattern.  Left a couple comments.

Mothership.run/Template Application.run.xml
piemonte: Oops, that's a sloppy mistake. Was playing around with configs and didn't remove that.

te-app/src/main/java/titanicsend/pattern/piemonte/MothershipAfterglow.java
```diff
@@ -0,0 +1,139 @@
...
+
+  private Map<Integer, Float> sidePhaseOffsets = new HashMap<>();
+  private Random random = new Random();
+
+  public MothershipAfterglow(LX lx) {
+    super(lx, TEShaderView.ALL_POINTS);
+
+    startModulator(this.phase);
+
+    controls.setRange(TEControlTag.SPEED, 1, 0, 1);
```
jkbelcher: Our speed parameter is configured as bipolar, so we usually range it from [-value] to [value], which means no movement when the dial is centered. Think you could adjust the pattern to handle that? So the movement would go in reverse when Speed is negative.

te-app/src/main/java/titanicsend/pattern/piemonte/MothershipAfterglow.java
```diff
@@ -0,0 +1,139 @@
...
+    int baseColor = calcColor();
+
+    int sideIndex = 0;
+    for (LXModel window : getModel().sub("window")) {
+      for (LXModel side : window.sub("s")) {
+        if (!sidePhaseOffsets.containsKey(sideIndex)) {
+          sidePhaseOffsets.put(sideIndex, random.nextFloat());
+        }
+
+        float sidePhaseOffset = sidePhaseOffsets.get(sideIndex) * randomness;
```
jkbelcher, line 75: Very cool, love the real-time adjustability of the offsets.

te-app/src/main/java/titanicsend/pattern/piemonte/MothershipAfterglow.java
```diff
@@ -0,0 +1,139 @@
...
+
+    addCommonControls();
+  }
+
+  @Override
+  protected void runTEAudioPattern(double deltaMs) {
+    float phase = this.phase.getValuef();
+    int dotSize = (int) getSize();
+    float fadeDistance = (float) getQuantity();
+    int numPulses = (int) getWow2();
```
jkbelcher: It seems like `getQuantity()` would be the source for `numPulses` rather than `wow2`?

te-app/src/main/java/titanicsend/pattern/piemonte/MothershipAfterglow.java
```diff
@@ -0,0 +1,139 @@
...
+    float randomness = (float) getWow1();
+
+    for (LXPoint point : model.points) {
```
jkbelcher: We have a helper for this on TEPattern, `clearPixels()`.  But I think all frames are pre-cleared now? Try it without and see...

te-app/src/main/java/titanicsend/pattern/piemonte/MothershipAfterglow.java
piemonte: 🙏 good catch!

piemonte:
@jkbelcher thanks for the help! incorporated the feedback and pulled out the dependency change. will test that separately and lyk.

piemonte:
@andrewlook @jkbelcher stripped out all the un-necessary stuff outside of the pattern work for whenever you may have a free moment. 🙏

jkbelcher (approved):
Tested bipolar speed, looks good!

